PR #38, opened 2015-04-09
second draft of enums post.

pnkfelix:
cc @nikomatsakis @aturon 

[rendered](https://github.com/pnkfelix/blog.rust-lang.org/blob/fsk-enums-post/_posts/2015-04-17-Enums-match-mutation-and-moves.md)


pnkfelix:
I already said this to @nikomatsakis, but I'll add it here: This draft may well be far too long for our target market for these blog posts.  Please feel free to be ruthless in suggesting edits, or just say "just section X should be the post."

bombless:
What does `#[cfg(alternative)]` mean?

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,734 @@
...
+style above allows the Rust compiler to assign the values itself; it
+will start from 0 and count upwards.
+
+Alternatively, one can manually assign specific values to enum
+variants, interrupting the automatic value assignment performed by the
+compiler. Thus, we could instead have written the below, and get the
+same assignment of discriminant values:
+
+```rust
+#[cfg(alternative)]
```
bombless: What does `#[cfg(alternative)]` mean?

pnkfelix: I cannot tell if you are suggesting that more text be added to the post, or if you just want more context.

My method for writing this post, and my overall goal, is that someone can cut-and-paste all of the code into a file and run it.  (Indeed, my method of authorship was to run the `.md` file through a script to convert to-and-from a `.rs` file.)

So a problem arises if I want to provide alternative definitions, such as this one here, since we do not want to cause compiler errors due to duplicate definitions.

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,740 @@
...
+The dereference `*s` works because `match` works on both "L-value" and
+"R-value" expressions; L-value expressions evaluate to memory
+locations, while R-value expressions evaluate to values that are then
+stored, if necessary, in temporary memory locations. It simplified our
+code because allowed us to sidestep writing `&mut` at the beginning of
+each of our patterns, which was not such a great burden for this tiny
+code snippet, but can be much more annoying when working with `match`
+expressions that have many more arms, amplifying the redundancy.
+
+## Avoiding unsoundness
```
nikomatsakis: Maybe you could move this into a gist, and just link to it from the post as a kind of appendix?

_posts/2015-04-17-Enums-match-mutation-and-moves.md
nikomatsakis: (Seems like we should prune these from the final post though)

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,740 @@
...
+```rust
+#[test]
+fn demo_one_letter_collect() {
+    let qpk = [Queen, Pawn, King];
+    let qpk: String = qpk.iter().map(|r|{one_letter(*r)}).collect();
+    assert_eq!(qpk, "QPK");
+}
+```
+
+Encoding shapes via ASCII characters has worked fine since the 1960's,
```
nikomatsakis: I love this line

_posts/2015-04-17-Enums-match-mutation-and-moves.md
pnkfelix: The whole "Avoiding unsoundness" section, you mean?  Okay.

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,740 @@
...
+    King, Queen, Rook, Bishop, Knight, Pawn,
+}
+```
+
+The `derive` line automatically generates support for three common
+operations: the `Copy, Clone` gives us the ability to freely copy or
+clone instances of `PieceShape`, the `Debug` provides support for
+printing each piece-rank into an output buffer, e.g. via
+`println!("piece: {:?}", piece)`, and `PartialEq, Eq` provides support
+for the `==` operator on instances of the enum.
```
whipsch: Perhaps the brief explanations here should be worded a bit differently, something like this?

> The `derive` line automatically generates support for three common operations: `Copy` and `Clone` give us the ability to freely copy or clone an instance of `PieceShape`; `Debug` allows us to print each piece-rank into an output buffer, e.g. via `println!("piece: {:?}", piece)`; and finally `PartialEq` and `Eq` enable use of the `==` operator on instances of `PieceShape`.

Also, I can see that you are trying to answer 'what the heck is this derive thing??' without having to explain attributes (or even `derive` specifically) and traits, but maybe they should at least be identified as such with some sort of note about the full explanation being beyond the scope of this post:

> `derive` is an attribute that tells the compiler to insert generated implementations of the traits we specified. Traits are **bla bla bla** and do **some things**.  We won't cover the details of attributes and traits in this tutorial, but for more information you can read the following sections of *The Rust Programming Language*: [Attributes](http://doc.rust-lang.org/nightly/book/attributes.html), [Traits](http://doc.rust-lang.org/nightly/book/traits.html).

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,740 @@
...
+These enums, where each variant is just a name with an associated
+(implicitly- or explicitly-assigned) integer value, are known as
+"C-style enums" in Rust parlance, since the syntax largely matches
+that used for the `enum` construct in the C language.
+
+(However, note that not every C enum immediately corresponds to an
+analogous enum in Rust; in particular, Rust enforces an invariant that
+each variant be distinct from all other variants in the enum, while C
+allows one to assign the same integer value to multiple enum
+variants.)
```
whipsch: This sentence about C enum correspondence seems a bit much to be included within the main portion of the tutorial.  Perhaps it should be turned into a footnote instead?

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,740 @@
...
+
+### Constructing enum instances
+
+After defining an enum, one accesses the variants it defines via the
+path syntax `enum_name::variant_name`. So using these pieces is
```
whipsch: Nitpicking: should this be written `EnumName::VariantName`?

aturon:
@pnkfelix 

Thanks for the draft! 

====

Overall, I think it's worth taking a step back and considering:

1. Who is the audience for this post?
2. What key message or two do you want to get across?

For (1), I'd say: a person who has heard something about Rust, and knows at least one non-functional language well (ranging from C++ to Ruby). Many will have heard something about pattern matching, but probably won't know how it works out in practice.

For (2), I'd say, first and foremost: enums and pattern matching are *really useful, powerful, and ergonomic*. Second, and less important: Rust has some interesting twists relating these concepts to ownership and borrowing.

I would say a *non*-goal is giving people enough knowledge to actually go write programs with enums and pattern matching directly; we have other documentation and guides that do that well. This is more about "selling" the concepts, explaining why they are an important aspect of Rust, and what interesting things Rust brings to the table.

So overall, I'd suggest doing a sweep through the post with these (or your own answers to them) in mind, ruthlessly editing and sweeping away material that doesn't fit the audience and goals.

====

Here are some more detailed comments/observations from an initial read through.

1. The post doesn't have much in the way of an introduction. If I'm a random programmer on the internet, why should I bother with this post? Is `enum`/`match` a "corner" of Rust, or a really useful feature that might draw me to the language? (I would aim this primarily at people who haven't programmed with functional languages/pattern matching; `match` should be an amazing revelation, and you should say so up front.)

2. I'd suggest avoiding "affine-typing" and other more academic terms in favor of "ownership", which we've been pretty consistently using when talking about Rust. 

3. You might want to say at the outset that the contents of the post are executable Rust code if pasted one after another.

4. "This is used in the second match expression, so that the shape held in p is left in place." I think someone not already familiar with Rust will have a hard time making sense of that sentence -- it may be worth just leaving it, and the parenthetical that follows, out.

5. "It is easy to write such a promotion in a functional programming style" -- not clear that a reader of this post will know what that means. Perhaps just include a defining remark: "that is, by producing a new value rather than mutating the old one."

6. "The invocation promote_to_queen is not passing a reference (mutable or otherwise) to p; it is creating a copy of the Piece." I worry that this will be confusing to people not already familiar with Rust. The [previous post](http://blog.rust-lang.org/2015/04/10/Fearless-Concurrency.html) in the series emphasized that argument passing always means a transfer of ownership, and in the case of references you're transferring ownership of a borrow. But it didn't talk at all about `Copy` or the notion that parameter passing might ever involve implicit copying. Maybe it's worth having a digression on `Copy` early in the post? Or you could remove `Copy` and explain this in terms of ownership transfer? (Or, alternatively, maybe this bad attempt can just be dropped?)

7. "Try, try again" -- this doesn't explain the use of `&mut` in the patterns, nor `mut`. (Actually, come to think of it, the use of variable-binding `mut` in general isn't touched on in this post, nor was it in the previous one.) This starts feeling very complicated, especially when `ref mut` patterns are introduced.

8. "Avoiding unsoundness". The opening paragraph was pretty dense/difficult to follow, and without a clear understanding of what you're worried about, the various examples you give become impenetrable. Can you expand the paragraph and explain a bit more slowly, in prose, the potential soundness problem?

nikomatsakis:
So I read it through with a more critical eye than previous reads. I think it's quite easy to read, but I do think we can pare it down somewhat. As @aturon said, I think the goal here is not to lay out all the details, but just convey that match is nice to use in practice. Some comments:

I wonder if we can remove the coverage of C-like enums? I guess people may wonder about it, so perhaps worth keeping. To obviate the question of `#[cfg(alternative)]`, perhaps we could just give the c-like enum another name like `enum PieceShapeExplicit { ... }`.

I thought the "binding in match patterns" felt a bit hard to follow. Specifically the sentence that starts "Rather than matching it as constant, one can bind the payload...". It seems like that is using a fair amount of technical jargon and is quite removed from the code example. I feel like it'd be better to give an example of the pattern `White(shape)` and explain that this introduces a variable `shape` with the associated shape value.

I feel like we can probably drop the mutation section and instead, perhaps, replace with a "ref binding" section. This could start out with an example using `ref foo` and passing it to a function that take `&Shape`, and then show `ref mut` in order to do the "in-place promotion" -- basically just skip to the right answer.

I'd probably just move the "Avoiding unsoundness" thing into a gist, but perhaps it does make sense to include some kind of stripped down version explaining how `ref` and `ref mut` patterns produce borrows and fit into Rust's general vision of things. But I don't know, we've not explained that "general vision" anywhere on this blog at least... (hmm, maybe a good idea for a post)

pnkfelix:
@aturon thanks for the great critical feedback.  I have been feeling like this material is super-dry and hard to get excited about.  Taking a step back and thinking about "why are enums and pattern-matching *cool*" is a really good idea ... I still remember how enthused my professors were to show us FP (specifically Haskell) precisely because of Pattern Matching ... I'll think over all of this feedback and see if I can come up with something more lively

aturon:
Thanks for the new draft, @pnkfelix! I think it’s much improved.

Some high level thoughts:

* You might consider re-ordering the statement/expression orientation section to the end, or at least until after the section on algebraic data types. I think getting to algebraic data types more quickly is good, and I feel like the L-value section is also the real “meat” of the post in terms of interaction with ownership. The statement/expression orientation bit feels somewhat less important (though still shows off some slick design work).

* One exercise that could be useful: at the top of each subsection, try to write a single sentence that summarizes why the thing in that subsection is valuable or important. What is the impact?

* The conclusion still has the flavor of “this post is a tour”, but the post should be as much about a sales pitch for Rust and these features in particular. I think you should feel more free to editorialize. You could say, for example, that the post has shown that Rust’s design works to incorporate some very powerful features from the functional programming world (algebraic types, pattern matching) while adapting them to more imperative styles, and to ownership/borrowing in particular. The result is both ergonomic and expressive.

Some detailed comments:

* “applicative style” of programming is likely to be unfamiliar to readers, as may be statement- versus expression- oriented. Not sure if it’s that’s a big deal, as long as it’s spelled out in the corresponding subsection.

* The Basics of match: this feels a lot stronger, nice work! In particular you quickly beat a path toward a compelling example where `match` is letting you get a lot done, very concisely. The one thing that’s missing is making some explicit value judgment/sales pitch about this — maybe just saying that the ability to simultaneously to case analysis and binding leads to powerful, clear and concise code, as the second example shows.

* Expression/statement-orientation. This is missing an explanation of the distinction and why it’s important; I’m not sure how widely familiar that is to the audience. The second example makes a neat point, but is a bit subtle (may want to be spelled out a bit more). In particular, it assumes knowledge that string literals are borrows, which is not very obvious :) Also, the bit saying “At the same time, we can access the data that is held within string, because a reference to that data is held by the borrowed variable” is a bit misleading, since on the second path you’re not actually borrowing from the string.

* The algebraic data types section is nice and clean.

* L-values: this is a pretty detailed section that demands careful reading, but I think manages to be quite clear. Nice work! One thing that you don’t talk about is the dereference `*t` in the match. I realize this feels obvious, but I think it’s worth reiterating that borrowing also entails *indirection*, and that the dereferencing here and the L-value semantics for `match` kind of “cancel each other out”.

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,613 @@
...
+will always fall into one of the above two cases. One will never
+encounter a `BinaryTree::Node` that does not have a left-hand child.
+There is no need to check for null.
+
+One *does* need to check whether a given `BinaryTree` is a `Leaf` or
+is a `Node`, but the compiler statically ensure such checks are done:
+you cannot accidentally interpret the data of a `Leaf` as if it were a
+`Node`, nor vice versa.
+
+```rust
```
pnkfelix: Maybe add intro before demo code here

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,613 @@
...
+
+Thus, the bindings of `payload` in `tree_weight_v2` both have type
+`i32`; the `i32` type implements `Copy`, so the weight is copied into
+`payload` in both arms.
+
+However, when matching a value of type `T`, a `ref`-pattern `ref i`
+will, on a successful match, merely *borrow* a reference into the
+matched data. In other words, a successful `ref i` match of a value of
+type `T` will imply that `i: &T`. Thus, in the `Node` arm of
+`tree_weight_v2`, `left` will be reference the left-hand box (which
```
pnkfelix: Typo: will be reference

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,613 @@
...
+    sometimes_initialize(23);  // this invocation will initialize `string`
+    sometimes_initialize(123); // this one will not
+}
+```
+
+The interesting thing about the above code is that after the `match`,
+we are not allowed to directly access `string`, because the compiler
+requires that the variable be initialized on every path through the
+program. At the same time, we *can*, via `borrowed`, access the data that
+is held *within* `string`, because a reference to that data is held by the
```
pnkfelix: "that *may* be held"?  (Or just "held"?) The text "is held" conveys stronger property than i wished

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,613 @@
...
+`borrowed` variable when we go through the first match arm, and we
+ensure `borrowed` itself is initialized on every execution path
+through the program that reaches the `println!` that uses `borrowed`.
+
+(The compiler ensures that no outstanding borrows of the
+`string` data could possible outlive `string` itself, and the
```
pnkfelix: Typo: could possible outlive

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,636 @@
...
+`match` enforces exhaustive case analysis, which helps catch bugs in
+program logic and ensures that the value of a match expression is
+well-defined.
+
+Case analysis is an important problem solving technique: Dividing a
+problem into some number of separate cases, and then analyzing each
+case individually.
+
+For this method of problem solving to work, the cases must be
+*collectively exhaustive*; otherwise, a case that was not covered
```
nikomatsakis: I think somewhere here we should define the term *exhaustive*, just to say something like "Rust makes sure you have a match arm that covers every possible variant"

nikomatsakis: Oh, I guess you do say that below -- but it seems to take a while to arrive. Maybe we can just make this "intro" snappier. Like "One really useful way to tackle a complex problem is to break it down into a set of individual cases and look at each case individually. But for this to work, you have to be sure that your breakdown is *exhaustive* -- that is, the individual cases you identified really cover all the possible scenarios. Using enums and match in Rust can help you with this."

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,636 @@
...
+}
+```
+
+The `suggest_guess_fixed` function illustrates that `match` can handle
+some cases early (and then immediately return from the function),
+while computing whatever values are needed from the remaining cases
+and letting them fall through to the remainder of the function
+body. We can add this special case handling via `match` without fear
+of overlooking a case, because `match` will enforce the case
+analysis to be exhaustive.
```
nikomatsakis: Maybe it's worth talking about this as a refactoring aid? I certainly rely a lot on being able to add a variant to an enum and being assured I will be directed to all the places in my code that need to be updated.

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,636 @@
...
+The following code uses range patterns to simplify things, but also,
+it is written in a style similar to a `switch` in a statement-oriented
+language like C (or C++, Java, et cetera), where the arms of the
+`match` are executed for their side-effect alone:
+
+```rust
+fn num_to_ordinal(x: u32) -> String {
+    let suffix;
+    match (x % 10, x % 100) {
+        (1, 1) | (1, 21...91) => {
```
nikomatsakis: this code blew my mind somehow. I have never thought about what a frickin' annoying problem it is to generate these suffixes.

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,673 @@
...
+Rust's `match` expression is a construct that offers an interesting
+combination of such features and restrictions. A `match` expression
+takes an input value, classifies it, and then jumps to code written to
+handle the identified class of data.
+
+In this post we explore how Rust processes such data via `match`.
+The crucial elements that `match` and its counterpart `enum` tie
+together are:
+
+* Structural pattern matching of algebraic data types: The `match`
```
pnkfelix: (todo: drop "of algebraic data types")

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,673 @@
...
+
+* Exhaustive case analysis, which ensures that no case is omitted
+  when processing an input.
+
+* `match` embraces both imperative and functional styles of
+  programming: The compiler's static analyses work hard to ensure
+  statement-oriented programming remains palatable, rather than
+  forcing everyone to adopt an expression-oriented mindset.
+
+* Destructuring bind of *L-values*: Rust encourages the developer to
```
pnkfelix: (todo: find less jargony way to write this bullet.  Maybe just "match can borrow or take ownership, as appropriate")

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,673 @@
...
+establish a foundation for the discussion: What does `match` look
+like, and how does it work?
+
+### The Basics of `match`
+
+The `match` expression in Rust has this form:
+
+```rust
+match INPUT_EXPRESSION {
+    PREDICATE_1 => RESULT_EXPRESSION_1,
```
pnkfelix: a reviewer in `#rust` recommends not using the term "PREDICATE" here.

todo: replace with "PATTERNS_i"

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,673 @@
...
+}
+```
+
+where each of the `PREDICATE_i` contains at least one *pattern*. A
+pattern describes a subset of the possible values to which
+`INPUT_EXPRESSION` could evaluate.
+The syntax `PREDICATE => RESULT_EXPRESSION` is called a "match arm",
+or simply "arm".
+
+Patterns can match atomic values, like integers or characters; they
```
pnkfelix: todo: reviewer from `#rust` recommends adding reference for "atomic value"

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,673 @@
...
+	suggest_guess(19, Answer::Bingo);
+}
+```
+
+(Incidentally, nearly all the code in this post is directly
+executable; you can cut-and-paste the code snippets into a file
+`demo.rs`, compile the file with `--test`, and run the resulting
+binary to see the tests run.)
+
+Patterns can also match structured data (e.g. tuples, slices, user-defined
```
pnkfelix: a reviewer in `#rust` says this paragraph may be too terminology heavy.

Also, at this point I don't think I've introduced side-conditions, so that text itself does not actually make sense.  Which opens the door for mucho simplification.

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,673 @@
...
+layout: post
+title: "Mixing matching, mutation, and moves in Rust"
+author: Felix S. Klock II
+description: "A tour of matching and enums in Rust."
+---
+
+One of the primary goals of the Rust project is to enable safe systems
+programming. Systems programming usually implies imperative
+programming, which in turns often implies side-effects, reasoning
+about shared, aliasable state, et cetera.
```
lambda: You use the term "aliasable state" (and "unaliased" below), which readers may be unfamiliar with without a definition. I would probably go with a term like "exclusive" rather than "unaliased"; not quite sure what to replace "shared, aliasable state" with, perhaps "shared state with multiple live references" or something like that.

_posts/2015-04-17-Enums-match-mutation-and-moves.md
```diff
@@ -0,0 +1,706 @@
...
+```
+
+Note that the code above now binds `payload` by a `ref mut`-pattern;
+if it did not use a `ref` pattern, then `payload` would be bound to a
+local copy of the integer, while we want to modify the actual integer
+*in the tree itself*. Thus we need a reference to that integer.
+
+Note also that the code is able to bind `left` and `right`
+simultaneously in the `Node` arm. The compiler knows that the two
+values cannot alias, and thus it allows both `&mut`-references to live
```
lambda: One more use of "alias", I think this could be rephrased "the two references cannot refer to the same part/piece/place/value/storage" (having a hard time coming up with both a clear and correct term to use).

PR #50, opened 2015-04-21
Escape left paren so it isn't interpreted as Markdown link url

nham:
Currently this link is broken because it's interpreting the parenthetical remark "(with Nightly, Beta, and Stable channels)" as a URL.


alexcrichton:
Thanks!

PR #51, opened 2015-04-23
Rust Once, Run Everywhere

alexcrichton:
(don't merge until tomorrow!)

### DIFF
--- a/_posts/2015-04-24-FFI-and-Rust.md
+++ b/_posts/2015-04-24-FFI-and-Rust.md
@@ -1,0 +1,265 @@
+---
+layout: post
+title: "FFI and Rust"
+author: Alex Crichton
+description: "Zero-cost and safe FFI in Rust"
+---
+
+
+Rust's quest for world domination was never destined to happen overnight, so
+Rust needs to be able to interoperate with the existing world just as easily
+as it talks to itself. To solve this problem, **Rust lets you communicate with C
+APIs at no extra cost while providing strong safety guarantees**.
+
+This is also referred to as Rust's foreign function interface (FFI) and is the
+method by which Rust communicates with other programming languages. Following
+Rust's design principles, this is a **zero cost abstraction** where function
+calls between Rust and C have identical performance to C function calls. FFI
+bindings can also leverage language features such as ownership and borrowing to
+provide a **safe interface**.
+
+In this post we'll explore how to encapsulate unsafe FFI calls to C in safe,
+zero-cost abstractions by looking at some examples of interacting with C.
+Working with C is, however, just an example, as we'll also see how Rust can
+easily talk to languages like Python and Ruby just as seamlessly as C.
+
+### Talking to C
+
+First, let's start with an example of calling C code from Rust and then
+demonstrate that Rust imposes no additional overhead. Starting off simple,
+here's a C program which will simply double all the input it's given:
+
+```c
+int double_input(int input) {
+    return input * 2;
+}
+```
+
+To call this from Rust, one would write this program:
+
+```rust
+extern crate libc;
+
+extern {
+    fn double_input(input: libc::c_int) -> libc::c_int;
+}
+
+fn main() {
+    let input = 4;
+    let output = unsafe { double_input(input) };
+    println!("{} * 2 = {}", input, output);
+}
+```
+
+And that's it! You can try this out for yourself by [checking out the code on
+GitHub][rust2c] and running `cargo run` from that directory. At the source level
+we can see that there's no burden in calling an external function, and we'll see
+soon that the generated code indeed has no overhead. There are, however, a few
+subtle aspects of this Rust program so let's cover each piece in detail.
+
+[rust2c]: https://github.com/alexcrichton/rust-ffi-examples/tree/master/rust-to-c
+
+First up we see `extern crate libc`. [This crate][libc] provides many useful
+type definitions for FFI bindings when talking with C, and it is necessary
+to ensure that both C and Rust agree on the types crossing the language
+boundary.
+
+[libc]: https://crates.io/crates/libc
+
+This leads us nicely into the next part of the program:
+
+```rust
+extern {
+    fn double_input(input: libc::c_int) -> libc::c_int;
+}
+```
+
+In Rust this is a **declaration** of an externally available function. You can
+think of this along the lines of a C header file. Here's where the compiler
+learns about the inputs and outputs of the function, and you can see above that
+this matches our definition in C. Next up we have the main body of the program:
+
+```rust
+fn main() {
+    let input = 4;
+    let output = unsafe { double_input(input) };
+    println!("{} * 2 = {}", input, output);
+}
+```
+
+We see one of the crucial aspects of FFI in Rust here, the `unsafe` block. The
+compiler knows nothing about the implementation of `double_input`, so it must
+assume that memory unsafety *could* happen in this scenario. This may seem
+limiting, but Rust has just the right set of tools to allow consumers to not
+worry about `unsafe` (more on this in a moment).
+
+Now that we've seen how to call a C function from Rust, let's see if we can
+verify this claim of zero overhead. Almost all programming languages can call
+into C one way or another, but it often comes at a cost with runtime type
+conversions or perhaps some language runtime juggling. To get a handle on what
+Rust is doing, let's go straight to the assembly code of the above `main`
+function's call to `double_input`:
+
+```
+mov    $0x4,%edi
+callq  3bc30 <double_input>
+```
+
+And as before, that's it! Here we can see that calling a C function from Rust
+involves precisely one call instruction after moving the arguments into place,
+exactly the same cost as it would be in C.
+
+### Safe Abstractions
+
+One of Rust's core design principles is its emphasis on ownership, and FFI is no
+exception here. When binding a C library in Rust you not only have the benefit
+of 0 overhead, but you are also able to make it *safer* than C can! Bindings
+can leverage the ownership and borrowing principles in Rust to codify comments
+typically found in a C header about how its API should be used.
+
+For example, consider a C library for parsing a tarball. This library will
+expose functions to read the contents of each file in the tarball, probably
+something along the lines of:
+
+```c
+// Gets the data for a file in the tarball at the given index, returning NULL if
+// it does not exist. The `size` pointer is filled in with the size of the file
+// if successful.
+const char *tarball_file_data(tarball_t *tarball, unsigned index, size_t *size);
+```
+
+This function is implicitly making assumptions about how it can be used,
+however, by assuming that the `char*` pointer returned cannot outlive the input
+tarball. When bound in Rust, this API might look like this instead:
+
+```rust
+pub struct Tarball { raw: *mut tarball_t }
+
+impl Tarball {
+    pub fn file(&self, index: u32) -> Option<&[u8]> {
+        unsafe {
+            let mut size = 0;
+            let data = tarball_file_data(self.raw, index as libc::c_uint,
+                                         &mut size);
+            if data.is_null() {
+                None
+            } else {
+                Some(slice::from_raw_parts(data as *const u8, size as usize))
+            }
+        }
+    }
+}
+```
+
+Here the `*mut tarball_t` pointer is *owned by* a `Tarball`, so we already have
+rich knowledge about the lifetime of the resource. Additionally, the `file`
+method returns a **borrowed slice** whose lifetime is connected to the same
+lifetime as the source tarball itself. This is Rust's way of indicating that the
+returned data cannot outlive the tarball, statically preventing bugs that may be
+encountered when just using C.
+
+A key aspect of the Rust binding here is that it is a safe function! Although it
+has an `unsafe` implementation (due to calling an FFI function), this interface
+is safe to call and will not cause tough-to-track-down segfaults. And don't
+forget, all of this is coming at 0 cost as the raw types in C are representable
+in Rust with no extra allocations or overhead.
+
+### Talking to Rust
+
+A major feature of Rust is that it does not have a garbage collector or
+runtime, and one of the benefits of this is that Rust can be called from C with
+no setup at all. This means that the zero overhead FFI not only applies when
+Rust calls into C, but also when C calls into Rust!
+
+Let's take the example above, but reverse the roles of each language. As before,
+all the code below is [available on GitHub][c2rust]. First we'll start off with
+our Rust code:
+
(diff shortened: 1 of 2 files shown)


_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+subtle aspects in play. First off we've got our function definition with a
+`#[no_mangle]` attribute. This instructs the compiler to not mangle the symbol
+name for the function `double_input`. Rust employs name mangling similar to C++
+to ensure that libraries do not clash with one another, and this attributes
+means that you don't have to guess a symbol name like
+`double_input::h485dee7f568bebafeaa` from C.
+
+Next we've got our function definition, and the most interesting part about
+this is the keyword `extern`. This is a specialized form of specifying the [ABI
+for a function][abi-fn] which enables C to know how to call this function.
```
sfackler: We're not really teaching C how to call it, but building it in a form that C knows how to call. Maybe rephrase it to "which causes the function to work in the way C expects"?

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
+---
+layout: post
+title: "FFI and Rust"
+author: Alex Crichton
+description: "Zero-cost and safe FFI in Rust"
+---
+
+
+Rust's quest for world domination was never destined to happen overnight, so
```
brson: Yes. World domination.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+Rust needs to be able to interoperate with the the existing world just as easily
+as it talks to itself. To solve this problem, **Rust lets you communicate with C
+APIs at no extra cost while providing strong safety guarantees**.
+
+This is also referred to as Rust's foreign function interface (FFI) and is the
+method by which Rust communicates with other programming languages. In following
+Rust's design principles, this is a **zero cost abstraction** which leverages
+language features such as ownership and borrowing to provide a **safe
+interface**.
```
brson: I don't see how the FFI itself leverages ownership and borrowing.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+
+In this post we'll explore how Rust's zero-cost FFI and ability to write safe
+abstractions by looking at some examples of interacting with C. C is, however,
```
huonw: "C. C" looks a little strange.

huonw:
I'd still like to see at least a mention of destructors in terms of building good abstractions around FFI.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+just an example, as we'll also see how Rust can easily talk to languages like
+Python and Ruby just as seamlessly as C.
```
brson: First sentence includes a stray 'how' that is never connected to another clause. Maybe "In this post we'll explore how to encapsulate unsafe FFI calls to C in *safe, zero-cost abstractions*."

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Talking to C
+
+First, let's start with an example of calling C code from Rust and then take a
+look at how there is no overhead. Starting off simple, here's a C program which
```
brson: Instead of 'take a look at how' you might say 'demonstrate that'. Instead of just 'no overhead', which is somewhat ambiguous, maybe 'Rust imposes no additional performance overhead', or 'the machine code generated by Rust to call C is equivalent to that generated by a C compiler'.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+    let input = 4;
+    let output = unsafe { double_input(input) };
+    println!("{} * 2 = {}", input, output);
+}
+```
+
+And that's it! You can try this out for yourself by [checking out the code on
+GitHub][rust2c] and running `cargo run` from that directory. As you can see
+there is 0 setup necessary to call into a C function and there's no overhead on
+Rust's side when calling it. There are, however, a few subtle aspects of this
```
brson: I don't know that I 'see' there's no overhead. The compiler could be doing anything.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+fn main() {
+    let input = 4;
+    let output = unsafe { double_input(input) };
+    println!("{} * 2 = {}", input, output);
+}
+```
+
+We see one of the crucial aspects of FFI in Rust here, the `unsafe` block. The
+compiler knows nothing about the implementation of `double_input`, so it must
+assume that memory unsafety *could* happen in this scenario. This may seem
```
brson: Instead of 'memory unsafety could happen' you might embellish a colorful scenario like 'so it must assume that `double_input` may vomit nasal-demons all over your heap, which would clearly be memory-unsafe'.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+limiting, but Rust has just the right set of tools to allow consumers to not
+worry about `unsafe` (more on this in a moment).
+
+Now that we've seen how to call a C function from Rust, let's see if we can
+verify this claim of zero overhead. Almost all programming languages can call
+into C one way or another, but it often comes at a cost with runtime type
+coversions or perhaps some language runtime juggling. To get a handle on what
```
brson: 'coversions'

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+Rust is doing, let's go straight to the assembly code of the above `main`
+function:
+
+```
+mov    $0x4,%edi
+callq  3bc30 <double_input>
+```
```
brson: This isn't the complete 'assembly' of `main`. May be worth clarification.

_posts/2015-04-24-FFI-and-Rust.md
alexcrichton: Hm I was trying to get both points up here to start emphasizing them both right off the bat, but I do agree that the wording here is a little odd. I'll see what I can do about this.

_posts/2015-04-24-FFI-and-Rust.md
alexcrichton: Hm I feel like this would need a different example, however. For example if we were passing in a pointer then C holding onto it or erroneously modifying it would be easy and bad to do, but when passing in an integer there's only so many bad things can happen. I do agree though it would be nice to emphasize this a bit more.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+method returns a **borrowed slice** whose lifetime is connected to the same
+lifetime as the source tarball itself. This is Rust's way of indicating that the
+returned data cannot outlive the tarball, statically preventing bugs that may be
+encountered when just using C.
+
+A key aspect of the Rust binding here is that it is a safe function! Although it
+has an `unsafe` implementation (due to calling an FFI function), this interface
+is safe to call and will not cause tough-to-track-down segfaults. And don't
+forget, all of this is coming at 0 cost as the raw types in C are representable
+in Rust with no extra allocations or overhead.
```
brson: This safe interface imposes an extra branch over using the raw interface because it checks for a null return and the caller is going to have to duplicate that check of the Option.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Talking to Rust
+
+A major feature of Rust is that it does not have a garbage collector or heavy
+runtime, and one of the benefits of this is that Rust can be called from C with
```
brson: May not want to equivocate here and just say 'does not have a garbage collector or runtime'.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+
+Next we've got our function definition, and the most interesting part about
+this is the keyword `extern`. This is a specialized form of specifying the [ABI
+for a function][abi-fn] which enables C to know how to call this function.
+
+[abi-fn]: http://doc.rust-lang.org/reference.html#extern-functions
+
+Finally, if you [take a look at the `Cargo.toml`][cargo-toml] you'll see that
+this library is not compiled as a normal Rust library (rlib) but instead as a
+static archive called a staticlib. This enables all the relevant Rust code to be
```
brson: Maybe "which Rust calls a 'staticlib'", since there's nothing special about it.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+compiled and linked together and the final executable should run and print
+`4 * 2 = 8`.
+
+Rust's lack of a garbage collector and runtime enables this seamless transition
+from C to Rust. The external C code does not need to perform any setup on Rust's
+behalf, making the transition that much cheaper.
+
+### Beyond C
+
+Up to now we've seen how FFI in Rust has 0 overhead and how we can use Rust's
```
brson: '0' -> 'zero' I think.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+[rb2rust]: https://github.com/alexcrichton/rust-ffi-examples/tree/master/ruby-to-rust
+[js2rust]: https://github.com/alexcrichton/rust-ffi-examples/tree/master/node-to-rust
+
+A common desire for writing C code in these languages is to speed up some
+component of a library or application that's performance critical. With the
+features of Rust we've seen here, however, Rust is just as suitable for this
+sort of usage. One of Rust's first production users,
+[Skylight](https://www.skylight.io), is a great example of the benefits that
+Rust can bring even to existing applications by improving the performance and
+memory usage of their data collection agent almost instantly by just using Rust.
```
brson: This sentence reads weird to me. I read it as 'Rust improves performance and memory ...' when it is saying 'Skylight improves performance and memory'.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+
+Many programmers in languages like Python and Ruby have little systems
+programming experience, so reaching down to C is often avoided as it's tough to
+ensure that the program won't crash in a difficult-to-debug way. Rust, however,
```
brson: You might avoid ascribing any particular behavior or thinking to large groups of developers. This can probably be phrased in a way that is less accusatory of the deficiencies of Python and Ruby programs - integrating C is just hard and Rust makes it easier.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,264 @@
...
+not only brings 0 cost FFI, but *also* the same safety guarantees the the
```
brson: '0' -> 'zero' again?

_posts/2015-04-24-FFI-and-Rust.md
alexcrichton: Is there really an extra branch in optimized code though? Any caller of this C function is required to have a check, so there must be at least one. In the case of rust when you `match` the option later on, it's likely to be inlined enough (due to this small implementation) that the conditional for the match and the conditional to create the option to be unified. Which is to say I'm not sure that this actually is imposing extra overhead.

_posts/2015-04-24-FFI-and-Rust.md
alexcrichton: Very good point, I'll shuffle this around a bit.

alexcrichton:
Thanks for the comments everyone! I've pushed some minor edits here and there to address them.

_posts/2015-04-24-FFI-and-Rust.md
brson: I'm not sure either.

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,285 @@
...
+
+Rust's quest for world domination was never destined to happen overnight, so
+Rust needs to be able to interoperate with the existing world just as easily as
+it talks to itself. In particular, **Rust makes it easy to communicate with C
+APIs without overhead, and to leverage its ownership system to provide much
+stronger safety guarantees for those APIs at the same time**.
+
+In more detail, Rust's *foreign function interface* (FFI) is the way that it
+communicated with other languages. Following Rust's design principles, the FFI
+provides a **zero cost abstraction** where function calls between Rust and C
```
nikomatsakis: Nit: "zero-cost abstraction"

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,285 @@
...
+The `unsafe` block is how the programmer takes responsibility for ensuring
+safety -- you are promising that the actual call you make will not, in fact,
+violate memory safety, and thus that Rust's basic guarantees are upheld.  This
+may seem limiting, but Rust has just the right set of tools to allow consumers
+to not worry about `unsafe` (more on this in a moment).
+
+Now that we've seen how to call a C function from Rust, let's see if we can
+verify this claim of zero overhead. Almost all programming languages can call
+into C one way or another, but it often comes at a cost with runtime type
+conversions or perhaps some language runtime juggling. To get a handle on what
```
nikomatsakis: Nit: "language-runtime" (language here is an adjective that applies to runtime, not juggling)

_posts/2015-04-24-FFI-and-Rust.md
```diff
@@ -0,0 +1,285 @@
+---
+layout: post
+title: "FFI and Rust"
+author: Alex Crichton
+description: "Zero-cost and safe FFI in Rust"
+---
+
+Rust's quest for world domination was never destined to happen overnight, so
+Rust needs to be able to interoperate with the existing world just as easily as
+it talks to itself. In particular, **Rust makes it easy to communicate with C
```
nikomatsakis: Maybe s/in particular/for this reason/? I found "in particular" odd here because the structure of the sentence that follows is not parallel to the one before -- that is, this is not a refinement of a more general claim. Before you said "rust needs to be able to do this". I would expect something like, "in particular, rust needs to be able to call C functions" or something, rather than "in particular, rust makes it easy to call C functions".

PR #53, opened 2015-04-24
Enable SmartyPants for Better Typography

killercup:
Enables transforming things like "--" to "–" and "we'll" to "we’ll".

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -12,6 +12,6 @@ github_username:  rust-lang
 highlighter: pygments
 markdown: redcarpet
 redcarpet:
-  extensions: ["with_toc_data"]
+  extensions: ["with_toc_data", "smart"]
 
 root: http://blog.rust-lang.org


killercup:
I noticed that http://blog.rust-lang.org/2015/04/24/Rust-Once-Run-Everywhere.html could benefit from some typographic enhancements. This just adds the extension to Redcarpet (as seen in the [Jekyll documentation](http://jekyllrb.com/docs/configuration/#redcarpet)).

steveklabnik:
I'm afraid of things like this due to lifetimes: they shouldn't be curly. Do you know how this interacts with this?

killercup:
I would be surprised if this did anything to text in back ticks or code blocks. I'll try it locally and see what happens, though.

killercup:
@steveklabnik, lifetimes render as expected, `'a` stays `'a`. SmartyPants ignores inline code and code blocks.

steveklabnik:
Seems fine then, thank you!

PR #61, opened 2015-05-25
Replaced private email with github email

xamgore:


alexcrichton:
The emails have all been removed now, so this is taken care of. Sorry for any spam you received!

PR #71, opened 2015-10-22
Remove Caspar's email address

caspark:
I started getting a lot more spam shortly after this was published, so better late than never ;)


nikomatsakis:
Sorry about that :(

PR #89, opened 2016-04-19
add a new blog post introducing MIR

nikomatsakis:
The images don't work in this local version because of the absolute paths. You can view it here: 

http://nikomatsakis.github.io/blog.rust-lang.org/2016/04/19/MIR.html

cc @rust-lang/core @rust-lang/compiler


_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+working on a plan to change our internal compiler pipeline, as shown
+here:
+
+<center>
+![Compiler Flowchart][flow]
+</center>
+
+That is, we are introducing a new intermediate representation (IR) of
+your program that we call **MIR**: MIR stands for *middle* IR, because
+the MIR comes between the existing HIR ("high-level IR", roughly an
```
ubsan: Nit: mid-level IR, not middle IR.

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+  agree about how to translate drops, or the precise order in which `match`
+  expression arms will be tested and executed (which can get quite
+  complex). With MIR, all of that logic is centralized in MIR construction, and
+  the later passes can just rely on that.
+
+- **Raising ambitions.** In addition to being more [DRY][dry], working
+  with MIR is just plain *easier*, because it contains a much more
+  primitive set of operations than ordinary Rust. This simplification
+  enables us to do a lot of things that were forbodingly complex
+  before. We'll look at one such case in those post -- non-zeroing
```
tsion: s/those post/this post/

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+        |           [ process(elem) ] ---------+ // <-- Loop edge
+        |
+        v // <-- Break edge
+       ...
+```
+-->
+
+The idea of "simplifying" a loop into uses of `goto` (in the form of a
+CFG) is sort of an interesting one. `goto` is in fact a *more general*
+construct that `loop`, and it's something we would not want to
```
tsion: s/that `loop`/than `loop`/

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+    process(elem);
+}
+```
+
+We then transformed this for loop to explicitly invoke
+`IntoIterator::into_iter(vec)`, yielding a value `iterator`, from
+which we extract the various elements. Well, this value `iterator`
+actually has a destructor, and it will need to be freed (in this case,
+its job is to free the memory that was used by the vector `vec`; this
+memory is no longer needed, since we've finished iterator over the
```
tsion: s/iterator/iterating/

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+accurately, *overwriting*. What this means is that, if ownership of
+`data` is moved, we will overwrite the stack slot for `data` with a
+specific, distinctive bit pattern that is not a valid pointer (we used
+to use zeroes, so we usually call this zeroing, but we've since
+shifted to something different). Then, when it's time to free `data`,
+we check whether it was overwritten. (As an aside, this is actually
+roughly the same thing that the equivalent C++ code would do.)
+
+But we'd like to do better than that. What we would *like* to do is to
+using boolean flags on the stack that tell us what needs to be freed.
```
tsion: s/using/use/

nrc:
Nice post! I guess the audience is 'average Rust user'. I think if that is the case, then it could be improved slightly by being more concrete about the user-facing goals (e.g., give an example of code which currently errors, but would succeed with NLLs) and explain some of the jargon inline rather than give links.

With a few extras, I think the blog could be really interesting to an audience of 'people with an interest in compilers'. For that audience, I think the post would benefit from:

* more discussion of how MIR fits into type checking, i.e., it relies on type checking, is used for borrow checking, etc. (c.f., the first flow chart, stating why we need type info to construct the MIR as opposed to just saying where it fits into the process)
* comparison with SSA (and other options) why MIR isn't SSA and why you don't use it
* what isn't explicit in the MIR (thinking of fat pointers for example, but also other trans stuff like DPS)
* what temporaries exist in the MIR (lvalues vs rvalues vs allocas (in trans))

I think that you could add this stuff without making the article much longer and make it really interesting to a bunch more people.

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+
+-   **Faster execution time**. You may have noticed that in the new compiler
+    pipeline, *optimization* appears twice. That's no accident: previously, the
+    compiler relied solely on LLVM to perform optimizations, but with MIR, we can
+    do some *Rust-specific* optimizations before ever hitting LLVM -- or, for that
+    matter, before monomorphizing code. Rust's rich type system should provide
+    fertile ground for going beyond LLVM's optimizations.
+
+    In addition, MIR will uncork some longstanding performance improvements
+    to the code Rust generates, like ["non-zeroing" drop][rfc320].
```
nrc: Given that we dive into non-zeroing drop later on, it would be nice if you could explain what it is up front (around here), rather than just linking.

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+  drop -- but as we'll see at the end, there are already many others
+  in the pipeline.
+
+Needless to say, we're excited, and the Rust community has stepped up
+in a big way to make MIR a reality. The compiler can bootstrap and run
+its test suite using MIR, and these tests have to pass on every new
+commit. Once we're able to run [Crater][crater] with MIR enabled and
+see no regressions across the entire [crates.io](http://crates.io/)
+ecosystem, we'll turn it on by default (or, you'll forgive a terrible
+(wonderful) pun, [launch MIR into orbit][orbit]).
```
nrc: I would cut this para or replace it with something much shorter - "MIR is in the compiler, will be turned on soon-ish"

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+
+This blog post begins with an overview of MIR's design, demonstrating
+some of the ways that MIR is able to abstract away the full details of
+the Rust language. Next, we look at how MIR will help with
+implementing [non-zeroing drops][rfc320], a long-desired
+optimization. If after this post you find you are hungry for more,
+have a look at the [RFC that introduced MIR][rfc1211], or jump right
+into [the code][].
```
nrc: missing link

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+```rust
+let mut iterator = vec.into_iter();
+while let Some(elem) = iterator.next() {
+    process(elem);
+}
+```
+
+By applying this rewriting, we can remove all `for` loops, but that
+still leaves multiple kinds of loops. So next we can imagine rewrite
+all `while let` loops into a simple `loop` combined with a `match`:
```
nrc: Maybe worth mentioning that this step is done in the AST-> HIR lowering (you already mentioned the HIR in the intro)

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+compiler automatically figures out which trait the method belongs to
+based on the type of the `iterator` and the set of traits in
+scope. But if we prefer to be more explicit, we could instead invoke
+the methods in the trait directly, using function call syntax:
+
+```rust
+// Rather than `vec.into_iter()`, we are calling
+// the function `IntoIterator::into_iter`. This is
+// exactly equivalent, just more explicit.
+let mut iterator = IntoIterator::into_iter(vec);
```
nrc: I would have used UFCS syntax here, is there a reason not to?

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+When we build a control-flow graph, explicit control-flow constructs
+like the `loop` or `break` keywords disappear. Instead, they are
+expressed by paths in the graph. So a loop simply appears as a cycle
+in the graph, and the `break` keyword translates into a path out of
+that cycle. To put it another way, through the CFG, the MIR
+representation effectively has a general `goto`, and so we can express
+the `loop` using that.
+
+Here is the running example from the preivous section, expressed as a
+control-flow graph:
```
nrc: I would like to see this as pseudo code with `goto`s and labels (just this one, not all the CFGs, and as well as the diagram). Maybe I'm just weird, but for me it is easier to make the transition from code to graph with that extra step.

_posts/2016-04-19-MIR.md
eddyb: It keeps the line shorter and doesn't have semantic implications, I think the point was method call desugaring, not path resolution.

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+CFG) is sort of an interesting one. `goto` is in fact a *more general*
+construct that `loop`, and it's something we would not want to
+introduce into the language proper, because we don't want people to be
+able to write spaghetti-like-code with complex control-flow that is
+hard to read and follow later. But it's fine to have such a construct
+in MIR, because we know that it will only be used in particular ways,
+such as to express a `loop` or a `break`.
+
+### Simplifying match expressions
+
```
nrc: I feel like somewhere before here you should mention that the MIR is typed.

_posts/2016-04-19-MIR.md
eddyb: I agree, I was surprised not to see goto or even just labelled break (to introduce goto).

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+When we build a control-flow graph, explicit control-flow constructs
+like the `loop` or `break` keywords disappear. Instead, they are
+expressed by paths in the graph. So a loop simply appears as a cycle
+in the graph, and the `break` keyword translates into a path out of
+that cycle. To put it another way, through the CFG, the MIR
+representation effectively has a general `goto`, and so we can express
+the `loop` using that.
+
+Here is the running example from the preivous section, expressed as a
```
huonw: "preivous" -> "previous"

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+           |                              |
+           +----------+        +----------+
+                      |        |
+                      v        v
+                   [ post_send() ]
+                    [ drop(data) ]
+                      [ return ]
+-->
+
+We can then transform this graph by identifying each place where
```
huonw: This paragraph and the next image don't appear in the rendered version, it looks like the previous comment isn't ended correctly: the closing `-->` appears to be converted into &ndash;> (`&ndash;>`).

_posts/2016-04-19-MIR.md
michaelwoerister: I agree.

_posts/2016-04-19-MIR.md
```diff
@@ -0,0 +1,725 @@
...
+using a **control-flow graph**.
+
+### Control-flow graphs
+
+In the examples so far, I was presenting MIR as a gradual
+simplification of Rust programs into other Rust programs. In reality,
+though, MIR is based on the notion of a
+[control-flow graph (CFG)][CFG]. If you've ever used a flow-chart,
+then the concept of a control-flow graph will be pretty familiar to
+you. It's a representation of your program that exploses the
```
michaelwoerister: exploses => exposes?

nikomatsakis:
Updated to address various bits of feedback. I reorganized the flow somewhat, introduced `goto`, and tried to give a bit more detail that may interest compiler enthusiasts, but without being too overwhelming (I think the primary audience is going to be more general than that).

PR #117, opened 2016-08-11
Typo fixed in compiler output

nbigaouette:
In the _Extended error messages_ section, the compiler output has a typo.

Maybe the compiler output needs to be adapted too?


alexcrichton:
Thanks!

PR #134, opened 2016-10-30
Fixed broken URL. Fixes #111.

### LINKED ISSUE #111: broken URL from 2015/04/10
https://blog.rust-lang.org/2015/04/10/Fearless-Concurrency.html

Search for "You can even share stack frames between threads" where a link points to a now non-existent functionality.

Could remove the link, or add a footnote or something to point to the relevant crates.

nbianca:


John-Steidley:
I think that we should use the link to 1.3's version of the documentation (https://doc.rust-lang.org/1.3.0/std/thread/fn.scoped.html) which has a deprecation notice and is the most up to date version of the documentation that exists in the standard library.

brson:
Thanks!

PR #135, opened 2016-10-30
Add license

nbianca:


John-Steidley:
This is meant to only fix the license, correct? You should remove the first commit from this PR.

brson:
Thanks! I'll have to take a closer look at the status of the contributions here before applying this.

steveklabnik:
The first commit here is already merged, but since the parent repo isn't around, and @nbianca hasn't responded in a very long time, I'm going to give this a close. We *should* have a license here, but I'd prefer the standard license we use for everything else, and we'll need all 66 contributors to sign off.

Issue #144, opened 2017-01-25
Remove unmunged e-mail addresses from old blog posts


YawarRaza7349:
Newer blog posts for releases don't include the e-mail addresses for contributors, so that's good. However, they're still listed alongside the contributor names for the older blog posts. It may help reduce future spam to those e-mail addresses if they were removed from the blog posts on the website. (If not, I'd at least like to remove my own e-mail address.)

I don't mind doing this myself and making a PR, but I thought I'd get some opinions from those who run the blog before doing so.

nikomatsakis:
I think removing all of the e-mail addresses would be fine, personally, particularly if we have stopped doing it on newer releases.

YawarRaza7349:
871c2211201885eae00e7b44788af78d7cd05d01

PR #151, opened 2017-03-10
Remove e-mail addresses from old posts

YawarRaza7349:
See #144. Oldest release blog posts now match the format of the newer pre-[thanks](thanks.rust-lang.org) blog posts in terms of contributor lists. I then also removed duplicate names from the lists, after determining that said duplicate names are most likely from the same person using different e-mail addresses. If you'd like me to separate these out into two pull requests, I can do so, but for now here's just the one pull request.


steveklabnik:
Nope that's great!

PR #153, opened 2017-03-14
Add announcement post for 1.16

steveklabnik:
This isn't 100% ready to ship, as I need to run thanks still, so the number is wrong.

I wonder if it's not a bit too long, especially the diagnostic bit.

r? @rust-lang/core

### DIFF
--- a/_posts/2017-03-16-Rust-1.16.md
+++ b/_posts/2017-03-16-Rust-1.16.md
@@ -1,0 +1,317 @@
+---
+layout: post
+title: "Announcing Rust 1.16"
+author: The Rust Core Team
+---
+
+The Rust team is happy to announce the latest version of Rust, 1.16.0. Rust is a
+systems programming language focused on safety, speed, and concurrency.
+
+If you have a previous version of Rust installed, getting Rust 1.16 is as easy as:
+
+```bash
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.16.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1160-2017-03-16
+
+### What's in 1.16.0 stable
+
+The largest addition to Rust 1.16 is `cargo check`. This new subcommand should speed up the development
+workflow in many cases.
+
+What does it do? Let's take a step back and talk about how `rustc` compiles your code. Compilation has many
+"passes", that is, there are many distinct steps that the compiler takes on the road from your source code
+to producing the final binary. You can see each of these steps (and how much time and memory they take)
+by passing `-Z time-passes` to a nightly compiler:
+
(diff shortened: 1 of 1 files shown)


_posts/2017-03-16-Rust-1.16.md
```diff
@@ -0,0 +1,317 @@
...
+
+Cargo now has a [`--version --verbose` flag], mirroring `rustc`.
+
+[`--version --verbose` flag]: https://github.com/rust-lang/cargo/pull/3604
+
+Crates.io now [can show off your TravisCI or AppVeyor badges] on your crate's page.
+
+[can show off your TravisCI or Appveyor badges]: https://github.com/rust-lang/cargo/pull/3546
+
+In addition, both Cargo and crates.io [understand categories]. Unlike keywords, which are free-form,
```
aturon: This probably deserves a more prominent mention.

aturon:
Looks great! I don't think it's too long -- just plenty to talk about :-)

_posts/2017-03-16-Rust-1.16.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+However, this would only happen in certain circumstances: sometimes in local variables, and for fields in
+structs. [This now happens nearly everywhere]. When combined with [some other related improvements], this
+results in a significant improvement in these sorts of diagnostics.
+
+[This now happens nearly everywhere]: https://github.com/rust-lang/rust/pull/38927
+[some other related improvements]: https://github.com/rust-lang/rust/pull/38154
+
+Finally, `Self` can now appear [in impl headers] and [in struct expressions].
```
petrochenkov: `Self` in impl headers rarely works in 1.16
The final piece was landed in https://github.com/rust-lang/rust/pull/40008, so it will be fully available only in 1.17

_posts/2017-03-16-Rust-1.16.md
steveklabnik: should this be up at the top then, maybe?

aturon: Of this section? I'm not sure. It just felt a bit buried.

But reflecting more, I'm not sure the front-end experience is quite where we want it to make a "big splash" here, so maybe a passing mention is better.

steveklabnik: That's fair. I'm happy to move it wherever, to be honest.

_posts/2017-03-16-Rust-1.16.md
```diff
@@ -201,11 +201,6 @@ results in a significant improvement in these sorts of diagnostics.
 [This now happens nearly everywhere]: https://github.com/rust-lang/rust/pull/38927
 [some other related improvements]: https://github.com/rust-lang/rust/pull/38154
 
-Finally, `Self` can now appear [in impl headers] and [in struct expressions].
```
petrochenkov: @steveklabnik 
My comment was only about impl headers, `Self` in struct expressions (`Self { a: 10 }`) is fully supported in 1.16

PR #156, opened 2017-03-16
Disable err class in syntax highlighting

bluss:
`'a` is always highlighted as an error (unmatched quote); until this is fixed, disable error highlighting.


bluss:
I imagine no existing post makes use of the error highlighting, so I'd just like to disable it (until it learns Rust).

bluss:
The error highlighting is visible in the code snippets in https://blog.rust-lang.org/2017/03/16/Rust-1.16.html

alexcrichton:
Thanks!

PR #172, opened 2017-05-15
Shrink Rustfest Berlin image

mattico:
Resized it to match the width it has on the page, and ran it through TinyJPG.

Original: 2.33MB
Now: 104.2KB


alexcrichton:
Thanks!

PR #175, opened 2017-06-08
Fix the link to the release notes

tomprogrammer:


steveklabnik:
The way we link here is actually the convention we use, thanks though!

tomprogrammer:
No problem, I was just a bit too late. It was already fixed in https://github.com/rust-lang/blog.rust-lang.org/commit/ae87c52e083d00ff28fbea1eb676f1d9c2655269.

steveklabnik:
oh, ah ha!

PR #186, opened 2017-07-19
Release announcement for Rust 1.19

steveklabnik:
r? @rust-lang/core

### DIFF
--- a/_posts/2017-07-20-Rust-1.19.md
+++ b/_posts/2017-07-20-Rust-1.19.md
@@ -42,7 +42,9 @@ union MyUnion {
 Unions are kind of like `enum`s, but they are "untagged". Enums have a "tag"
 that stores which variant is the correct one at runtime; unions elide this tag.
 
-That means that reading or writing a union's field is unsafe:
+Since we can interpret the data held in the union using the wrong variant and
+Rust can't check this for us, that means reading or writing a union's field is
+unsafe:
 
 ```rust
 let u = MyUnion { f1: 1 };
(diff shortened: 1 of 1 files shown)


_posts/2017-07-20-Rust-1.19.md
```diff
@@ -0,0 +1,210 @@
...
+union MyUnion {
+    f1: u32,
+    f2: f32,
+}
+```
+
+Unions are kind of like `enum`s, but they are "untagged". Enums have a "tag"
+that stores which variant is the correct one at runtime; unions elide this tag.
+
+That means that reading or writing a union's field is unsafe:
```
carols10cents: I don't get why unions eliding the tag means reading or writing a union's field is unsafe exactly... This isn't really the right place to elaborate completely, but I wonder if we could make this sentence just a little bit clearer, something like (my best guess as to the reason here):

"Since we can interpret the data held in the union using the wrong variant and Rust can't check this for us, that means reading or writing a union's field is unsafe:"

carols10cents: (I'm not sure of the right thing to say here so I didn't just send in a commit for this one)

steveklabnik: I'm happy with that wording if you want to send it in 👍

steveklabnik: (I can also do it if you don't want to)

carols10cents: wooooo i got it right?!?! I can send it in :)

PR #190, opened 2017-07-31
List item does not show up on website

dns2utf8:
I noticed that the last bulletpoint of that list is empty in the generated HTML.
Since I do not know how to test / build the blog locally, I am sending this PR rather blindly.


dns2utf8:
I was able to build the jekyll page locally and found the bug:
`[some text]: some more text`
means the whole statement is a comment. With some linebreaking I was able to "escape" the comment.

steveklabnik:
Thanks!

PR #194, opened 2017-08-31
Fix sort order in example

dmit:
♪ ...ell emm enn oh PEE, kyoo ARE ess tee... ♪


steveklabnik:
uuuuuuuuuuuuuuuuuuuuuugh thank you

PR #197, opened 2017-09-05
Add clarifying sentence about stable percentage

jonathandturner:
Based on feedback from @Manishearth, add a clarifying sentence about the percentage of people on the stable compiler.


Manishearth:
r+

PR #199, opened 2017-09-14
Replace "less" with "fewer" as constraints is a countable noun

locks:
I hope I got this correction right, English likes to trick me 😁


steveklabnik:
You are 100% right and this even trips up many native speakers!

PR #205, opened 2017-10-13
intrinsic never get stable

liigo:


steveklabnik:
Thanks! I believe this will break the link though, since you changed only the title. Would you mind fixing that, squashing and rebasing? Thanks!

ashleygwilliams:
hi! this PR is quite stale, so i am going to close it. if you would like to follow up please reopen this and make the requested changes. thanks!

PR #225, opened 2018-01-05
doc.creates.io does not support https

RalfJung:
Also see https://github.com/rust-lang/crates.io/issues/621


ashleygwilliams:
thank you!

PR #240, opened 2018-03-29
speed bump -> speed boost

StyMaar:
Edit an ambiguity between “bump up in speed” and https://5.imimg.com/data5/BC/JM/MY-30112025/rubber-speed-bump-500x500.jpg


steveklabnik:
Thanks so much!

PR #245, opened 2018-05-07
First draft of 1.26 announcement

steveklabnik:
I will update with the library section in the morning.

r? @rust-lang/core

### DIFF
--- a/_posts/2018-05-10-Rust-1.26.md
+++ b/_posts/2018-05-10-Rust-1.26.md
@@ -180,7 +180,7 @@ fn foo() -> impl Iterator<Item = i32> {
 
 and be done with it. Working with [futures] is very similar.
 
-[futures]: https://github.com/rust-lang-nursery/futures-rs
+[futures]: https://crates.io/crates/futures
 
 It's important to note that sometimes trait objects are still
 what you need. You can only use `impl Trait` if your function returns
--- a/_posts/2018-05-10-Rust-1.26.md
+++ b/_posts/2018-05-10-Rust-1.26.md
@@ -531,10 +531,10 @@ See the [detailed release notes][notes] for more.
 
 ### Cargo features
 
-Cargo received two big changes this release: first, [binaries now include
-lockfiles](https://github.com/rust-lang/cargo/pull/5093), which means that
-`cargo install` can use those lockfiles when building your package. Second,
-Cargo's binary [now shares the same version as
+Cargo didn't receive many big new features this release but rather saw a steady
+stream of stability and performance improvements. Cargo should now resolve lock
+files even faster, backtrack more intelligently, and require manual `cargo
+update` invocations less. Cargo's binary [now also shares the same version as
 `rustc`](https://github.com/rust-lang/cargo/pull/5083).
 
 See the [detailed release notes][notes] for more.
--- a/_posts/2018-05-10-Rust-1.26.md
+++ b/_posts/2018-05-10-Rust-1.26.md
@@ -504,6 +504,27 @@ See the [detailed release notes][notes] for more.
 
 ### Library stabilizations
 
+We stabilized [`fs::read_to_string`](https://doc.rust-lang.org/std/fs/fn.read_to_string.html),
+a convenience over `File::open` and `io::Read::read_to_string` for easily reading an entire
+file into memory at once:
+
+```rust
+use std::fs;
+use std::net::SocketAddr;
+
+let foo: SocketAddr = fs::read_to_string("address.txt")?.parse()?;
+```
+
+You can now [format numbers as hexidecimal with `Debug`
+formatting](https://github.com/rust-lang/rust/pull/48978):
+
+```rust
+assert!(format!("{:02x?}", b"Foo\0") == "[46, 6f, 6f, 00]")
+```
+
+Trailing commas [are now supported by all macros in the standard
+library](https://github.com/rust-lang/rust/pull/48056).
+
 See the [detailed release notes][notes] for more.
 
 ### Cargo features
--- a/_posts/2018-05-10-Rust-1.26.md
+++ b/_posts/2018-05-10-Rust-1.26.md
@@ -40,7 +40,7 @@ is undergoing some final, minor copy-edits, and being prepared for print. As
 such, with this release, we are recommending the second edition over the
 first. You can [read it on
 doc.rust-lang.org](https://doc.rust-lang.org/book/second-edition/) or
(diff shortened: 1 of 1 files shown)


_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,520 @@
...
+    
+    if x % 2 == 0 {
+        Box::new(iter.filter(|x| x % 2 == 0))
+    } else {
+        Box::new(iter)
+    }
+}
+```
+
+Here, we may return a filterd iterator, or maybe not. There's two different
```
CryZe: typo: filtered

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,520 @@
...
+types that can be be returned, and so we must use a trait object.
+
+Oh, and one last thing: to make the syntax a bit more symmetrical, you can
+use `impl Trait` in argument position too. That is:
```
est31: You should point out that in this instance, impl Trait doesn't function as existential any more, but is mere sugar.

est31: Alternatively, remove the existential mention in the top. Just if you mention the technical term, do it correctly :).

Centril: (And that turbofish does not work with it..)

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,520 @@
...
+  = note: expected type `()`
+             found type `std::iter::Filter<std::iter::Map<std::vec::IntoIter<{integer}>, [closure@src/main.rs:7:14: 7:23]>, [closure@src/main.rs:8:17: 8:31]>`
+```
+
+That's a huge 'found type'. Each adapter in the chain adds a new type.
+Additionally, we have that closure in there. Previously, we'd have had
+to use a trait object here, but now we can simply do
+
+```rust
+fn foo() -> impl Iterator<Item=i32> {
```
Centril: Some spacing would be nice here so: `Item = i32` (rustfmt formats this way).

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,520 @@
...
+
+and be done with it. Working with futures is very similar.
+
+It's important to note that sometimes trait objects are still
+what you need. You can only use `impl Trait` if your function returns
+a single type; if you want to return multiple, you need dynamic dispatch.
+For example:
+
+```rust
+fn foo(x: i32) -> Box<Iterator<Item=i32>> {
```
Centril: Spacing, here too.

_posts/2018-05-10-Rust-1.26.md
steveklabnik: added a small note about this.

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,545 @@
...
+about how people learn Rust since the first book was written, and this
+version is an improvement in every way.
+
+We've shipped the draft of the second edition on the website for a while now,
+but with a disclaimer that it was a work in progress. At this point, the book
+is undergoing some final, minor copy-edits, and being prepared for print. As
+such, with this release, we are recommending the second edition over the
+first. You can [read it on
+doc.rust-lang.org](https://doc.rust-lang.org/book/second-edition/) or
+locally via `rustup doc`.
```
Mark-Simulacrum: `rustup doc --book`?

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,545 @@
...
+
+// after
+fn foo(x: impl Trait) {
+```
+
+which can look a bit nicer for short signatures.
+
+> Side note for you type theorists out there: this isn't an existential, still
+> a universal. In other words, `impl Trait` is universal in an input position, but
+> existential in an output position.
```
Mark-Simulacrum: This section seems to imply `impl Trait` is a replacement for `Box<Trait>`, but it isn't -- `Box<Trait>` is more powerful: it allows for returning both `1.0` (f32) and `2` (i32). I'm not sure it's worth mentioning this in the blog, but maybe a sentence would be good? We may eventually want to extend `impl Trait` to auto-generate impls for an implicit enum, but that's not the case currently.

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,545 @@
...
+Here, we don't know how long `s` is, so we can write the first two arms, each with
+different lengths. This also means we need a `_` term, since we aren't covering
+every possible length, nor could we!
+
+#### Speed improvements
+
+We continue to work on the speed of the compiler. We discovered that deeply
+nesting types was non-linear in some cases, and [a fix was
+implemented](https://github.com/rust-lang/rust/pull/48296). We're seeing up
+to a 12% reduction in compile times from this change, but many other smaller
```
Mark-Simulacrum: Technically, our benchmarks on perf have a maximum decrease of 17.6%... but some cases could have even larger decreases, in theory.

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,545 @@
...
+```rust
+fn foo() -> impl Iterator<Item = i32> {
+    vec![1, 2, 3]
+        .into_iter()
+        .map(|x| x + 1)
+        .filter(|x| x % 2 == 0)
+}
+```
+
+and be done with it. Working with futures is very similar.
```
sinkuu: A link to the `futures` crate would be nice.

_posts/2018-05-10-Rust-1.26.md
```diff
@@ -0,0 +1,545 @@
...
+        Some(name) => name.push_str(", world"),
+        None => (),
+    }
+}
+```
+
+the compiler will automatically borrow by mutable reference, and `name` will
+be bound as `ref mut` too.
+
+We think this will remove a signficant papercut for new and old Rustaceans
```
tatsuya6502: typo: signficant -> significant

_posts/2018-05-10-Rust-1.26.md
steveklabnik: that's covered just above

> It's important to note that sometimes trait objects are still
> what you need. You can only use `impl Trait` if your function returns
> a single type; if you want to return multiple, you need dynamic dispatch.
> For example:

PR #263, opened 2018-07-27
Improve site legibility

jonhoo:
This may be a little controversial, but I figured I'd open a PR to get a conversation started. I realize that this PR may not be merged as-is :)

The current blog.rust-lang.org styles override most user settings (like font size and page font), and also overrides font properties like letter-spacing and line-height in a number of places which makes many fonts harder to read.

This patch removes unnecessary letter-spacing and line-height settings, as well as the forced Arial/Helvetica font override, an instead uses the user's chosen font and the font's settings for letter-spacing and line-height. It also removes the global override for initial page font size, so that the user's font size settings are preserved.

The blockquote styles are also update so that the quoted text is formatted similarly to the surrounding text (no italics, and no larger font size), and with a blacker grey to make the text more legible.

The net result of the change is that the page will look less "uniform" across different browsers, OSes, and devices, but the overall aesthetic should be preserved, user preferences are respected, and legibility should generally be improved.

To take this further, all `font-size` style settings that use the `px` unit should also be changed to a relative unit like `%` or `em` so that they scale proportionally for users that have increased or decreased their browser default font size, but I'll leave that for another day.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -15,9 +15,7 @@ dl, dd, ol, ul, figure {
  */
 body {
     font-family: $base-font-family;
-    font-size: $base-font-size;
-    line-height: $base-line-height;
-    font-weight: 300;
+    font-size: 1em;
     color: $text-color;
     background-color: $background-color;
 }
@@ -110,9 +108,6 @@ blockquote {
     color: $grey-color;
     border-left: 4px solid $grey-color-light;
     padding-left: $spacing-unit / 2;
-    font-size: 18px;
-    letter-spacing: -1px;
-    font-style: italic;
 
     > :last-child {
         margin-bottom: 0;
--- a/_sass/_layout.scss
+++ b/_sass/_layout.scss
@@ -17,9 +17,8 @@
 }
 
 .site-title {
-    font-size: 26px;
-    line-height: 56px;
-    letter-spacing: -1px;
+    font-size: 150%;
+    line-height: 4rem;
     margin-bottom: 0;
     float: left;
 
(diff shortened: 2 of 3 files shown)


jonhoo:
@rust-highfive r? @GuillaumeGomez

GuillaumeGomez:
Can you add screenshots of before and after this change please? :)

jonhoo:
:+1:

Yeah, I tried to figure out who's been touching style files lately, and it seems like changes to the styles are just [really rare](https://github.com/rust-lang/blog.rust-lang.org/commits/gh-pages/_sass). It mostly seems to be @huonw back in 2015, with nearly no changes since, so I don't know that this is really anyone's responsibility at the moment..

GuillaumeGomez:
Well, let's give them a few days and if no one comes around then I'll just r+ it. Ping me in a few days.

aturon:
cc @ashleygwilliams @steveklabnik

jonhoo:
ping @GuillaumeGomez @aturon

jonhoo:
bump @GuillaumeGomez @aturon again?

GuillaumeGomez:
Well I suppose it's good to go then (a few days became a few weeks after all). Thanks!

@bors: r+

GuillaumeGomez:
Oups hold on, I think bors doesn't run here. We need someone with merge rights.

jonhoo:
gentle nudge @aturon

steveklabnik:
Sorry about taking so long here; I like it. Thanks!

PR #268, opened 2018-09-11
Announcing Rust 1.29

steveklabnik:
r? @rust-lang/core

### DIFF
--- a/_posts/2018-09-13-Rust-1.29.md
+++ b/_posts/2018-09-13-Rust-1.29.md
@@ -169,6 +169,13 @@ will now try to fix up lockfiles that have been corrupted by a `git
 merge`](https://github.com/rust-lang/cargo/pull/5831/). You can pass
 `--locked` to disable this behavior.
 
+`cargo doc` has also grown a new flag:
+[`--document-private-items`](https://github.com/rust-lang/cargo/pull/5543).  By
+default, `cargo doc` only documents public things, as the docs it produces are
+intended for end-users. But if you're working on your own crate, and you have
+internal documentation for yourself to refer to, `--document-private-items`
+will generate docs for all items, not just public ones.
+
 See the [detailed release notes][notes] for more.
 
 ## Contributors to 1.29.0
--- a/_posts/2018-09-13-Rust-1.29.md
+++ b/_posts/2018-09-13-Rust-1.29.md
@@ -1,0 +1,177 @@
+---
+layout: post
+title: "Announcing Rust 1.29"
+author: The Rust Core Team
+---
+
+The Rust team is happy to announce a new version of Rust, 1.29.0. Rust is a
+systems programming language focused on safety, speed, and concurrency.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.29.0 is as easy as:
+
+```bash
+$ rustup update stable
+```
+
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Cc @rust-lang/release

_posts/2018-09-13-Rust-1.29.md
```diff
@@ -0,0 +1,177 @@
...
+
+The 1.29 release is fairly small; Rust 1.30 and 1.31 are going to have a lot
+in them, and so much of the 1.29 cycle was spent preparing for those
+releases. The two most significant things in this release aren't even language
+features: they're new abilities that Cargo has grown, and they're both about lints.
+
+* `cargo fix` can automatically fix your code that has warnings
+* `cargo clippy` is a bunch of lints to catch common mistakes and improve your Rust code
+
+### `cargo fix`
```
ashleygwilliams: it would be amazing if these could be contributed to the cargo docs and/or link to the cargo docs.

_posts/2018-09-13-Rust-1.29.md
```diff
@@ -0,0 +1,177 @@
...
+
+## What's in 1.29.0 stable
+
+The 1.29 release is fairly small; Rust 1.30 and 1.31 are going to have a lot
+in them, and so much of the 1.29 cycle was spent preparing for those
+releases. The two most significant things in this release aren't even language
+features: they're new abilities that Cargo has grown, and they're both about lints.
+
+* `cargo fix` can automatically fix your code that has warnings
+* `cargo clippy` is a bunch of lints to catch common mistakes and improve your Rust code
```
ashleygwilliams: this is probably a product call that is not worth debating on this PR, but both of these deal with lints... we might want to specify that cargo fix won't fix cargo clippy lints (i am assuming that is tru)

Centril: Seems like this is specified on line 147-148.

pietroalbini:
The post is missing `cargo doc --document-private-items`, can you add it? We forgot to include it in the release notes too, but there is already a PR open.

petrochenkov:
[Proc macro API](https://doc.rust-lang.org/nightly/proc_macro/index.html) was also stabilized in 1.29.

Aaronepower:
@petrochenkov Talking @Mark-Simulacrum users won't actually be able to use proc macros despite being stable.

petrochenkov:
@Aaronepower 
I was talking about the library part, not `#[proc_macro]`/`#[proc_macro_attribute]` definitions.
Proc macro derives are stable and can use the API now instead of parsing strings.

Mark-Simulacrum:
In conversations with @alexcrichton it was my understanding that this cycle's stabilization with proc macros aren't quite as interesting as the next cycle's functionality for most end-users which is primarily why we're not really pushing it publicly; if that's not the case we can add it here I guess.

steveklabnik:
Yes; that's why I didn't include it. I'd rather spend time explaining it all in the next release.

_posts/2018-09-13-Rust-1.29.md
steveklabnik: the cargo docs don't have this yet. I'll put it on my to-do list.

PR #273, opened 2018-10-10
Add 1.29.2 release announcement

pietroalbini:
r? @steveklabnik 

cc @rust-lang/core @rust-lang/release


steveklabnik:
Actually, 1.29.1 didn't have one either; let's go with this for now and I'll backfill them both later.

Mark-Simulacrum:
I think we've generally omitted thanks sections historically on point releases. I don't think we should continue doing so though so we can fix it from now on.

pietroalbini:
We would need to fix the thanks site though, it doesn't show anything [for point releases](https://thanks.rust-lang.org/rust/1.27.1).

PR #281, opened 2018-10-24
Rust 1.30

steveklabnik:
r? @rust-lang/core 

@ashleygwilliams will be making me a GIF to include in the post tonight or tomorrow, but we can't forget that before a merge.


_posts/2018-10-25-Rust-1.30.md
```diff
@@ -0,0 +1,355 @@
...
+[externmacro]: https://github.com/rust-lang/rust/pull/50911/
+[Rust 1.15]: https://blog.rust-lang.org/2017/02/02/Rust-1.15.html
+
+## Module system improvements
+
+The module system has long been a pain point of new Rustaceans; several of
+its rules felt awkward in practice. These changes are the first steps we're
+taking to make the module system feel more straightforward.
+
+[`mod.rs` files are now optional][optionalmod]. Imagine we had a `foo`
```
Mark-Simulacrum: This is being dropped from the release (last minute bug found)

petrochenkov: This one is un-stabilized again in https://github.com/rust-lang/rust/pull/55315 due to [issues](https://github.com/rust-lang/rust/issues/55094).

_posts/2018-10-25-Rust-1.30.md
```diff
@@ -0,0 +1,355 @@
...
+section. There's two more. The first is that you can [`use` an extern crate
+without needing `::`][nocoloncolon], that is:
+
+```rust
+// old
+let json = ::serde_json::from_str("...");
+
+// new
+let json = serde_json::from_str("...");
+```
```
Mark-Simulacrum: This doesn't demonstrate `use` but the text indicates it should?

_posts/2018-10-25-Rust-1.30.md
```diff
@@ -0,0 +1,355 @@
...
+
+Many users found the need to move `foo.rs` into `foo/mod.rs` to be an
+unncessary, strage requirement. With the new layout, you create `src/foo` and
+put `bar.rs` in it and you're done.
+
+[optionalmod]: https://github.com/rust-lang/rust/pull/54072
+
+There's two changes to `use` as well. Well, three changes: we already
+mentioned that you can use `use` to bring macros into scope in the macros
+section. There's two more. The first is that you can [`use` an extern crate
```
petrochenkov: This not about `use`, this is about anything but `use` actually.
What happens is that extern crate names are put into prelude, so in relative paths they can be used from the whole crate, similarly to e.g. `Vec` or `Copy`.
(Paths in `use` are absolute on 2015 edition, so they are not affected by this change.)

CryZe:
I feel like the fact that libcore can be used on stable now deserves its own section (panic handling being unstable made it impossible to use stable for any actual libcore only binaries or libraries before).

_posts/2018-10-25-Rust-1.30.md
petrochenkov: I think you can mention tool attributes instead (like `#[rustfmt::skip]`), they are well hidden in https://github.com/rust-lang/rust/blob/master/RELEASES.md#misc in the release notes.

pietroalbini:
cc @rust-lang/release

_posts/2018-10-25-Rust-1.30.md
```diff
@@ -0,0 +1,355 @@
...
+1.33.
+
+See the [detailed release notes][notes] for more.
+
+### Cargo features
+
+The largest feature of Cargo in this release is that we now [have a progress
+bar!](https://github.com/rust-lang/cargo/pull/5995/)
+
+> GIF goes here
```
pietroalbini: Just a reminder to put the gif here (or maybe asciinema?).

carols10cents:
I just pushed some changes without seeing the other comments-- I was confused by the `use` section as well and I tried to clarify but I'm not sure if I succeeded.

PLEASE check my work and back out anything that's wrong ❤️

_posts/2018-10-25-Rust-1.30.md
jtgeibel: FYI, I've just opened up #55331.  The `tool_lints` feature doesn't actually land until 1.31, even though the compiler warns and recommends the new syntax.

_posts/2018-10-25-Rust-1.30.md
otavio: I think a GIF is nicer as it starts automatically, so it is more eye-catching.

dtolnay:
As part of the procedural macro stabilizations, it would be good to call out that span information on errors in generated code are transparently getting way better. For example in Serde derives (notice where the span points to):

```rust
#[derive(Serialize)]
struct Demo {
    ok: String,
    bad: std::thread::Thread,
}
```

#### Before

```console
error[E0277]: the trait bound `std::thread::Thread: _IMPL_SERIALIZE_FOR_Demo::_serde::Serialize` is not satisfied
 --> src/main.rs:3:10
  |
3 | #[derive(Serialize)]
  |          ^^^^^^^^^ the trait `_IMPL_SERIALIZE_FOR_Demo::_serde::Serialize` is not implemented for `std::thread::Thread`
  |
  = note: required by `_IMPL_SERIALIZE_FOR_Demo::_serde::ser::SerializeStruct::serialize_field`
```

#### After

```console
error[E0277]: the trait bound `std::thread::Thread: serde::Serialize` is not satisfied
 --> src/main.rs:7:5
  |
7 |     bad: std::thread::Thread,
  |     ^^^ the trait `serde::Serialize` is not implemented for `std::thread::Thread`
  |
  = note: required by `serde::ser::SerializeStruct::serialize_field`
```

The improvement comes automatically with no code changes for macros that are based on syn / quote / proc-macro2.

The libproc_macro stabilizations to enable this actually landed in 1.29 but @alexcrichton had wanted to delay drawing attention to it until the big bang stabilization of procedural macros in 1.30.

_posts/2018-10-25-Rust-1.30.md
petrochenkov: @jtgeibel 
The release notes talk about tool attributes (`#[rustfmt::skip]`) rather than tool lints (`#[allow(clippy::something)]`).
Tool attributes were indeed stabilized on 1.30.

_posts/2018-10-25-Rust-1.30.md
```diff
@@ -40,18 +40,19 @@ struct Pet {
...
 advanced macros, "attribute-like procedrual macros" and "function-like
 procedural macros."
 
-Attribute-like macros are similar to custom derive macros, but instead of
-generating code for `#[derive]`, they allow you to create new, custom
-attributes of your own. They're also more flexible; derive only works for
-structs and enums; attributes can go on other places as well, like functions.
-As an example of using an attribute-like macro, you might have something like
-this when using a web application framework:
+Attribute-like macros are similar to custom derive macros, but instead of generating code
```
petrochenkov: Wording nit: attribute macros are not "-like" attributes, they *are* attributes.

(While function-like macros are obviously not functions, so the "-like" suffix is justified there.)

_posts/2018-10-25-Rust-1.30.md
```diff
@@ -260,25 +263,29 @@ mod baz {
...
 
 * Is `a` the name of a crate? Then we're looking for `b::c` inside of it.
 * Is `a` the keyword `crate`? Then we're looking for `b::c` from the root of our crate.
 * Otherwise, we're looking for `a::b::c` from the current spot in the module hierarchy.
 
-Since these rules apply uniformly everywhere, you'll need to tweak your
-imports much less when moving code around.
+The old behavior of `use` paths always starting from the crate root still applies. But
+after making a one-time switch to the new style, these rules will apply uniformly to
+paths everywhere, and you'll need to tweak your imports much less when moving code around.
```
petrochenkov: I don't know what this paragraph is talking about, the changes stabilized in 1.30 neither reduce "import tweaking when moving code around" in any way, nor "lead to a more straightforward understanding of how paths resolve".

All of those are supposed benefits of the 2018 edition module system, that's not stabilized yet and require explicit switching to 2018 edition.
Right now (2015 edition) `crate::` in paths is supported to simplify migration to 2018 rather than anything else.

petrochenkov: A bit more constructively:
- *Non*-imports starting with a crate name require less tweaking when moving code around due to the "`::serde_json::from_str` -> `serde_json::from_str`" change.
- If `crate::` is used consistently in imports (the "still on 2015, but ready for 2018 migration" mode), then we have more straightforward *import* logic - they always start with a crate name or a special keyword.

steveklabnik: i was purely talking about moving code within submodules, not moving to or from 2015/2018, like your first bullet point

steveklabnik:
Thanks folks, I'm pretty sure I got everything here.

PR #295, opened 2018-12-04
1.31 announcement

steveklabnik:
A first draft.

r? @rust-lang/core

PR #296, opened 2018-12-05
Fix wrong code block

CYBAI:


CYBAI:
Ah, sorry... I didn't see #295 before sending this patch. I'm going to close this one and comment in that PR instead 🙇

PR #306, opened 2018-12-06
Update 2018-12-06-Rust-1.31-and-rust-2018.md

nrc:
Add something about rustup self update to fix install problems

r? @alexcrichton

### DIFF
--- a/posts/2018-12-06-Rust-1.31-and-rust-2018.md
+++ b/posts/2018-12-06-Rust-1.31-and-rust-2018.md
@@ -350,6 +350,9 @@ Work on IDE support is not finished, in particular code completion is not up
 to scratch in the RLS-based editors. However, if you mainly want support for
 types, documentation, and 'go to def', etc. then you should be happy.
 
+If you have probelms installing any of the tools with Rustup, try running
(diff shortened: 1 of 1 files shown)


posts/2018-12-06-Rust-1.31-and-rust-2018.md
```diff
@@ -350,6 +350,9 @@ Work on IDE support is not finished, in particular code completion is not up
 to scratch in the RLS-based editors. However, if you mainly want support for
 types, documentation, and 'go to def', etc. then you should be happy.
 
+If you have probelms installing any of the tools with Rustup, try running
```
Mark-Simulacrum:
```suggestion
If you have problems installing any of the tools with Rustup, try running
```

alexcrichton:
I'm not 100% sure this will actually fix the issues, so I'd personally prefer to hold off until we can get to the bottom of this perhaps? (it could be bad in the meantime to have a suggestion which doesn't work)

steveklabnik:
I'm pretty sure that this fixes issues; has anything come up in the last few days? should i merge this or not?

PR #319, opened 2018-12-07
Fix social icons

Aloso:
I changed social icons in the footer and added tooltips to them. The same modifications have already been accepted at rust-lang/www.rust-lang.org


Aloso:
I haven't figured out yet how to run the website. Could you explain or point me in the right direction?

steveklabnik:
> I haven't figured out yet how to run the website locally. Could you explain or point me in the right direction?

`cargo run` and then `python -m SimpleHTTPServer` inside the `site` directory.

Aloso:
> Your git information is different than your github account

That's a mistake. I'm not sure if I have to use my username or my name. My IDE (IntelliJ) suggests my username, but I believe git from cmd uses my name by default. The emails should match.

Aloso:
@steveklabnik I just noticed that I used the wrong email address. I hope this is not a problem!

steveklabnik:
Not at all! Thanks a ton :)

Issue #308, opened 2018-12-06
Add publish date to blog posts and add it to the front page


ocamlmycaml:
with the increasing number of posts, it'd be good to see when a blog is published in the front page, as well as in the blog post's page. It's hard to tell if an announcement is old or outdated right now.

I'm not sure how this blog works (and I plan on looking into this issue myself later today) and whether we can bring back the dates for existing posts, but at least it'd be good to have it going forward

ocamlmycaml:
Great~!

So i did something. I'm very proud of my design skills /s

Aloso:
I tried it out, it looks like this:

![blog layout](https://user-images.githubusercontent.com/15658558/49669665-5c796a00-fa62-11e8-92b0-c2dd4ba8d36d.png)

A big headline with the year is shown everywhere a new year starts in the list.

This change required a handlebars helper function and a new `bool` field in the `Post` struct. I turned the list into a `<table>`. I'll make a PR.

steveklabnik:
This has now been fixed!

PR #323, opened 2018-12-07
Add publish date and author to posts

Aloso:
Fixes issue [#308](https://github.com/rust-lang/blog.rust-lang.org/issues/308)

This is what it looks like:

blog layout

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,7 +7,7 @@ use std::{
 
 use comrak::ComrakOptions;
 
-use handlebars::Handlebars;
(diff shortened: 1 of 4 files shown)


Aloso:
I now also added publish date and author(s) to all blog entries. They are directly under the title, like before. It is formatted like

Jan. 21, 2016 • The Rust Core Team

Aloso:
The wrong metadata was a mistake.

steveklabnik:
Thank you!

PR #327, opened 2018-12-10
Bring back the RSS feed

### LINKED ISSUE #307: Bring back the RSS feed
The RSS feed is gone; I will be bringing it back, but not right this second.

sondr3:
Closes #307

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -91,8 +91,7 @@ impl Blog {
                 ..ComrakOptions::default()
             };
 
-            let contents =
-                comrak::markdown_to_html(&contents[end_of_yaml + 5..], &options);
+            let contents = comrak::markdown_to_html(&contents[end_of_yaml + 5..], &options);
 
             // finally, the url.
             let mut url = PathBuf::from(&*filename);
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -91,8 +91,7 @@ impl Blog {
                 ..ComrakOptions::default()
             };
 
-            let contents =
-                comrak::markdown_to_html(&contents[end_of_yaml + 5..], &options);
+            let contents = comrak::markdown_to_html(&contents[end_of_yaml + 5..], &options);
```
sondr3: I ran `cargo fmt` and it formatted a few lines outside of what I did.

templates/feed.hbs
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="en">
+    <generator uri="https://blog.rust-lang.org/" version="0.1.0">Rust Blog</generator>
+    <link href="https://blog.rust-lang.org/feed.xml" rel="self" type="application/atom+xml" />
```
sondr3: The URL, title and subtitle are all hard coded into the template, I could change this in the `render_feed` method if desired.

sondr3:
You can see a preview of how the feed looks here: https://www.eons.io/feed.xml (hosted on my personal website). I had to restrict the amount of posts to 10 because the feed ballooned to about 1.4Mb, which my Firefox couldn't handle.

src/main.rs
steveklabnik: that's cool! i appreciate it

templates/feed.hbs
steveklabnik: I think that for now, it being hardcoded is perfectly file. It's unclear if this code will ever be used for anything but the blog.

PR #335, opened 2018-12-13
Increase font size

### LINKED ISSUE #332: Font very small
While the font sizes on the _main page_ have been increased, they are still very small on the blog. An appropriate font size is _essential_ for comfortable reading.

Aloso:
Fixes #332 

After some complaints, the font size of the main site (www.rust-lang.org) was increased, but not of the blog. This PR makes the same change as has been done on the main site; this also means that the websites look more similar.


Aloso:
?r @steveklabnik

PR #338, opened 2018-12-14
[WIP] Add syntax highlighting using prismjs

Aloso:
We still don't have syntax highlighting on the blog, so I tried out a small javascript library called [PRISM](https://prismjs.com/), and got it working with few changes.

## Advantages

- Lightweight: ~14 KB gzipped
- Supports Rust and TOML syntax
- Seems to be pretty fast
- Understands the annotations from markdown, e.g. `<code class="language-rust">`
- The default theme looks good

## Disadvantages

- I noticed that highlighting is incorrect in few cases
- The text contrast isn't perfect, but the CSS can be adjusted

## Screenshots

e_closures

e_cargo

e_repr


steveklabnik:
Interesting! I've always used highlight.js in the past. Did the web site decide to go with this library? I'm happy to do whatever they're doing.

Aloso:
The website hasn't decided anything yet AFAIK, so I tried this out on my own.

However, I just noticed that doc.rust-lang.org uses highlight.js, so it might be better to use the same library everywhere.

lzutao:
Can we close this as of #466?

PR #346, opened 2018-12-17
fix the spelling of my weird last name

killercup:


killercup:
I kinda want to r+ this myself as I'm currently traveling with my passport *and* my ID ;)

PR #350, opened 2018-12-19
1.31.1 announcement

steveklabnik:
r? @rust-lang/release

### DIFF
--- a/posts/2018-12-20-Rust-1.31.1.md
+++ b/posts/2018-12-20-Rust-1.31.1.md
@@ -27,7 +27,27 @@ This patch release fixes a build failure on on `powerpc-unknown-netbsd` by
 way of [an update to the `libc`
 crate](https://github.com/rust-lang/rust/pull/56562) used by the compiler.
 
-Additionally, the Rust Language Server was updated to fix two critical bugs:
+Additionally, the Rust Language Server was updated to fix two critical bugs.
+First, [hovering over the type with documentation above single-line
+attributes led to 100% CPU
+usage:](https://github.com/rust-lang/rls/pull/1170)
 
-* <https://github.com/rust-lang/rls/issues/1171>
-* <https://github.com/rust-lang/rls/pull/1170>
+```rust
+/// Some documentation
+#[derive(Debug)] // Multiple, single-line
+#[allow(missing_docs)] // atributes
+pub struct MyStruct { /* ... */ }
+```
+
+[Go to definition was fixed for std types](https://github.com/rust-lang/rls/pull/1171):
+Before, using the RLS on `HashMap`, for example, tried to open this file
+
+```text
+~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/libstd/collections/hash/map.rs
+```
+
+and now RLS goes to the correct location (for Rust 1.31, note the extra `src`):
+
+```text
+~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/collections/hash/map.rs
+```
--- a/posts/2018-12-20-Rust-1.31.1.md
+++ b/posts/2018-12-20-Rust-1.31.1.md
@@ -1,0 +1,33 @@
+---
+layout: post
+title: "Announcing Rust 1.31.1"
+author: The Rust Release Team
+---
+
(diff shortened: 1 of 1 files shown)


posts/2018-12-20-Rust-1.31.1.md
```diff
@@ -0,0 +1,33 @@
...
+1.31.1][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1311-2018-12-20
+
+## What's in 1.31.1 stable
+
+This patch release fixes a build failure on on `powerpc-unknown-netbsd` by
+way of [an update to the `libc`
+crate](https://github.com/rust-lang/rust/pull/56562).
```
Mark-Simulacrum:
```suggestion
crate](https://github.com/rust-lang/rust/pull/56562) used by the compiler.
```

I feel like otherwise it might communicate slightly the wrong message (i.e., libc crate is somehow special to the compiler/distribution).

posts/2018-12-20-Rust-1.31.1.md
```diff
@@ -0,0 +1,33 @@
...
+
+Additionally, the Rust Language Server was updated to fix two critical bugs:
+
+* <https://github.com/rust-lang/rls/issues/1171>
+* <https://github.com/rust-lang/rls/pull/1170>
```
Mark-Simulacrum: I feel like we should have some words here, not just links to issues...

pietroalbini: Yeah, also because that's the main motivation for the point release. The `libc` update is just a trivial fix we included since we were doing a point release anyway.

steveklabnik: I don't know a ton about what was fixed and why it was so important, any suggestions?

pietroalbini: I can try writing something tomorrow, but I'm not too familiar with RLS either. It would be great if @nrc or @Xanewok could write a few lines on the fixes.

Xanewok: TL;DR is that some hover requests (for top-most items with doc comments? Would have to double check) caused RLS to enter an infinite loop (causing 100% cpu usage) and the other adjusts rust src path to account for src/ and thus go to definition for libstd stuff works again.

I’ll write some lines in couple of hours.

Xanewok: [Hovering over the type with documentation above single-line attributes led to 100% CPU usage:](https://github.com/rust-lang/rls/pull/1170)
```rust
/// Some documentation
#[derive(Debug)] // Multiple, single-line
#[allow(missing_docs)] // atributes
pub struct MyStruct { /* ... */ }
```

[Go to definition fixed for std types](https://github.com/rust-lang/rls/pull/1171):
Before, RLS for `HashMap` tried to open this file
```
~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/libstd/collections/hash/map.rs
```
and now RLS goes to the correct location (for Rust 1.31, note the extra `src`):
```
~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/collections/hash/map.rs
```

@alexheretic is my understanding correct for https://github.com/rust-lang/rls/pull/1170?

alexheretic: @Xanewok yep, certain hover requests could cause an infinite loop trying to extract documentation. It wasn't as common as every type, only the top type in the source file. I'd just avoid being too specific.

posts/2018-12-20-Rust-1.31.1.md
```diff
@@ -0,0 +1,53 @@
...
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.31.1][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1311-2018-12-20
+
+## What's in 1.31.1 stable
+
+This patch release fixes a build failure on on `powerpc-unknown-netbsd` by
```
pietroalbini: Double `on`.

steveklabnik: nice catch!

PR #359, opened 2019-01-16
Announcing Rust 1.32.0

steveklabnik:
r? @rust-lang/core @rust-lang/release 

I am a little unsure about the library section; it feels like there's really no cohesion so "go look at this list of functions" seems to be the best, I think, but I'm not 100% sure. Thoughts?

### DIFF
--- a/posts/2019-01-17-Rust-1.32.0.md
+++ b/posts/2019-01-17-Rust-1.32.0.md
@@ -1,0 +1,274 @@
+---
+layout: post
+title: "Announcing Rust 1.32.0"
+author: The Rust Release Team
+---
+
+The Rust team is happy to announce a new version of Rust, 1.32.0. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.32.0 is as easy as:
+
+```
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.32.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1320-2019-01-17
+
+## What's in 1.32.0 stable
+
+Rust 1.32.0 has a few quality of life improvements, switches the default
+allocator, and makes additional functions `const`. Read on for a few
+highlights, or see the [detailed release notes][notes] for additional
+information.
+
+#### The `dbg` macro
+
+First up, a small quality of life improvement. Are you a "printf debugger"? If you are, and
+you've wanted to print out some value while working on some code, you have to do this:
+
+```rust
+let x = 5;
+
+println!("{:?}", x);
+
+// or maybe even this
+println!("{:#?}", x);
+```
+
+This isn't the *largest* speed bump, but it is a lot of stuff to simply show the value of `x`.
+Additionally, there's no context here. If you have several of these `println!`s, it can be hard
+to tell which is which, unless you add your own context to each invocation, causing even more work.
+
+In Rust 1.32.0, [we've added a new macro,
+`dbg!`](https://github.com/rust-lang/rust/pull/56395/), for this purpose:
+
+```rust
+fn main() {
+    let x = 5;
+    
+    dbg!(x);
+}
+```
+
+If you run this program, you'll see:
+
+```text
+[src/main.rs:4] x = 5
+```
+
+You get the file and line number of where this was invoked, as well as the
+name and value. Additionally, `println!` prints to the standard output, so
+you really should be using `eprintln!` to print to standard error. `dbg!`
+does the right thing and goes to `stderr` by default.
+
+It even works in more complex circumstances. Consider this factorial example:
+
+```rust
+fn factorial(n: u32) -> u32 {
+    if n <= 1 {
+        n
+    } else {
+        n * factorial(n - 1)
+    }
+}
+```
+
+If we wanted to debug this, we might write it like this with `eprintln!`:
+
+```rust
+fn factorial(n: u32) -> u32 {
+    eprintln!("n: {}", n);
+
+    if n <= 1 {
+        eprintln!("n <= 1");
+
+        n
+    } else {
+        let n = n * factorial(n - 1);
+
+        eprintln!("n: {}", n);
+
+        n
+    }
+}
+```
+
+We want to log `n` on each iteration, as well as have some kind of context
+for each of the branches. We see this output for `factorial(4)`:
+
+```text
+n: 4
+n: 3
+n: 2
+n: 1
+n <= 1
+n: 2
+n: 6
+n: 24
+```
+
+This is servicable, but not particularly great. Maybe we could work on how we
+print out the context to make it more clear, but now we're not debugging our code,
+we're figuring out how to make our debugging code better.
+
+Consider this version using `dbg!`:
+
+```rust
+fn factorial(n: u32) -> u32 {
+    if dbg!(n <= 1) {
+        dbg!(1)
+    } else {
+        dbg!(n * factorial(n - 1))
+    }
+}
+```
+
+We simply wrap each of the various expressions we want to print with the macro. We
+get this output instead:
+
+```text
+[src/main.rs:3] n <= 1 = false
+[src/main.rs:3] n <= 1 = false
+[src/main.rs:3] n <= 1 = false
+[src/main.rs:3] n <= 1 = true
+[src/main.rs:4] 1 = 1
+[src/main.rs:5] n * factorial(n - 1) = 2
+[src/main.rs:5] n * factorial(n - 1) = 6
+[src/main.rs:5] n * factorial(n - 1) = 24
+[src/main.rs:11] factorial(4) = 24
+```
+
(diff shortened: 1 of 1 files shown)


posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+debugging experience! We are contining to work on support for `gdb` and
+friends as well, of course.
+
+#### `jemalloc` is removed by default
+
+Long, long ago, Rust had a large, Erlang-like runtime. We chose to use
+[jemalloc] instead of the system allocator, because it often improved
+performance over the default system one. Over time, we shed more and more of
+this runtime, and eventually almost all of it was removed, but jemalloc
+didn't. We didn't have a way to choose a custom allocator, and so we couldn't
```
dtolnay: This "didn't" is saying "jemalloc didn't removed" or maybe "jemalloc didn't was removed". You either want:

> was removed ... wasn't

or

> got removed ... didn't

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+
+In the 2018 edition, `macro_rules` macros can also use `?`, like this:
+
+```rust
+macro_rules! bar {
+    ($(a)?) => {}
+}
+```
+
+The `?` will match zero or one repitions of the pattern, similar to the
```
pietroalbini:
```suggestion
The `?` will match zero or one repetitions of the pattern, similar to the
```

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+already-existing `*` for "zero or more" and `+` for "one or more."
+
+#### Final module improvements
+
+In Rust 1.30.0, [we announced several improvements to the module
```
pietroalbini:
```suggestion
In Rust 1.31.0, [we announced several improvements to the module
```

steveklabnik: This was in 1.30 https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html#module-system-improvements

steveklabnik: We did this in 1.30 *and* 1.31. https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html#module-system-improvements hmm

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+
+The Rust team is happy to announce a new version of Rust, 1.32.0. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.32.0 is as easy as:
+
+```
+$ rustup update stable
```
nrc:
```suggestion
$ rustup update
```

Because this will do a Rustup self-update and specifying a toolchain does not (this is fixed on master, but of course you have to self-update to get the fix :-( )

steveklabnik: hmmm, so doing this will fail to install stable for anyone who doesn't have stable installed. we've always done it this way in every release post, what's changed since then?

BatmanAoD: @nrc Is a Rustup self-update required for the 1.32 toolchain update?

posts/2019-01-17-Rust-1.32.0.md
BatmanAoD: "was...wasn't" is much better than "got...didn't".

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+
+The `?` will match zero or one repetitions of the pattern, similar to the
+already-existing `*` for "zero or more" and `+` for "one or more."
+
+#### Final module improvements
+
+In Rust 1.30.0, [we announced several improvements to the module
+system](https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#module-system-changes).
+We have one last tweak landing in 1.32.0 and the 2018 edition. Nicknamed
+["uniform paths"](https://github.com/rust-lang/rust/pull/56759/), the simplest way to explain
```
BatmanAoD: The top-level comment in this PR doesn't seem very helpful; it might be better to link directly to this comment: https://github.com/rust-lang/rust/pull/56759#issuecomment-450051210

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+```rust
+enum Color { Red, Green, Blue }
+
+use Color::*;
+```
+
+This code did *not* previously compile, as `use` statements had to start with
+`super`, `self`, or `crate`. With this change, this code will work, and do
+what you probably expect: import the variants of the `Color` enum defined
+above the `use` statement.
```
BatmanAoD: Does the uniform-path stabilization _change_ the behavior of any code? The single-sentence description linked above is:

 > If import path resolves without errors, then it's resolved in exactly same way as a non-import path.

Is that summary _supposed_ to mean "more import paths now resolve without errors, and these paths are resolved in exactly the same way as non-import paths"?

BatmanAoD: Okay, then, I'd recommend taking out the "simplest way to explain is..." phrasing, and say:

 > Nicknamed ["uniform paths"](link), it permits previously invalid import path statements to be resolved exactly the same way as non-import paths. For example:
>
> (enum example code)
>
> This code did *not* (etc)..., or `crate`. Now that the compiler supports uniform paths, the `use` statement will bring all variants of the `Color` enum into scope.

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+## What's in 1.32.0 stable
+
+Rust 1.32.0 has a few quality of life improvements, switches the default
+allocator, and makes additional functions `const`. Read on for a few
+highlights, or see the [detailed release notes][notes] for additional
+information.
+
+#### The `dbg` macro
+
+First up, a small quality of life improvement. Are you a "printf debugger"? If you are, and
```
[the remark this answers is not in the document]
BatmanAoD: @ashleygwilliams On the one hand, it's great to know that there are people out there who have used `println!` but never `printf`!

...on the other hand, "printf debugger" is in quotes here because it's actually [a relatively common phrase](https://www.google.com/search?q=%22printf+debugger%22) meaning "someone who uses print statements to debug their code".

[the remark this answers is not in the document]
BatmanAoD: Actually, even more common is ["print debugger"](https://www.google.com/search?q=%22print+debugger%22), which makes no assumptions at all about your chosen language's IO story. So that's probably preferable.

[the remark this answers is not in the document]
ashleygwilliams: hey yeah i come from JS land, we call em "console.log debuggers"- think "print debugger" is great <3 (also i am a print debugger)

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+// or maybe even this
+println!("{:#?}", x);
+```
+
+This isn't the *largest* speed bump, but it is a lot of stuff to simply show the value of `x`.
+Additionally, there's no context here. If you have several of these `println!`s, it can be hard
+to tell which is which, unless you add your own context to each invocation, causing even more work.
+
+In Rust 1.32.0, [we've added a new macro,
+`dbg!`](https://github.com/rust-lang/rust/pull/56395/), for this purpose:
```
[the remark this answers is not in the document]
steveklabnik: we generally link to the stabilization PR, not its docs

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -0,0 +1,274 @@
...
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1320-2019-01-17
+
+## What's in 1.32.0 stable
+
+Rust 1.32.0 has a few quality of life improvements, switches the default
+allocator, and makes additional functions `const`. Read on for a few
+highlights, or see the [detailed release notes][notes] for additional
+information.
+
```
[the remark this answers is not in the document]
steveklabnik: Yes; the blog post does not include everything in `RELEASES.md`, as you say, they're not as important.

posts/2019-01-17-Rust-1.32.0.md
alexcrichton: What @nrc means is that `rustup update` does a self update at the end and `rustup update stable` does not. A recent change has happened to do a self-update on any toolchain update but it's not a published version yet.

This is particularly problematic with the 1.31.0 release where a self-update was needed first to get full usage of things like `rustup component add preview`. This in theory is fixed already for anyone coming from the 1.31.0 toolchain, but it will still be broken for anyone coming from a previous 1.30.0 or previous toolchain.

I'd recommend switching this to `rustup update`, they basically mean the same thing anyway

steveklabnik: I've added a small section suggesting users also update rustup

PR #365, opened 2019-01-18
Add Optional release tag

dustinknopoff:
@Mark-Simulacrum unsure the preference for writing JSON for releases to a file and the preferred layout.


src/main.rs
```diff
@@ -32,32 +32,50 @@ struct Post {
...
     url: String,
     published: String,
+    release: bool,
 }
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 struct YamlHeader {
     title: String,
     author: String,
+    #[serde(default = "default_release")]
```
Mark-Simulacrum: The default value for bools is false.

src/main.rs
```diff
@@ -258,12 +281,21 @@ impl Blog {
...
+            json!({ "posts": posts, "feed_updated":  time::now_utc().rfc3339().to_string() });
 
         self.render_template("feed.xml", "feed", data)?;
         Ok(())
     }
 
+    fn generate_releases_feed(&self) -> Result<(), Box<Error>> {
+        let posts: Vec<_> = self.posts.iter().filter(|post| post.release).collect();
+        let data =
+            json!({"releases": posts, "feed_updated": time::now_utc().rfc3339().to_string()});
```
Mark-Simulacrum: It would probably be worth making an actual struct and deriving serialize on it.

posts/2019-01-17-Rust-1.32.0.md
```diff
@@ -265,16 +264,15 @@ primitives now provide conversion functions to and from byte-arrays with
...
+- `le` - little endian
+- `be` - big endian
 
 See the [detailed release notes][notes] for more details.
 
 ### Cargo features
 
-Cargo gained [`cargo c` as an alias for `cargo
-check`](https://github.com/rust-lang/cargo/pull/6218/), and now [allows
+Cargo gained [`cargo c` as an alias for `cargo check`](https://github.com/rust-lang/cargo/pull/6218/), and now [allows
```
Mark-Simulacrum: If you could split the post edits into a separate PR that'd be great!

steveklabnik: yes, big 👍  here

src/main.rs
dustinknopoff: I was getting errors leaving default as:
```suggestion
    #[serde(default = "false")]
```
or

```suggestion
    #[serde(default = false)]
```
So implemented it as a function. Is this ok?

src/main.rs
dustinknopoff: Do you want it to be 
```rust
struct Releases {
    releases: Vec<Post>,
    fead_updated: Time
}
```

or something else/with more?

src/main.rs
Mark-Simulacrum: You just want `#[serde(default)]`.

src/main.rs
Mark-Simulacrum: Yep, that's perfect - I prefer having the struct to inline json macro calls.

dustinknopoff: I'm still unsure how to you actually want to write this to a file.

Mark-Simulacrum: Something along the lines of the following. Does that make sense?

```rust
fs::write(self.directory.join("releases.json"), serde_json::to_string(&json)?)?
```

Mark-Simulacrum:
Hm, looking at the data within the `Post` struct, it looks like this would be essentially serializing *all* of the post contents as well. I think it makes sense to extract the URL and title and just put those in the releases.json file -- to do so you'll probably want a `TitlePostView` or some better named struct that has those fields and map from `Post` to that struct, cloning or borrowing (doesn't really matter) the title and URL from the posts.

Let's also add the release tag to the 1.32 announcement within this PR (I don't think there's any need to update more than one release ago).

Mark-Simulacrum:
Thank you! This looks great; could you squash all the commits into one? After that I think we can merge.

PR #366, opened 2019-02-05
Fix typo in Rust 1.31 release post

fadel:
Fixed typo in the "Cargo features" section. The "extern crate" statement renamed "foo" to "bar", while the following examples renamed it to "baz" instead, which might be confusing.


nikomatsakis:
Thanks!

PR #370, opened 2019-02-27
First draft of 1.33 announcement

steveklabnik:

The release team is considering backing out the UB change, so I left it out.

I considered adding a section about the various matching features but the other stuff seems bigger and this is already pretty long. @Mark-Simulacrum if you feel that this is important, could you add in that stuff?

### DIFF
--- a/posts/2019-02-28-Rust-1.33.0.md
+++ b/posts/2019-02-28-Rust-1.33.0.md
@@ -1,0 +1,158 @@
+---
+layout: post
+title: "Announcing Rust 1.33.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.33.0. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.33.0 is as easy as:
+
+```console
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.33.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1330-2019-02-28
+
+## What's in 1.33.0 stable
+
+The two largest features in this release are significant improvements to
+`const fn`s, and the stabilization of a new concept: "pinning."
+
+### `const fn` improvements
+
+With `const fn`, you can [now do way more
+things!](https://github.com/rust-lang/rust/pull/57175/) Specifically:
+
+* irrefutable destructuring patterns (e.g. `const fn foo((x, y): (u8, u8)) { ... }`)
+* `let` bindings (e.g. `let x = 1;`)
+* mutable `let` bindings (e.g. `let mut x = 1;`)
+* assignment (e.g. `x = y`) and assignment operator (e.g. `x += y`)
+  expressions, even where the assignment target is a projection (e.g. a struct
+  field or index operation like `x[3] = 42`)
+* expression statements (e.g. `3;`)
+
+You're also [able to call `const unsafe fn`s inside a `const
+fn`](https://github.com/rust-lang/rust/pull/57067/), like this:
+
+```rust
+const unsafe fn foo() -> i32 { 5 }
+const fn bar() -> i32 {
+    unsafe { foo() }
+}
+```
+
+With these additions, many more functions in the standard library are able to
+be marked as `const`. We'll enumerate those in the library section below.
+
+### Pinning
+
+This release introduces a new concept for Rust programs, implemented as two
+types: the [`std::pin::Pin<P>`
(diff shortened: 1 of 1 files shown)


posts/2019-02-28-Rust-1.33.0.md
```diff
@@ -0,0 +1,158 @@
...
+
+You're also [able to call `const unsafe fn`s inside a `const
+fn`](https://github.com/rust-lang/rust/pull/57067/), like this:
+
+```rust
+const unsafe fn foo() -> i32 { 5 }
+const fn bar() -> i32 {
+    unsafe { foo() }
+}
+```
```
Mark-Simulacrum: It's not clear to me whether this is actually something we should call out as I believe `const unsafe fn` isn't actually interesting in that there's nothing unsafe that it can do yet?

cc @eddyb @oli-obk @rust-lang/compiler

posts/2019-02-28-Rust-1.33.0.md
```diff
@@ -0,0 +1,158 @@
...
+> in memory, meaning it cannot be moved elsewhere and its memory cannot be
+> deallocated until it gets dropped. We say that the pointee is "pinned".
+
+This feature will largely be used by library authors, and so we won't talk a
+lot more about the details here. Consult the docs if you're interested in
+digging into the details. However, the stabilization of this API is important
+to Rust users generally because it is a significant step forward towards a
+highly anticipated Rust feature: `async`/`await`. We're not quite there yet,
+but this stabilization brings us one step closer. You can track all of the
+necessary features at [https://areweasyncyet.rs/](https://areweasyncyet.rs/).
```
pietroalbini:
```suggestion
necessary features at [areweasyncyet.rs](https://areweasyncyet.rs/).
```

posts/2019-02-28-Rust-1.33.0.md
cramertj: You can't e.g. dereference raw pointers, but it's certainly possible to call functions which have internal `unsafe` invariants (think `str::from_utf8_unchecked`).

Mark-Simulacrum: Ah, okay, that makes sense. Seems fine to include as such.

Centril:
I realize it might make the blog post long, but I think the match improvements (least the exhaustiveness change + `if let p | q = e { ... }` and arbitrary_self_types matter to the average Rustacean (the changes are not niche) so I'd try to mention them.

Mark-Simulacrum:
@steveklabnik and I discussed that a bit and we came to the conclusion that the improvements are mostly not important enough to be called out explicitly -- and they are mentioned in the release notes prominently.

I think the receivers fall in a similar boat -- if you need them, you'll find out about them.

I don't feel strongly at all though so if you want to suggest some text we can include it.

Centril:
@Mark-Simulacrum  I'll think it over; what's our deadline for merging this?

Mark-Simulacrum:
Approximately 4pm UTC tomorrow is when it needs to be fully ready; it'll be merged between 5-7pm.

carols10cents:
Coinciding with this release is when we're turning on the requirement of having a verified email to be able to publish to crates.io-- it's actually going to start at 2019-03-01 00:00 UTC. Should this be called out explicitly in the release notes even though it's not actually part of the release? I'm ok with "no" being an answer, in that case I'll probably make a URLO post or something :)

Centril:
@carols10cents I think people won't care about the distinction and this feels like an important PSA so I think it's a good idea to include a bit of text about it. I had personally forgot to check that I'm in the clear so I appreciate the reminder.

posts/2019-02-28-Rust-1.33.0.md
```diff
@@ -0,0 +1,158 @@
...
+[`Duration::as_micros`]: https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_micros
+[`Duration::as_nanos`]: https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_nanos
+
+See the [detailed release notes][notes] for more details.
+
+### Cargo features
+
+[Cargo should now rebuild a crate if a file was modified during the initial
+build.](https://github.com/rust-lang/cargo/pull/6484/)
+
```
carols10cents: @steveklabnik here's some suggested text for the verified email requirement, feel free to edit as you see fit! ❤️ 

```suggestion
### Crates.io 

[As previously announced](https://users.rust-lang.org/t/a-verified-email-address-will-be-required-to-publish-to-crates-io-starting-on-2019-02-28/22425), coinciding with this release, crates.io will require that you have a verified email address to publish. Starting at 2019-03-01 00:00 UTC, if you don't have a verified email address and run `cargo publish`, you'll get an error.

This ensures we can comply with DMCA procedures. If you haven't heeded the warnings cargo printed during the last release cycle, head on over to [crates.io/me](https://crates.io/me) to set and verify your email address. This email address will never be displayed publicly and will only be used for crates.io operations.

```

carols10cents:
@steveklabnik I think i messed up where i added the suggested content so I just pushed another commit to this branch to do what I meant.

PR #376, opened 2019-04-12
Fetch social images via HTTPS from www.rust-lang.org.

hannob:
I noticed that accessing the rust blog would cause HTTP requests in firefox due to the referenced opengraph image (I believe Firefox accesses them due to activity stream).

With the goal of an HTTPS web in mind I think it'd be good to convert these to HTTPS. The normal rust page (www.rust-lang.org) references the same images from its own host via HTTPS, so I guess changing it to that URL should suffice.


pietroalbini:
Thanks!

PR #378, opened 2019-04-23
Improve typography of 2019 roadmap post

wezm:
The double dashes `--` kind of stuck out to me so I thought I'd tidy them up and also added some typographic quotation marks at the same time.


Manishearth:
Thanks!

(I could swear markdown automatically handled these, but ICBW)

wezm:
> (I could swear markdown automatically handled these, but ICBW)

Some Markdown implementations do, but not all unfortunately.

PR #394, opened 2019-07-03
Blog post; 1.36.0 announcement

Centril:

Written in collaboration with @pietroalbini and @Mark-Simulacrum.

### DIFF
--- a/posts/2019-07-04-Rust-1.36.0.md
+++ b/posts/2019-07-04-Rust-1.36.0.md
@@ -1,0 +1,219 @@
+---
+layout: post
+title: "Announcing Rust 1.36.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.36.0.
+Rust is a programming language that is empowering everyone to build reliable and efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust 1.36.0 is as easy as:
+
+```console
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the appropriate page on our website,
+and check out the [detailed release notes for 1.36.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1360-2019-07-04
+
+## What's in 1.36.0 stable
+
+This release brings many changes, including the stabilization of the [`Future`] trait,
+the [`alloc`][alloc-crate] crate, the [`MaybeUninit<T>`] type, [NLL for Rust 2015][felix-blog],
+a new `HashMap<K, V>` implementation, and [`--offline`] support in Cargo.
+Read on for a few highlights, or see the [detailed release notes][notes] for additional information.
+
+### The [`Future`] is here!
+
+[`Future`]: https://doc.rust-lang.org/std/future/trait.Future.html
+[pr-future]: https://github.com/rust-lang/rust/pull/59739
+
+In Rust 1.36.0 the long awaited [`Future`] trait has been [stabilized][pr-future]!
+
+With this stabilization, we hope to give important crates, libraries,
+and the ecosystem time to prepare for `async` / `.await`,
+which we'll tell you more about in the future.
+
+### The [`alloc`][alloc-crate] crate is stable
+
+[alloc-crate]: https://doc.rust-lang.org/alloc/index.html
+
+Before 1.36.0, the standard library consisted of the crates `std`, `core`, and `proc_macro`.
+The `core` crate provided core functionality such as `Iterator` and `Copy`
+and could be used in `#![no_std]` environments since it did not impose any requirements.
+Meanwhile, the `std` crate provided types like `Box<T>` and OS functionality
+but required a global allocator and other OS capabilities in return.
+
+Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`,
+are now available in the `alloc` crate. The `std` crate then re-exports these parts.
+While `#![no_std]` binaries using `alloc` still require nightly Rust,
+`#![no_std]` library crates can use the `alloc` crate in stable Rust.
+Meanwhile, normal binaries, without `#![no_std]`, can depend on such library crates.
+We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries
+prior to stabilizing support for `#![no_std]` binaries using `alloc`.
+
+If you are the maintainer of a library that only relies on some allocation primitives to function,
+consider making your library `#[no_std]` compatible by using the following at the top of your `lib.rs` file:
+
+```rust
+#![no_std]
+
+extern crate alloc;
+
+use alloc::vec::Vec;
+```
+
+### [`MaybeUninit<T>`] instead of [`mem::uninitialized`]
+
+[`MaybeUninit<T>`]: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html
+[`mem::uninitialized`]: https://doc.rust-lang.org/std/mem/fn.uninitialized.html
+[gankro-blog]: https://gankro.github.io/blah/initialize-me-maybe/
+[pr-60445]: https://github.com/rust-lang/rust/pull/60445
+
+In previous releases of Rust, the [`mem::uninitialized`] function has allowed you to bypass Rust's
+initialization checks by pretending that you've initialized a value at type `T` without doing anything.
+One of the main uses of this function has been to lazily allocate arrays.
+
+However, [`mem::uninitialized`] is an incredibly dangerous operation that essentially
+cannot be used correctly as the Rust compiler assumes that values are properly initialized.
+For example, calling `mem::uninitialized::<bool>()` causes *instantaneous __undefined behavior__*
+as, from Rust's point of view, the uninitialized bits are neither `0` (for `false`)
+nor `1` (for `true`) -- the only two allowed bit patterns for `bool`.
+
+To remedy this situation, in Rust 1.36.0, the type [`MaybeUninit<T>`] has been [stabilized][pr-60445].
+The Rust compiler will understand that it should not assume that a [`MaybeUninit<T>`] is a properly initialized `T`.
+Therefore, you can do gradual initialization more safely and eventually use `.assume_init()`
+once you are certain that `maybe_t: MaybeUninit<T>` contains an initialized `T`.
+
+As [`MaybeUninit<T>`] is the safer alternative, starting with Rust 1.38,
+the function [`mem::uninitialized`] will be deprecated.
+
+To find out more about uninitialized memory, [`mem::uninitialized`],
+and [`MaybeUninit<T>`], read [Alexis Beingessner's blog post][gankro-blog].
+The standard library also contains extensive documentation about [`MaybeUninit<T>`].
+
+### NLL for Rust 2015
+
+[nll-2018]: https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#non-lexical-lifetimes
+[soundness]: https://en.wikipedia.org/wiki/Soundness
+[felix-blog]: http://blog.pnkfx.org/blog/2019/06/26/breaking-news-non-lexical-lifetimes-arrives-for-everyone/
+[crater-nll]: https://github.com/rust-lang/rust/issues/60680#issuecomment-495089654
+
+[In the announcement for Rust 1.31.0][nll-2018], we told you about NLL (Non-Lexical Lifetimes),
+an improvement to the language that makes the borrow checker smarter and more user friendly.
+For example, you may now write:
+
+```rust
+fn main() {
+    let mut x = 5;
+    let y = &x;
+    let z = &mut x; // This was not allowed before 1.31.0.
+}
+```
+
+In 1.31.0 NLL was stabilized only for Rust 2018,
+with a promise that we would backport it to Rust 2015 as well.
+With Rust 1.36.0, we are happy to announce that we have done so! NLL is now available for Rust 2015.
+
+With NLL on both editions, we are closer to removing the old borrow checker.
+However, the old borrow checker unfortunately accepted some [unsound][soundness] code it should not have.
+As a result, NLL is currently in a "migration mode" wherein we will emit warnings instead
+of errors if the NLL borrow checker rejects code the old AST borrow checker would accept.
+Please see [this list][crater-nll] of public crates that are affected.
+
+To find out more about NLL, MIR, the story around fixing soundness holes,
+and what you can do about the warnings if you have them, read [Felix Klock's blog post][felix-blog].
+
+### A new [`HashMap<K, V>`] implementation
+
+[`hashbrown`]: https://crates.io/crates/hashbrown
+[`HashMap<K, V>`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
+[pr-hashbrown]: https://github.com/rust-lang/rust/pull/58623
+[SwissTable]: https://abseil.io/blog/20180927-swisstables
+[pr-hashbrown-perf]: https://perf.rust-lang.org/compare.html?start=b57fe74a27590289fd657614b8ad1f3eac8a7ad2&end=abade53a649583e40ed07c26ee10652703f09b58&stat=wall-time
+
+In Rust 1.36.0, the `HashMap<K, V>` implementation has been [replaced][pr-hashbrown]
+with the one in the [`hashbrown`] crate which is based on the [SwissTable] design.
+While the interface is the same, the `HashMap<K, V>` implementation is now
+[faster on average][pr-hashbrown-perf] and has lower memory overhead.
+Note that unlike the `hashbrown` crate,
+the implementation in `std` still defaults to the SipHash 1-3 hashing algorithm.
+
+### [`--offline`] support in Cargo
+
+[`--offline`]: https://doc.rust-lang.org/cargo/commands/cargo-build.html#cargo_build_manifest_options
+[`cargo fetch`]: https://doc.rust-lang.org/cargo/commands/cargo-fetch.html
+[nrc-blog]: https://www.ncameron.org/blog/cargo-offline/
+[relnotes-cargo]: https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-136-2019-07-04
+
+During most builds, Cargo doesn't interact with the network.
+Sometimes, however, Cargo has to.
+Such is the case when a dependency is added and the latest compatible version needs to be downloaded.
+At times, network access is not an option though, for example on an airplane or in isolated build environments.
+
+In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`].
+The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies.
+When the required crates are not available offline, and a network access would be required,
+Cargo will exit with an error.
+To prepopulate the local cache in preparation for going offline,
+use the [`cargo fetch`] command, which downloads all the required dependencies for a project.
+
+To find out more about [`--offline`] and [`cargo fetch`], read [Nick Cameron's blog post][nrc-blog].
+
+For information on other changes to Cargo, see the [detailed release notes][relnotes-cargo].
+
+### Library changes
+
+[`dbg!`]: https://doc.rust-lang.org/std/macro.dbg.html
+
+The [`dbg!`] macro now supports multiple arguments.
+
+Additionally, a number of APIs have been made `const`:
+
+[`Layout::from_size_align_unchecked`]: https://doc.rust-lang.org/core/alloc/struct.Layout.html#method.from_size_align_unchecked
+[`mem::needs_drop`]: https://doc.rust-lang.org/std/mem/fn.needs_drop.html
+[`NonNull::dangling`]: https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.dangling
+[`NonNull::cast`]: https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.cast
+
+- [`Layout::from_size_align_unchecked`]
+- [`mem::needs_drop`]
+- [`NonNull::dangling`]
+- [`NonNull::cast`]
+
+New APIs have become stable, including:
+
+[`Iterator::copied`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.copied
+[`VecDeque::rotate_left`]: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.rotate_left
+[`VecDeque::rotate_right`]: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.rotate_right
+[`BorrowMut<str> for String`]: https://github.com/rust-lang/rust/pull/60404
+[`str::as_mut_ptr`]: https://doc.rust-lang.org/std/primitive.str.html#method.as_mut_ptr
+[`pointer::align_offset`]: https://doc.rust-lang.org/std/primitive.pointer.html#method.align_offset
+[`Read::read_vectored`]: https://doc.rust-lang.org/std/io/trait.Read.html#method.read_vectored
+[`Write::write_vectored`]: https://doc.rust-lang.org/std/io/trait.Write.html#method.write_vectored
+[`task::Waker`]: https://doc.rust-lang.org/std/task/struct.Waker.html
+[`task::Poll`]: https://doc.rust-lang.org/std/task/enum.Poll.html
+
+- [`task::Waker`] and [`task::Poll`]
+- [`VecDeque::rotate_left`] and [`VecDeque::rotate_right`]
+- [`Read::read_vectored`] and [`Write::write_vectored`]
+- [`Iterator::copied`]
+- [`BorrowMut<str> for String`]
+- [`str::as_mut_ptr`]
+
+Other library changes are available in the [detailed release notes][notes].
+
+### Other changes
+
+[relnotes-clippy]: https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-136
+
+Detailed 1.36.0 release notes are available for [Rust][notes],
+[Cargo][relnotes-cargo], and [Clippy][relnotes-clippy].
+
+## Contributors to 1.36.0
+
+Many people came together to create Rust 1.36.0. We couldn't have done it
+without all of you. [Thanks!](https://thanks.rust-lang.org/rust/1.36.0/)


posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+If you don't have it already, you can [get `rustup`][install] from the appropriate page on our website, and check out the [detailed release notes for 1.36.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1360-2019-07-04
+
+## What's in 1.36.0 stable
+
+This release brings many changes, including the stabilization of the [`Future`] trait, the [`alloc`][alloc-crate] crate, the [`MaybeUninit<T>`] type, [NLL on 2015][felix-blog], a new `HashMap<K, V>` implementation, and [`--offline`] support in Cargo. Read on for a few highlights, or see the [detailed release notes][notes] for additional information.
+
+### The [`Future`] is here!
```
Centril: r? @cramertj

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+[`Future`]: https://doc.rust-lang.org/std/future/trait.Future.html
+[pr-future]: https://github.com/rust-lang/rust/pull/59739
+
+In Rust 1.36.0 the long awaited [`Future`] trait has been [stabilized][pr-future]!
+
+With this stabilization, we hope to give important crates, libraries, and  the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
+
+### The [`alloc`][alloc-crate] crate is stable
```
Centril: r? @SimonSapin

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+```rust
+#![no_std]
+
+extern crate alloc;
+
+use alloc::vec::Vec;
+```
+
+### [`MaybeUninit<T>`] instead of [`mem::uninitialized`]
```
Centril: r? @RalfJung

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+However, [`mem::uninitialized`] is an incredibly dangerous operation that essentially cannot be used correctly as the Rust compiler assumes that values are properly initialized. For example, calling `mem::uninitialized::<bool>()` causes *instantaneous __undefined behavior__* as the random bits in memory could be something other than `0` and `1`.
+
+To remedy this situation, in Rust 1.36.0, the type [`MaybeUninit<T>`] has been [stabilized][pr-60445]. The Rust compiler will understand that it should not assume that a [`MaybeUninit<T>`] is a properly initialized `T`. Therefore, you can do gradual initialization more safely and eventually use `.assume_init()` once you are certain that `maybe_t: MaybeUninit<T>` contains an initialized `T`.
+
+As [`MaybeUninit<T>`] is the safer alternative, starting with Rust 1.38, the function [`mem::uninitialized`] will be deprecated.
+
+To read more about uninitialized memory, [`mem::uninitialized`], and [`MaybeUninit<T>`], read [Alexis Beingessner's blog post][gankro-blog] about the subject. The standard library also contains extensive documentation about [`MaybeUninit<T>`].
+
+### NLL on Rust 2015
```
Centril: r? @rust-lang/wg-compiler-nll

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+[`hashbrown`]: https://crates.io/crates/hashbrown
+[`HashMap<K, V>`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
+[pr-hashbrown]: https://github.com/rust-lang/rust/pull/58623
+[SwissTable]: https://abseil.io/blog/20180927-swisstables
+[pr-hashbrown-perf]: https://perf.rust-lang.org/compare.html?start=b57fe74a27590289fd657614b8ad1f3eac8a7ad2&end=abade53a649583e40ed07c26ee10652703f09b58&stat=wall-time
+
+In Rust 1.36.0, the `HashMap<K, V>` implementation has been [replaced][pr-hashbrown] with the one in the [`hashbrown`] crate which is based on the [SwissTable] design. While the interface is the same, the `HashMap<K, V>` implementation is now on [average faster][pr-hashbrown-perf] and has lower memory overhead. Note that unlike the `hashbrown` crate, the implementation in `std` continues to use SipHash 1-3.
+
+### [`--offline`] support in Cargo
```
Centril: r? @ehuss

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+With this stabilization, we hope to give important crates, libraries, and  the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
+
+### The [`alloc`][alloc-crate] crate is stable
+
+[alloc-crate]: https://doc.rust-lang.org/alloc/index.html
+
+Before 1.36.0, the standard library consisted of the crates `std`, `core`, and `proc_macro`. The `core` crate provided core functionality such as `Iterator` and `Copy` and could be used in `#![no_std]` environments since it did not impose any requirements. Meanwhile, the `std` crate provided types like `Box<T>` and OS functionality but required a global allocator and other OS capabilities in return.
+
+Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. As a result, `#![no_std]` applications defining an allocator can use libraries that depend on the `alloc` crate.  While `#![no_std]` applications still require nightly, our hope is that a `#![no_std]` compatible ecosystem of libraries can be built up meanwhile.
```
BatmanAoD: The bit about requiring nightly is confusing to me. Is it correct that as of this release, `#![no_std]` can be used in `lib` crates but not executable crates? That may be worth calling out explicitly.

posts/2019-07-04-Rust-1.36.0.md
BatmanAoD: I think "for" is more appropriate than "on"; I'll go ahead and submit suggestions for this change in verbiage throughout the post, so be sure to either accept or reject all of them together.

```suggestion
### NLL for Rust 2015
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+To read more about uninitialized memory, [`mem::uninitialized`], and [`MaybeUninit<T>`], read [Alexis Beingessner's blog post][gankro-blog] about the subject. The standard library also contains extensive documentation about [`MaybeUninit<T>`].
+
+### NLL on Rust 2015
+
+[nll-2018]: https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#non-lexical-lifetimes
+[soundness]: https://en.wikipedia.org/wiki/Soundness
+[felix-blog]: http://blog.pnkfx.org/blog/2019/06/26/breaking-news-non-lexical-lifetimes-arrives-for-everyone/
+[crater-nll]: https://github.com/rust-lang/rust/issues/60680#issuecomment-495089654
+
+[Previously in Rust 1.31.0][nll-2018] we told you about NLL (Non-Lexical Lifetimes), an improvement to the language that makes the borrow checker smarter and more user friendly. For example, you may now write:
```
BatmanAoD:
```suggestion
[In the announcement for Rust 1.31.0][nll-2018], we told you about NLL (Non-Lexical Lifetimes), an improvement to the language that makes the borrow checker smarter and more user friendly. For example, you may now write:
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+```rust
+fn main() {
+    let mut x = 5;
+    let y = &x;
+    let z = &mut x; // This was not allowed before 1.31.0.
+}
+```
+
+In 1.31.0 NLL was stabilized for Rust 2018 only with a promise that we would back port it to Rust 2015 as well. With Rust 1.36.0, we are happy to announce that we have done so! NLL is now available in Rust 2015 as well.
```
BatmanAoD:
```suggestion
In 1.31.0 NLL was stabilized only for Rust 2018, with a promise that we would backport it to Rust 2015 as well. With Rust 1.36.0, we are happy to announce that we have done so! NLL is now available for Rust 2015.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+With NLL on both editions, we are closer to removing the old borrow checker. However, the old borrow checker unfortunately accepted some [unsound][soundness] code it should not have. As a result, NLL is currently in a "migration mode" wherein we will emit warnings instead of errors if the NLL borrow checker rejects code the old AST borrow checker would accept. Please see [this list][crater-nll] to see if one of your crates are affected.
```
BatmanAoD: Private code may be affected, and non-owners may be indirectly affected.

```suggestion
With NLL on both editions, we are closer to removing the old borrow checker. However, the old borrow checker unfortunately accepted some [unsound][soundness] code it should not have. As a result, NLL is currently in a "migration mode" wherein we will emit warnings instead of errors if the NLL borrow checker rejects code the old AST borrow checker would accept. Please see [this list][crater-nll] of public crates that are affected.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### A new `HashMap<K, V>` implementation
+
+[`hashbrown`]: https://crates.io/crates/hashbrown
+[`HashMap<K, V>`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
+[pr-hashbrown]: https://github.com/rust-lang/rust/pull/58623
+[SwissTable]: https://abseil.io/blog/20180927-swisstables
+[pr-hashbrown-perf]: https://perf.rust-lang.org/compare.html?start=b57fe74a27590289fd657614b8ad1f3eac8a7ad2&end=abade53a649583e40ed07c26ee10652703f09b58&stat=wall-time
+
+In Rust 1.36.0, the `HashMap<K, V>` implementation has been [replaced][pr-hashbrown] with the one in the [`hashbrown`] crate which is based on the [SwissTable] design. While the interface is the same, the `HashMap<K, V>` implementation is now on [average faster][pr-hashbrown-perf] and has lower memory overhead. Note that unlike the `hashbrown` crate, the implementation in `std` continues to use SipHash 1-3.
```
BatmanAoD:
```suggestion
In Rust 1.36.0, the `HashMap<K, V>` implementation has been [replaced][pr-hashbrown] with the one in the [`hashbrown`] crate which is based on the [SwissTable] design. While the interface is the same, the `HashMap<K, V>` implementation is now [faster on average][pr-hashbrown-perf] and has lower memory overhead. Note that unlike the `hashbrown` crate, the implementation in `std` continues to use the SipHash 1-3 hashing algorithm.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### [`--offline`] support in Cargo
+
+[`--offline`]: https://doc.rust-lang.org/cargo/commands/cargo-build.html#cargo_build_manifest_options
+[`cargo fetch`]: https://doc.rust-lang.org/cargo/commands/cargo-fetch.html
+
+During most builds, Cargo doesn't interact with the network. Sometimes, however, Cargo has to. Such is the case when a dependency is added and the latest compatible version needs to be downloaded. At times, network access is not an option though, for example on an airplane or in isolated build environments.
+
+In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required ones are not available offline, and a network access would be required, Cargo will error out. To prepopulate the local cache, it's possible to use the [`cargo fetch`] command, which downloads all the required dependencies beforehand.
```
nrc:
```suggestion
In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required ones are not available offline, and a network access would be required, Cargo will error out. To prepopulate the local cache, it's possible to use the [`cargo fetch`] command, which downloads all the required dependencies for a project.
```

posts/2019-07-04-Rust-1.36.0.md
nrc: r+ with the suggested change

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+[pr-hashbrown-perf]: https://perf.rust-lang.org/compare.html?start=b57fe74a27590289fd657614b8ad1f3eac8a7ad2&end=abade53a649583e40ed07c26ee10652703f09b58&stat=wall-time
+
+In Rust 1.36.0, the `HashMap<K, V>` implementation has been [replaced][pr-hashbrown] with the one in the [`hashbrown`] crate which is based on the [SwissTable] design. While the interface is the same, the `HashMap<K, V>` implementation is now on [average faster][pr-hashbrown-perf] and has lower memory overhead. Note that unlike the `hashbrown` crate, the implementation in `std` continues to use SipHash 1-3.
+
+### [`--offline`] support in Cargo
+
+[`--offline`]: https://doc.rust-lang.org/cargo/commands/cargo-build.html#cargo_build_manifest_options
+[`cargo fetch`]: https://doc.rust-lang.org/cargo/commands/cargo-fetch.html
+
+During most builds, Cargo doesn't interact with the network. Sometimes, however, Cargo has to. Such is the case when a dependency is added and the latest compatible version needs to be downloaded. At times, network access is not an option though, for example on an airplane or in isolated build environments.
```
BatmanAoD: Prior to this change, do _all_ new dependencies cause network access, at least in order to determine whether the locally cached version (if any) is up to date? If so, I would leave out the "latest compatible version" clause.

```suggestion
During most builds, Cargo doesn't interact with the network. Sometimes, however, it does need network access, for instance when a new dependency is added. But in some environments, such as an airplane or an isolated build environment, network access is not an option.
```

posts/2019-07-04-Rust-1.36.0.md
pnkfelix: r+

posts/2019-07-04-Rust-1.36.0.md
BatmanAoD:
```suggestion
In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required crates are not available offline, and a network access would be required, Cargo will exit with an error. To prepopulate the local cache in preparation for going offline, use the [`cargo fetch`] command, which downloads all the required dependencies.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required ones are not available offline, and a network access would be required, Cargo will error out. To prepopulate the local cache, it's possible to use the [`cargo fetch`] command, which downloads all the required dependencies beforehand.
+
+[relnotes-cargo]: https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-136-2019-07-04
+
+For more information on other changes to Cargo, see the [detailed release notes][relnotes-cargo].
```
BatmanAoD:
```suggestion
For information on other changes to Cargo, see the [detailed release notes][relnotes-cargo].
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Library changes
+
+[`dbg!`]: https://doc.rust-lang.org/std/macro.dbg.html
+
+The [`dbg!`] macro now supports multiple arguments.
+
+Moreover, a number of APIs have been made `const`:
```
BatmanAoD:
```suggestion
A number of APIs have been made `const`:
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+- [`BorrowMut<str> for String`]
+- [`str::as_mut_ptr`]
+
+Other library changes are available in the [detailed release notes][notes].
+
+### Other changes
+
+[relnotes-clippy]: https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-136
+
+There are other changes in the Rust 1.36 release: check out what changed in [Rust][notes], [Cargo][relnotes-cargo], and [Clippy][relnotes-clippy].
```
BatmanAoD:
```suggestion
Detailed 1.36.0 release notes are available for [Rust][notes], [Cargo][relnotes-cargo], and [Clippy][relnotes-clippy].
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### [`--offline`] support in Cargo
+
+[`--offline`]: https://doc.rust-lang.org/cargo/commands/cargo-build.html#cargo_build_manifest_options
+[`cargo fetch`]: https://doc.rust-lang.org/cargo/commands/cargo-fetch.html
+
+During most builds, Cargo doesn't interact with the network. Sometimes, however, Cargo has to. Such is the case when a dependency is added and the latest compatible version needs to be downloaded. At times, network access is not an option though, for example on an airplane or in isolated build environments.
+
+In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required ones are not available offline, and a network access would be required, Cargo will error out. To prepopulate the local cache, it's possible to use the [`cargo fetch`] command, which downloads all the required dependencies for a project.
```
BatmanAoD: In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required crates are not available offline, and a network access would be required, Cargo will exit with an error. To prepopulate the local cache in preparation for going offline, use the [`cargo fetch`] command, which downloads all the required dependencies for a project.

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+This release brings many changes, including the stabilization of the [`Future`] trait, the [`alloc`][alloc-crate] crate, the [`MaybeUninit<T>`] type, [NLL on 2015][felix-blog], a new `HashMap<K, V>` implementation, and [`--offline`] support in Cargo. Read on for a few highlights, or see the [detailed release notes][notes] for additional information.
+
+### The [`Future`] is here!
+
+[`Future`]: https://doc.rust-lang.org/std/future/trait.Future.html
+[pr-future]: https://github.com/rust-lang/rust/pull/59739
+
+In Rust 1.36.0 the long awaited [`Future`] trait has been [stabilized][pr-future]!
+
+With this stabilization, we hope to give important crates, libraries, and  the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
```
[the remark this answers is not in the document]
Centril: It isn't; we hope to release it 1.38.0 but I think we shouldn't say so in a blog post unless we are 100% sure it will happen to not raise false hopes.

posts/2019-07-04-Rust-1.36.0.md
BatmanAoD:
```suggestion
In Rust 1.36, a new Cargo flag has been stabilized: [--offline]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required crates are not available offline, and a network access would be required, Cargo will exit with an error. To prepopulate the local cache in preparation for going offline, use the [cargo fetch] command, which downloads all the required dependencies for a project.
```

posts/2019-07-04-Rust-1.36.0.md
Centril: The reason for having "Moreover, ..." here is that I find that it doesn't flow well as prose otherwise if you just start with "A..."

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### [`--offline`] support in Cargo
+
+[`--offline`]: https://doc.rust-lang.org/cargo/commands/cargo-build.html#cargo_build_manifest_options
+[`cargo fetch`]: https://doc.rust-lang.org/cargo/commands/cargo-fetch.html
+
+During most builds, Cargo doesn't interact with the network. Sometimes, however, Cargo has to. Such is the case when a dependency is added and the latest compatible version needs to be downloaded. At times, network access is not an option though, for example on an airplane or in isolated build environments.
+
+In Rust 1.36, a new Cargo flag has been stabilized: [--offline]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required crates are not available offline, and a network access would be required, Cargo will exit with an error. To prepopulate the local cache in preparation for going offline, use the [cargo fetch] command, which downloads all the required dependencies for a project.
```
Centril:
```suggestion
In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required crates are not available offline, and a network access would be required, Cargo will exit with an error. To prepopulate the local cache in preparation for going offline, use the [`cargo fetch`] command, which downloads all the required dependencies for a project.
```

posts/2019-07-04-Rust-1.36.0.md
Centril: I have no idea; r? @nrc

posts/2019-07-04-Rust-1.36.0.md
BatmanAoD: That's fair enough, but it's not consistent with lines 133 and 153. If we do have a transition word, I'd suggest "additionally", because I usually think of "moreover" as implying a closer relationship between the two sentences than what we have here.

Centril: Alright; let's go with "additionally" then. :)

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+For information on other changes to Cargo, see the [detailed release notes][relnotes-cargo].
+
+### Library changes
+
+[`dbg!`]: https://doc.rust-lang.org/std/macro.dbg.html
+
+The [`dbg!`] macro now supports multiple arguments.
+
+Moreover, a number of APIs have been made `const`:
```
Centril:
```suggestion
Additionally, a number of APIs have been made `const`:
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+With this stabilization, we hope to give important crates, libraries, and  the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
+
+### The [`alloc`][alloc-crate] crate is stable
+
+[alloc-crate]: https://doc.rust-lang.org/alloc/index.html
+
+Before 1.36.0, the standard library consisted of the crates `std`, `core`, and `proc_macro`. The `core` crate provided core functionality such as `Iterator` and `Copy` and could be used in `#![no_std]` environments since it did not impose any requirements. Meanwhile, the `std` crate provided types like `Box<T>` and OS functionality but required a global allocator and other OS capabilities in return.
+
+Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. As a result, `#![no_std]` applications defining an allocator can use libraries that depend on the `alloc` crate.  While `#![no_std]` applications still require nightly, our hope is that a `#![no_std]` compatible ecosystem of libraries can be built up meanwhile.
```
Centril:
```suggestion
Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. As a result, `#![no_std]` applications defining an allocator can use libraries that depend on the `alloc` crate.  While executable `#![no_std]` crates using `alloc` still require nightly, our hope is that a `#![no_std]` compatible ecosystem of libraries can be built up meanwhile.
```

jonas-schievink: I think we usually say "`#![no_std]` binaries" instead of "executable `#![no_std]` crates"

Centril: Updated, thanks.

posts/2019-07-04-Rust-1.36.0.md
nrc: No, Cargo would check the cached index. Nobody seems quite sure when that cached index must be updated, but that is what mostly causes network access without `--offline`. At the least, Cargo does it when it can't resolve a version.

I have a draft blog post about --offline which I'll publish at about the same time as the release (depending on how long it takes to finish the post :-) ).

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+This release brings many changes, including the stabilization of the [`Future`] trait, the [`alloc`][alloc-crate] crate, the [`MaybeUninit<T>`] type, [NLL for Rust 2015][felix-blog], a new `HashMap<K, V>` implementation, and [`--offline`] support in Cargo. Read on for a few highlights, or see the [detailed release notes][notes] for additional information.
+
+### The [`Future`] is here!
+
+[`Future`]: https://doc.rust-lang.org/std/future/trait.Future.html
+[pr-future]: https://github.com/rust-lang/rust/pull/59739
+
+In Rust 1.36.0 the long awaited [`Future`] trait has been [stabilized][pr-future]!
+
+With this stabilization, we hope to give important crates, libraries, and  the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
```
cramertj:
```suggestion
With this stabilization, we hope to give important crates, libraries, and the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
```

posts/2019-07-04-Rust-1.36.0.md
cramertj: LGTM.

posts/2019-07-04-Rust-1.36.0.md
BatmanAoD: > Nobody seems quite sure when that cached index must be updated...

🤣

posts/2019-07-04-Rust-1.36.0.md
BatmanAoD: Should both uses of the word "applications" be changed to "binaries" for consistency?

It might be better to change the ordering. Here's a draft (formatted as a quote block for readability, rather than submitted as a suggestion):

> Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in Stable Rust, and non-`#![no_std]` binaries can depend on such library crates. This should facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]`+`alloc` binaries.

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+With this stabilization, we hope to give important crates, libraries, and the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
+
+### The [`alloc`][alloc-crate] crate is stable
+
+[alloc-crate]: https://doc.rust-lang.org/alloc/index.html
+
+Before 1.36.0, the standard library consisted of the crates `std`, `core`, and `proc_macro`. The `core` crate provided core functionality such as `Iterator` and `Copy` and could be used in `#![no_std]` environments since it did not impose any requirements. Meanwhile, the `std` crate provided types like `Box<T>` and OS functionality but required a global allocator and other OS capabilities in return.
+
+Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. As a result, `#![no_std]` applications defining an allocator can use libraries that depend on the `alloc` crate.  While `#![no_std]` binaries using `alloc` still require nightly, our hope is that a `#![no_std]` compatible ecosystem of libraries can be built up meanwhile.
```
BatmanAoD:
```suggestion
Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in Stable Rust, and normal binaries (without `#![no_std]`) can depend on such library crates. This should facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```

Centril: Looks good aside from:

+ `Stable` shouldn't be capitalized.
+ Use comma instead of parenthesis.
+ "This should facilitate" -> "We hope this will facilitate" (we cannot predict the future).

Centril:
```suggestion
Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in stable Rust, and normal binaries, without `#![no_std]`, can depend on such library crates. We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```

BatmanAoD:
```suggestion
Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in stable Rust, and normal binaries, i.e. those without `#![no_std]`, can depend on such library crates. We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```

BatmanAoD: ...oh, you _just_ did this yourself. Sorry.

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+In Rust 1.36.0 the long awaited [`Future`] trait has been [stabilized][pr-future]!
+
+With this stabilization, we hope to give important crates, libraries, and the ecosystem time to prepare for `async` / `.await`, which we'll tell you more about in the future.
+
+### The [`alloc`][alloc-crate] crate is stable
+
+[alloc-crate]: https://doc.rust-lang.org/alloc/index.html
+
+Before 1.36.0, the standard library consisted of the crates `std`, `core`, and `proc_macro`. The `core` crate provided core functionality such as `Iterator` and `Copy` and could be used in `#![no_std]` environments since it did not impose any requirements. Meanwhile, the `std` crate provided types like `Box<T>` and OS functionality but required a global allocator and other OS capabilities in return.
```
SimonSapin: Instead of `Box<T>`, maybe use an example like `std::fs::File` that is not also in liballoc?

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in stable Rust, and normal binaries, without `#![no_std]`, can depend on such library crates. We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```
SimonSapin: > While `#![no_std]` binaries using `alloc` still require nightly Rust,

I think this is either wrong or misleading. As far as I know `#![no_std]` binaries in general, independently of the `alloc` crate, require Nightly on many platforms because of ``error: requires `start` lang_item``. (But for some micro-controller / embedded targets the `start` lang items is not involved, and full “binaries” can be built on stable with a custom linker script.)

Consider rewriting to something like:

> While `#![no_std]` binaries still require nightly Rust,

SimonSapin: > are now defined in the `alloc` crate.

Nit: consider “are now *available* in the `alloc` crate”. The definitions have been there for a long time, only `#[unstable]`.

Centril: Heh; this is what the blog post said at the start but I changed it after @jonas-schievink pointed out that you can use `#![no_main]` on stable and stuff. (https://github.com/jonas-schievink/rubble/blob/master/rubble-demo/src/main.rs)

Centril:
```suggestion
Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now available in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in stable Rust, and normal binaries, without `#![no_std]`, can depend on such library crates. We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```

SimonSapin: Nit: the third sentence is fairly long, has multiple commas, and makes two statements. Consider splitting it into two sentences at “and [normal binaries…]”

By the way, using a source single line for the entire paragraph makes it harder for github review comments to point to a specific part. Consider using [semantic line breaks](https://sembr.org/).

Centril: (I didn't have time to use semantic line feeds; otherwise I would have used them -- I agree this is awful... =P)

Centril:
```suggestion
Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now defined in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in stable Rust. Normal binaries, without `#![no_std]`, can depend on such library crates. We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```

posts/2019-07-04-Rust-1.36.0.md
Centril: The point here was to have an example of something that is in liballoc in opposition to OS functionality.

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### The [`alloc`][alloc-crate] crate is stable
+
+[alloc-crate]: https://doc.rust-lang.org/alloc/index.html
+
+Before 1.36.0, the standard library consisted of the crates `std`, `core`, and `proc_macro`. The `core` crate provided core functionality such as `Iterator` and `Copy` and could be used in `#![no_std]` environments since it did not impose any requirements. Meanwhile, the `std` crate provided types like `Box<T>` and OS functionality but required a global allocator and other OS capabilities in return.
+
+Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now available in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in stable Rust, and normal binaries, without `#![no_std]`, can depend on such library crates. We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```
Centril:
```suggestion
Starting with Rust 1.36.0, the parts of `std` that depend on a global allocator, e.g. `Vec<T>`, are now available in the `alloc` crate. The `std` crate then re-exports these parts. While `#![no_std]` binaries using `alloc` still require nightly Rust, `#![no_std]` library crates can use the `alloc` crate in stable Rust. Meanwhile, normal binaries, without `#![no_std]`, can depend on such library crates. We hope this will facilitate the development of a `#![no_std]` compatible ecosystem of libraries prior to stabilizing support for `#![no_std]` binaries using `alloc`.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### A new `HashMap<K, V>` implementation
+
+[`hashbrown`]: https://crates.io/crates/hashbrown
+[`HashMap<K, V>`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
+[pr-hashbrown]: https://github.com/rust-lang/rust/pull/58623
+[SwissTable]: https://abseil.io/blog/20180927-swisstables
+[pr-hashbrown-perf]: https://perf.rust-lang.org/compare.html?start=b57fe74a27590289fd657614b8ad1f3eac8a7ad2&end=abade53a649583e40ed07c26ee10652703f09b58&stat=wall-time
+
+In Rust 1.36.0, the `HashMap<K, V>` implementation has been [replaced][pr-hashbrown] with the one in the [`hashbrown`] crate which is based on the [SwissTable] design. While the interface is the same, the `HashMap<K, V>` implementation is now [faster on average][pr-hashbrown-perf] and has lower memory overhead. Note that unlike the `hashbrown` crate, the implementation in `std` continues to use the SipHash 1-3 hashing algorithm.
```
SimonSapin: Maybe “continues to use the SipHash 1-3 hashing algorithm by default”, or “continues to default to the SipHash 1-3 hashing algorithm”? `std::collections::HashMap` supports custom hashers.

Centril: Maybe ", the implementation in `std` still defaults to the SipHash 1-3 hashing algorithm." ?

posts/2019-07-04-Rust-1.36.0.md
SimonSapin: Oh, never mind. I was gonna write that it’s still not liballoc that adds a Nigthly-only requirement but that’s not true: it requires `#[alloc_error_handler]` which is not stable. (I thought it was.) https://github.com/rust-lang/rust/issues/51540

So yeah, “`#![no_std]` binaries using `alloc` still require nightly Rust” is accurate in 1.36.

posts/2019-07-04-Rust-1.36.0.md
SimonSapin: Yes, that works.

posts/2019-07-04-Rust-1.36.0.md
Centril: Alright; thanks for double checking! =)

posts/2019-07-04-Rust-1.36.0.md
Centril: Cool;
```suggestion
In Rust 1.36.0, the `HashMap<K, V>` implementation has been [replaced][pr-hashbrown] with the one in the [`hashbrown`] crate which is based on the [SwissTable] design. While the interface is the same, the `HashMap<K, V>` implementation is now [faster on average][pr-hashbrown-perf] and has lower memory overhead. Note that unlike the `hashbrown` crate, the implementation in `std` still defaults to the SipHash 1-3 hashing algorithm.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+### [`MaybeUninit<T>`] instead of [`mem::uninitialized`]
+
+[`MaybeUninit<T>`]: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html
+[`mem::uninitialized`]: https://doc.rust-lang.org/std/mem/fn.uninitialized.html
+[gankro-blog]: https://gankro.github.io/blah/initialize-me-maybe/
+[pr-60445]: https://github.com/rust-lang/rust/pull/60445
+
+In previous releases of Rust, the [`mem::uninitialized`] has allowed you bypass Rust's initialization checks by pretending that you've initialized a value at type `T` without doing anything. One of the main uses of this function has been to lazily allocate arrays.
```
RalfJung:
```suggestion
In previous releases of Rust, the [`mem::uninitialized`] function has allowed you bypass Rust's initialization checks by pretending that you've initialized a value at type `T` without doing anything. One of the main uses of this function has been to lazily allocate arrays.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+However, [`mem::uninitialized`] is an incredibly dangerous operation that essentially cannot be used correctly as the Rust compiler assumes that values are properly initialized. For example, calling `mem::uninitialized::<bool>()` causes *instantaneous __undefined behavior__* as the random bits in memory could be something other than `0` and `1`.
```
RalfJung:
```suggestion
However, [`mem::uninitialized`] is an incredibly dangerous operation that essentially cannot be used correctly as the Rust compiler assumes that values are properly initialized. For example, calling `mem::uninitialized::<bool>()` causes *instantaneous __undefined behavior__* as the random bits in memory could be something other than `0` (for `false`) and `1` (for `true`) -- the only two allowed bit patterns for `bool`.
```

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,164 @@
...
+
+### [`--offline`] support in Cargo
+
+[`--offline`]: https://doc.rust-lang.org/cargo/commands/cargo-build.html#cargo_build_manifest_options
+[`cargo fetch`]: https://doc.rust-lang.org/cargo/commands/cargo-fetch.html
+
+During most builds, Cargo doesn't interact with the network. Sometimes, however, Cargo has to. Such is the case when a dependency is added and the latest compatible version needs to be downloaded. At times, network access is not an option though, for example on an airplane or in isolated build environments.
+
+In Rust 1.36, a new Cargo flag has been stabilized: [`--offline`]. The flag alters Cargo's dependency resolution algorithm to only use locally cached dependencies. When the required crates are not available offline, and a network access would be required, Cargo will exit with an error. To prepopulate the local cache in preparation for going offline, use the [`cargo fetch`] command, which downloads all the required dependencies for a project.
+
```
nrc:
```suggestion

[nrc-blog]: https://www.ncameron.org/blog/cargo-offline/

This [blog post][nrc-blog] has a more detailed explanation.

```

posts/2019-07-04-Rust-1.36.0.md
RalfJung: Also, this perpetuates the myth that uninitialized memory contains a "random bit pattern". That's a bad mental model that keeps people from understand the true nature of `undef`. OTOH, explaining the "three-state boolean" stuff here is a bit over the top. I don't have a good idea right now for what to do instead.

Centril:
Thanks everyone!
The blog post is now ready. 🚀

posts/2019-07-04-Rust-1.36.0.md
```diff
@@ -0,0 +1,219 @@
...
+[pr-60445]: https://github.com/rust-lang/rust/pull/60445
+
+In previous releases of Rust, the [`mem::uninitialized`] function has allowed you to bypass Rust's
+initialization checks by pretending that you've initialized a value at type `T` without doing anything.
+One of the main uses of this function has been to lazily allocate arrays.
+
+However, [`mem::uninitialized`] is an incredibly dangerous operation that essentially
+cannot be used correctly as the Rust compiler assumes that values are properly initialized.
+For example, calling `mem::uninitialized::<bool>()` causes *instantaneous __undefined behavior__*
+as the random bits in memory could be something other than `0` (for `false`) and `1` (for `true`)
```
RalfJung: Maybe saying "uninitialized bits" here is a bit better, to at least not *explicitly* perpetuate the myth about uninitialized memory being "just" uninitialized bits? The linked blog post then gives further details.

Centril: Updated, please take a look.

RalfJung: Much better, thanks!

The "could" here still sounds like there is any non-determinism involved, which in fact there is not. Do you think it would be confusing to replace it by "are"?

> For example, calling `mem::uninitialized::<bool>()` causes *instantaneous __undefined behavior__* as the uninitialized bits are neither `0` (for `false`) nor `1` (for `true`)  -- the only two allowed bit patterns for `bool`.

Centril: I wonder whether we should be clearer about "abstract machine vs. low-level implementation perspective" by saying:

> For example, calling `mem::uninitialized::<bool>()` causes _instantaneous **undefined behavior**_ as, according to Rust's abstract machine, the uninitialized bits are neither `0` (for `false`) nor `1` (for `true`)  -- the only two allowed bit patterns for `bool`.

Centril: Changed to "from Rust's point of view" + "are" per discussion on Zulip. Thanks!

Centril:
And done! For real this time. =P

PR #402, opened 2019-09-06
Add support for multiple blogs

pietroalbini:
Some teams (for example the compiler team) expressed their interest in creating separate blogs on blog.rust-lang.org, to post content that might not interest the community as a whole. This PR adds support for multiple, isolated blogs while keeping the top-level one unchanged. It doesn't add any other blog yet so it's safe to merge it.

Adding a new blog is simple: a subdirectory of `posts/` needs to be created with a `blog.yml` in it (use `posts/blog.yml` as a guide), and then posts can just be written there. It's possible to nest subdirectories even if the parents aren't blogs (for example `teams/compiler` without a blog in `teams`).

There is future work that can be done to improve this:

* Create an "all" blog that contains all the posts. This will especially be useful for users with feed readers that want to read everything.
* Add a page or navigation bar to switch between blogs. At the moment there is no way to go to another blog without knowing the link. Doing this will need some design work and I'm not the best person to do that.

### DIFF
--- a/posts/blog.yml
+++ b/posts/blog.yml
@@ -1,0 +1,4 @@
+title: Rust Blog
+index-title: The Rust Programming Language Blog
+description: Empowering everyone to build reliable and efficient software.
+maintained-by: the Rust Team
--- a/src/blogs.rs
+++ b/src/blogs.rs
@@ -1,0 +1,123 @@
+use crate::posts::Post;
+use serde_derive::{Deserialize, Serialize};
+use std::error::Error;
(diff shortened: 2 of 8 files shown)


skade:
Looks good, merge conflicts need to be resolved.

pietroalbini:
Rebased on top of master.

nikomatsakis:
Shall we merge then, @skade ?

skade:
@nikomatsakis I'm fine with merging! (I just don't have merge rights)

PR #405, opened 2019-09-25
1.38 release post

Mark-Simulacrum:
cc @rust-lang/release @rust-lang/core

### DIFF
--- a/posts/2019-09-26-Rust-1.38.0.md
+++ b/posts/2019-09-26-Rust-1.38.0.md
@@ -1,0 +1,137 @@
+---
+layout: post
+title: "Announcing Rust 1.38.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.38.0. Rust is a programming language that is empowering everyone to build reliable and efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust 1.38.0 is as easy as:
+
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the appropriate page on our website.
+
+[install]: https://www.rust-lang.org/install.html
+
+## What's in 1.38.0 stable
+
+### Pipelined compilation
+
+[internals-pipelined]: https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199
+
+To compile a crate, the compiler doesn't need the dependencies to be fully built. Instead, it just needs their "metadata" (i.e. the list of types, dependencies, exports...). This metadata is produced early in the compilation process. Starting with Rust 1.38.0, Cargo will take advantage of this by automatically starting to build dependent crates as soon as metadata is ready.
+
+While the change doesn't have any effect on builds for a single crate, during testing [we got reports][internals-pipelined] of 10-20% compilation speed increases for optimized, clean builds of some crate graphs. Other ones did not improve much, and the speedup depends on the hardware running the build, so your mileage might vary. No code changes are needed to benefit from this.
+
+### Linting some incorrect uses of `mem::{uninitialized, zeroed}`
+
+As [previously announced](https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#maybeuninitt%3E-instead-of-mem::uninitialized), `std::mem::uninitialized` is essentially impossible to use safely. Instead, `MaybeUninit<T>` should be used.
+
+We have not yet deprecated `mem::uninitialized`; this will be done in a future release. Starting in 1.38.0, however, `rustc` will provide a lint for a narrow class of incorrect initializations using `mem::uninitialized` or `mem::zeroed`.
+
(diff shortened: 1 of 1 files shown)


posts/2019-09-26-Rust-1.38.0.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Linting some incorrect uses of `mem::{uninitialized, zeroed}`
+
+As [previously announced](https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#maybeuninitt%3E-instead-of-mem::uninitialized), `std::mem::uninitialized` is hard to use soundly. Instead, `MaybeUninit<T>` should be used.
+
+We have not yet deprecated `mem::uninitialized`; this will be done in a future release. Starting in 1.38.0, however, `rustc` will provide a lint for a narrow class of incorrect initializations using `mem::uninitialized` or `mem::zeroed`.
+
+It is undefined behavior for some types, such as `&T` and `Box<T>`, to ever contain an all-`0` bit pattern, because they represent pointer-like objects that cannot be `null`. It is therefore an error to use `mem::uninitialized` or `mem::zeroed` to initialize one of these types, so the new error will attempt to warn whenever one of those functions is used to initialize one of them, either directly or as a member of a larger `struct`. The check is recursive, so the following code will emit a warning:
```
jonas-schievink:
```suggestion
It is undefined behavior for some types, such as `&T` and `Box<T>`, to ever contain an all-`0` bit pattern, because they represent pointer-like objects that cannot be `null`. It is therefore an error to use `mem::uninitialized` or `mem::zeroed` to initialize one of these types, so the new lint will attempt to warn whenever one of those functions is used to initialize one of them, either directly or as a member of a larger `struct`. The check is recursive, so the following code will emit a warning:
```

posts/2019-09-26-Rust-1.38.0.md
```diff
@@ -0,0 +1,137 @@
...
+- [`Duration::div_duration_f32`] and [`Duration::div_duration_f64`]
+- [`Duration::div_f32`] and [`Duration::div_f64`]
+- [`Duration::from_secs_f32`] and [`Duration::from_secs_f64`]
+- [`Duration::mul_f32`] and [`Duration::mul_f64`]
+- Euclidean remainder and division operations -- [`div_euclid`],
+  [`rem_euclid`] -- for all integer primitives. `checked`,
+  `overflowing`, and `wrapping` versions are also available.
+
+[`<*const T>::cast`]: https://doc.rust-lang.org/std/primitive.pointer.html#method.cast
+[`<*mut T>::cast`]: https://doc.rust-lang.org/std/primitive.pointer.html#method.cast
```
jonas-schievink:
```suggestion
[`<*mut T>::cast`]: https://doc.rust-lang.org/std/primitive.pointer.html#method.cast-1
```

posts/2019-09-26-Rust-1.38.0.md
```diff
@@ -0,0 +1,137 @@
...
+
+[internals-pipelined]: https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199
+
+To compile a crate, the compiler doesn't need the dependencies to be fully built. Instead, it just needs their "metadata" (i.e. the list of types, dependencies, exports...). This metadata is produced early in the compilation process. Starting with Rust 1.38.0, Cargo will take advantage of this by automatically starting to build dependent crates as soon as metadata is ready.
+
+While the change doesn't have any effect on builds for a single crate, during testing [we got reports][internals-pipelined] of 10-20% compilation speed increases for optimized, clean builds of some crate graphs. Other ones did not improve much, and the speedup depends on the hardware running the build, so your mileage might vary. No code changes are needed to benefit from this.
+
+### Linting some incorrect uses of `mem::{uninitialized, zeroed}`
+
+As [previously announced](https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#maybeuninitt%3E-instead-of-mem::uninitialized), `std::mem::uninitialized` is hard to use soundly. Instead, `MaybeUninit<T>` should be used.
```
BatmanAoD: The 1.36 announcement says that it is "essentially impossible" to use soundly. But it's not unsound to use to initialize plain-old-data (in the C++ sense), is it? If that's correct, then "hard to use soundly" seems reasonable; if not, then maybe we should go back to using stronger language.

Centril: I'd be fine with "essentially impossible" as well -- I tweaked this because it sounded 100% certain but there are cases where it is OK... for example: `mem::uninitialized::<MaybeUninit<u8>>()`. ;)

Actually, let's change back to "essentially impossible".

Centril:
```suggestion
As [previously announced](https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#maybeuninitt%3E-instead-of-mem::uninitialized), `std::mem::uninitialized` is essentially impossible to use safely. Instead, `MaybeUninit<T>` should be used.
```

Mark-Simulacrum:
This should be ready for the release, and will be merged in ~2 hours.

PR #406, opened 2019-09-25
create the "Inside Rust" blog

nikomatsakis:
This PR creates the "Rust Team Blog", which will be visible at `blog.rust-lang.org/team`. This is a draft PR because I'd like to resolve a few questions:

- [x] prep an announcement blog post for the main blog 
- [x] Settled the name ("Inside Rust" wins!)
- [x] Modify PR to use the "Inside Rust" name and adjust other things
- [x] How should we expose this blog on the site? Right now, you'd just have to guess the URL.
    - I'm not sure this has to block us from deploying the blog now though -- there is a lot of demand to have a place 
- [x] I added an ad-hoc tag `team: core` to the first post -- I'd like each post to be tagged with a team, and ideally I'd like that to be visible in the post (and to enable filtering by team/working group). Can we at least enforce that we have a `team:` tag somewhere for now?
- [x] Any thoughts on the other "front matter" in the blog.yml file? (e.g., tagphrase)
- [x] And of course any thoughts on the text of the inaugural post itself?
- [x] A draft of blog policies are visible here https://github.com/rust-lang/wg-governance/issues/23
- [x] Remove "dual layout" system and replace with a change to blog configuration.

### DIFF
--- a/src/posts.rs
+++ b/src/posts.rs
@@ -102,7 +102,7 @@ impl Post {
 
         // Enforce extra conditions
         if manifest.requires_team && team_string.is_none() {
-            panic!("blog post at path `{}` lacks team/team_url metadata", path.display());
+            panic!("blog post at path `{}` lacks team metadata", path.display());
         }
 
         // If they supplied team, it should look like `team-text <team-url>`
--- a/src/posts.rs
+++ b/src/posts.rs
@@ -1,4 +1,6 @@
+use crate::blogs::Manifest;
 use comrak::ComrakOptions;
+use regex::Regex;
 use serde_derive::{Deserialize, Serialize};
 use std::error::Error;
 use std::path::{Path, PathBuf};
@@ -9,11 +11,14 @@ struct YamlHeader {
     author: String,
     #[serde(default)]
     release: bool,
+    team: Option<String>,
+    layout: String,
 }
 
 #[derive(Debug, Clone, Serialize)]
 pub(crate) struct Post {
     pub(crate) filename: String,
+    pub(crate) layout: String,
     pub(crate) title: String,
     pub(crate) author: String,
     pub(crate) year: String,
@@ -24,10 +29,13 @@ pub(crate) struct Post {
     pub(crate) url: String,
     pub(crate) published: String,
     pub(crate) release: bool,
+    pub(crate) has_team: bool,
+    pub(crate) team: String,
+    pub(crate) team_url: String,
 }
 
 impl Post {
-    pub(crate) fn open(path: &Path) -> Result<Self, Box<dyn Error>> {
+    pub(crate) fn open(path: &Path, manifest: &Manifest) -> Result<Self, Box<dyn Error>> {
         // yeah this might blow up, but it won't
         let filename = path.file_name().unwrap().to_str().unwrap();
 
@@ -49,6 +57,8 @@ impl Post {
             author,
             title,
             release,
+            team: team_string,
+            layout,
         } = serde_yaml::from_str(yaml)?;
         // next, the contents. we add + to get rid of the final "---\n\n"
         let options = ComrakOptions {
@@ -84,6 +94,34 @@ impl Post {
 
         let published = published.rfc3339().to_string();
 
+        // validate for now that the layout is specified as "post"
+        match &*layout {
+            "post" => (),
+            _ => panic!("blog post at path `{}` should have layout `post`", path.display()),
+        };
+
+        // Enforce extra conditions
+        if manifest.requires_team && team_string.is_none() {
+            panic!("blog post at path `{}` lacks team/team_url metadata", path.display());
+        }
+
+        // If they supplied team, it should look like `team-text <team-url>`
+        let (team, team_url) = match team_string {
+            Some(s) => {
+                lazy_static::lazy_static! {
+                    static ref R: Regex = Regex::new(r"(?P<name>[^<]*) <(?P<url>[^>]+)>").unwrap();
+                }
+                let captures = match R.captures(&s) {
+                    Some(c) => c,
+                    None => panic!("team from path `{}` should have format `$name <$url>`",
+                                   path.display()),
+                };
+                (Some(captures["name"].to_string()), Some(captures["url"].to_string()))
+            }
+
+            None => (None, None)
+        };
+
         Ok(Self {
             filename,
             title,
@@ -96,6 +134,10 @@ impl Post {
             url,
             published,
             release,
+            layout,
+            has_team: team.is_some(),
+            team: team.unwrap_or_default(),
+            team_url: team_url.unwrap_or_default(),
         })
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,10 +2,10 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.6.9"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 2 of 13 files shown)


nikomatsakis:
cc @rust-lang/core

Mark-Simulacrum:
"Rust internals" blog, perhaps? To mirror users.rust-lang.org vs internals.rust-lang.org -- though I'll admit that distinction isn't super clear either.

pietroalbini:
Maybe "Inside Rust"?

steveklabnik:
I was *very* confused when I saw the email for this appear in my inbox, given that I already think of the blog as the "rust team blog." So 👍 to a different name.

I think "development blog" would be fine. I also like the "internals blog" idea; I agree the users/internals split isn't always super obvious, but I think it's probably fine.

nikomatsakis:
Another thought I had was the "Rust Teams blog" -- using the plural there kind of makes it sound mildly 
different to me. I guess that "Rust Team" and "Rust Project" feel more analogous and that Rust Teams is kind of more clearly "Blog for the Rust teams". Still, it's not a big change.

Anyway, my main concern with the name internals is that it sounds like the blog is about the internals of Rust, which, I guess it is, but only from a certain POV. (Feels a lot like "dev blog" to me.) 

Another thought: give it a more creative name, like "Inside the Rust project"? "Rust insider's blog" (that sounds exclusive). 

Final thought: Just use the main blog for everything? :) But I still think it's useful to have a separate channel, it'll encourage us to post a lot more stuff on here (e.g., it would feel weird for me to post minutes from compiler team meetings on the main blog, right? But I intend to start posting those here.)

nikomatsakis:
Updated the main issue with the names proposed thus far. I'd definitely like to get this setup because I have an announcement I want to post. =) I guess most any name will work out ok.

wesleywiser:
When I first saw mentions of the team blog, I assumed each team was getting their own blog. IE, `blog.rlo/compiler` for the compiler team, `blog.rlo/lang` for the lang team, `blog.rlo/community` for the community team, etc. Perhaps that idea is worth exploring more? 

I think there's some interesting advantages to that idea:

- Resolves the naming issue, obviously.
- Allows each blog to be more targeted for its audience. For example, a dedicated compiler team blog could be used to advertise large scale changes to rustc that impact contributors but are of little interest to most others. 
  - Periodically, the compiler team also has issues it wants to broadcast more widely and ask for help from the community with. A dedicated blog would be a good place to mention those.
- Implicitly has the "ad hoc team tag" you mention by virtue of each team having their own blog.
- Can be advertised from the Governance pages for each team.

nikomatsakis:
@wesleywiser Yeah, I definitely considered that, but it seems to be rather a lot of blogs, and I worry that they smaller teams or working groups won't get as much "circulation" this way. I think in my *ideal* world you'd be able to pick-and-choose the teams you want to "follow". 

Still, it *would* resolve the naming issue, it's true. =) 

>Periodically, the compiler team also has issues it wants to broadcast more widely and ask for help from the community with. A dedicated blog would be a good place to mention those.

I absolutely expect to post those on this blog, along with meeting notes and the like, and posts from smaller groups like `self-profile`. If people think that would be inappropriate for a single blog, then I agree we should shift.

nikomatsakis:
(The other thing is that I think a lot of posts will be kind of "cross-team" -- e.g., are async-await updates lang, or compiler?)

nikomatsakis:
OK! I propose people give their opinion on one of two possibilities:

- name the blog "Inside Rust" (add 🎉 emoji)
- create per-team / per-working-group blogs (add 🚀 emoji)

I don't know which one I prefer. Some thoughts [here](https://github.com/rust-lang/blog.rust-lang.org/pull/406#issuecomment-535531509) and [here](https://github.com/rust-lang/blog.rust-lang.org/pull/406#issuecomment-535547139). Feel free to leave your own.

steveklabnik:
An additional possible per-blog option is to create tags, and have them be on the same blog, but you could filter by tag.

nikomatsakis:
@steveklabnik yep, that's what I was hoping we would do.

Also worth pointing out: this is not an irreversible decision. We can always break out sub-blogs later, or add special urls that filter by tag.

nikomatsakis:
OK, it seems like the "poll" has spoken:

"Inside Rust" it is! I think the plan is that we'll enforce the `team:` tags (informally, perhaps, to start) and at some point enable filtering in the UI (and perhaps RSS feeds or whatever as well (do people still use RSS?)). I'll take a look later to see how hard it would be to reject posts without the `team:` tags, can't be *that* hard, right?

I think the only other major question that we probably want to settle before deploying is:

>  How should we expose this blog on the site? Right now, you'd just have to guess the URL. 

Of course we don't *have*  to link the site from the main page, but it seems like it'd be nice to. @skade @Manishearth, any thoughts on this? Should I just take a stab to see if I can figure out how to add it to that "main list" at the top of the site of pages? (i.e., next to "Blog") It feels like a major change to the site UI that might get kind of "busy".

skade:
We should link it from the page somewhere.

nikomatsakis:
Hmm, we no longer have a "contribute page", right?

nikomatsakis:
But maybe we can link to it from the Governance and Community pages, or something?

nikomatsakis:
OK, I went a little crazy, and upgraded the code a bit to respect the `team` and `team_url` metadata, and incorporated that into the layout:

image

This is about the limit of my web-design skills.

I decided against making `team` an enum because I didn't want to add PRs to the blog each time some new team wants to publish or whatever. On the other hand, maybe just saying "core" is a bit terse. I don't know, seems ok for now, we can improve.

nikomatsakis:
So, the only question is how to incorporate this into the main site -- and @skade I realize now that this is not something we can do in *this* PR anyway.

The URL would be `blog.rust-lang.org/inside-rust`, I believe.

nikomatsakis:
or maybe @steveklabnik =)

nikomatsakis:
Idea: I will look into creating a link to the Inside Rust blog from the main blog header or something also, since it seems likely people would try to navigate that way. That plus a link on governance page seems good.

nikomatsakis:
@pietroalbini 

> That should allow us to keep only one layout for blog posts.

I started with that but rejected it because it didn't let us check for sure that the blog had a team/team-url supplied. That said, maybe I could post that in the blog configuration instead.

pietroalbini:
Yeah, a `required_team: true` in the blog config could work.

nikomatsakis:
@pietroalbini ok I changed the setup as requested. There is still no link to the inside-rust blog from anywhere else, not sure what to do about that.

nikomatsakis:
how it looks now

image

nikomatsakis:
I added some code to list other blogs in the "posts index" page. It...doesn't look great yet. But maybe there's something there.

image

nikomatsakis:
Probably it would look better in the upper-left corner :)

src/posts.rs
```diff
@@ -84,6 +94,34 @@ impl Post {
...
 
+        // validate for now that the layout is specified as "post"
+        match &*layout {
+            "post" => (),
+            _ => panic!("blog post at path `{}` should have layout `post`", path.display()),
+        };
+
+        // Enforce extra conditions
+        if manifest.requires_team && team_string.is_none() {
+            panic!("blog post at path `{}` lacks team/team_url metadata", path.display());
```
pietroalbini: Nit:

```suggestion
            panic!("blog post at path `{}` lacks team metadata", path.display());
```

nikomatsakis:
I think I'll revert the "see also" work I did, as I think it's pretty horrible looking.

nikomatsakis:
OK, I decided to take one more stab at the idea. I realized that the current "Posts" text was pretty minimal anyway and added the ability to customize the html that appears in the index. Now you have something like this:

image

and on the Inside Rust blog:

image

I've pushed those commits. Obviously would love feedback on the text itself!

nrc:
In the Rust blog, maybe s/we/the core team to be consistent with the other blurb, otherwise lgtm

nikomatsakis:
OK -- I've pushed a few minor commits including an "Announcing the Inside Rust blog!" post for the main blog. I've dated that post for **tomorrow, October 3rd**, since I figured I'd rather announce this thing tomorrow vs tonight.

Therefore, if you've any last feedback, please make it known before tomorrow!

cc @rust-lang/core

nikomatsakis:
OK, I'm going to pull the trigger here!

PR #422, opened 2019-10-13
fix typo in blog post about async-await diagnostics

qutesy:
If I'm reading the code correctly, and if I understand the Sync requirement correctly, the problem here is that bar() _locks_ the mutex, not that bar() unlocks the mutex.


alexcrichton:
cc @davidtwco

PR #425, opened 2019-10-14
Add post about rustup 1.20.0

pietroalbini:
This post announces the release of rustup 1.20.0, scheduled to be released tomorrow. I wrote the post with help from @kinnison and a proof-read by @Centril.

Mark and Steve on Discord agreed about making a post on the main blog.

cc @rust-lang/core

### DIFF
--- a/posts/2019-10-15-Rustup-1.20.0.md
+++ b/posts/2019-10-15-Rustup-1.20.0.md
@@ -29,7 +29,7 @@ Previous versions of rustup installed a few components by default along with eac
 
 To address this problem, rustup 1.20.0 introduces the concept of "profiles". They are groups of components you can choose to download while installing a Rust toolchain. The profiles available at this time are `minimal`, `default`, and `complete`:
 
-* The **minimal** profile includes as few components as possible to get a working compiler (`rustc`, `rust-std`, and `cargo`). It's recommended to use this component on CIs and Windows systems.
+* The **minimal** profile includes as few components as possible to get a working compiler (`rustc`, `rust-std`, and `cargo`). It's recommended to use this component on Windows systems if you don't use local documentation, and in CI.
 * The **default** profile includes all the components previously installed by default (`rustc`, `rust-std`, `cargo`, and `rust-docs`) plus `rustfmt` and `clippy`. This profile will be used by rustup by default, and it's the one recommended for general use.
 * The **complete** profile includes all the components available through rustup, including `miri` and IDE integration tools (`rls` and `rust-analysis`).
 
--- a/posts/2019-10-15-Rustup-1.20.0.md
+++ b/posts/2019-10-15-Rustup-1.20.0.md
@@ -1,0 +1,80 @@
+---
+layout: post
+title: "Announcing Rustup 1.20.0"
+author: The Rustup Working Group
+---
+
+The rustup working group is happy to announce the release of rustup version 1.20.0. [Rustup][install] is the recommended tool to install [Rust][rust], a programming language that is empowering everyone to build reliable and efficient software.
+
+If you have a previous version of rustup installed, getting rustup 1.20.0 is as easy as:
+
+```
+rustup self update
+```
+
+If you don't have it already, you can [get rustup][install] from the appropriate page on our website.
+
+[rust]: https://www.rust-lang.org
+[install]: https://rustup.rs
+
+## What's new in rustup 1.20.0
+
+The highlights of this release are profiles support, the ability to get the latest available nightly with all the components you need, and improvements to the `rustup doc` command. You can also check out [the changelog][changelog] for a list of all the changes included in this release.
+
+[changelog]: https://github.com/rust-lang/rustup.rs/blob/master/CHANGELOG.md
+
+### Profiles
+
+Previous versions of rustup installed a few components by default along with each toolchain: the compiler (`rustc`), the package manager (`cargo`), the standard library (`rust-std`), and offline documentation (`rust-docs`). While this approach is fine while developing software locally, some of the components (like `rust-docs`) slowed down the installation due to the large amount of installed files, especially on CIs and on Windows.
+
+To address this problem, rustup 1.20.0 introduces the concept of "profiles". They are groups of components you can choose to download while installing a Rust toolchain. The profiles available at this time are `minimal`, `default`, and `complete`:
+
+* The **minimal** profile includes as few components as possible to get a working compiler (`rustc`, `rust-std`, and `cargo`). It's recommended to use this component on CIs and Windows systems.
+* The **default** profile includes all the components previously installed by default (`rustc`, `rust-std`, `cargo`, and `rust-docs`) plus `rustfmt` and `clippy`. This profile will be used by rustup by default, and it's the one recommended for general use.
+* The **complete** profile includes all the components available through rustup, including `miri` and IDE integration tools (`rls` and `rust-analysis`).
+
+To change the rustup profile you can use the `rustup set profile` command. For example, to select the minimal profile you can use:
+
+```
+rustup set profile minimal
+```
+
+It's also possible to choose the profile at installation time, either interactively by choosing the "Customize installation" option or programmaticaly by passing the `--profile=<name>` flag. As usual it will be possible to install individual components not included in your profile with: `rustup component add`.
+
+### Installing the latest compatible nightly
+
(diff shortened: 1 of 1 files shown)


posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+rustup set profile minimal
+```
+
+It's also possible to choose the profile at installation time, either interactively by choosing the "Customize installation" option or programmaticaly by passing the `--profile=<name>` flag. As usual it will be possible to install individual components not included in your profile with: `rustup component add`.
+
+### Installing the latest compatible nightly
+
+While all components are guaranteed to be present on stable releases of [tier 1 platforms][tiers], the same guarantee doesn't apply to nightly builds. Frequently, tools such as `rustfmt`, `clippy`, or `rls` are missing in the latest nightly. If you depend on these tools, that makes updating nighties hard, as rustup will prevent the upgrade if a component you previously installed is missing.
```
Mark-Simulacrum: I think "all components" here is too strong. For example, we don't guarantee that we ship a working miri on stable/beta I believe. Maybe we can say "the default profile" or so?

posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+
+Starting from rustup 1.20.0, if a component you previously installed is missing in the latest nightly, `rustup update` will walk backwards in time to find the most recent release with all the components you need. If there are no new nightlies with all the components you need you'll either need to wait another day or remove some of them.
```
Mark-Simulacrum:
```suggestion
Starting from rustup 1.20.0, if a component you previously installed is missing in the latest nightly, `rustup update` will walk backwards in time to find the most recent release with all the components you need. If there are no new nightlies with all the components you need you'll either need to wait or remove some of them.
```

Let's not be too optimistic :)

posts/2019-10-15-Rustup-1.20.0.md
pietroalbini: Well, we also guarantee rls and related components on stable. What about this?

```suggestion
While most components are guaranteed to be present on stable releases of [tier 1 platforms][tiers], the same guarantee doesn't apply to nightly builds. Frequently, tools such as `rustfmt`, `clippy`, or `rls` are missing in the latest nightly. If you depend on these tools, that makes updating nighties hard, as rustup will prevent the upgrade if a component you previously installed is missing.
```

posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+title: "Announcing Rustup 1.20.0"
+author: The Rustup Working Group
+---
+
+The rustup working group is happy to announce the release of rustup version 1.20.0. [Rustup][install] is the recommended tool to install [Rust][rust], a programming language that is empowering everyone to build reliable and efficient software.
+
+If you have a previous version of rustup installed, getting rustup 1.20.0 is as easy as:
+
+```
+rustup self update
```
nrc: Perhaps worth menitioning that for most users, you'll get the update automatically the next time you run `rustup update`.

posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+
+## What's new in rustup 1.20.0
+
+The highlights of this release are profiles support, the ability to get the latest available nightly with all the components you need, and improvements to the `rustup doc` command. You can also check out [the changelog][changelog] for a list of all the changes included in this release.
+
+[changelog]: https://github.com/rust-lang/rustup.rs/blob/master/CHANGELOG.md
+
+### Profiles
+
+Previous versions of rustup installed a few components by default along with each toolchain: the compiler (`rustc`), the package manager (`cargo`), the standard library (`rust-std`), and offline documentation (`rust-docs`). While this approach is fine while developing software locally, some of the components (like `rust-docs`) slowed down the installation due to the large amount of installed files, especially on CIs and on Windows.
```
nrc: "installation due to the large amount of installed files" I think this is only true on Windows, elsewhere it is just the size which makes things slow. It's unnecessary on CI, but not any slower than it would be locally.

posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+
+To address this problem, rustup 1.20.0 introduces the concept of "profiles". They are groups of components you can choose to download while installing a Rust toolchain. The profiles available at this time are `minimal`, `default`, and `complete`:
```
nrc: "installing a Rust toolchain" - maybe "a new Rust toolchain" to emphasise that profiles only have an effect on new installs.

posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+
+* The **minimal** profile includes as few components as possible to get a working compiler (`rustc`, `rust-std`, and `cargo`). It's recommended to use this component on CIs and Windows systems.
```
nrc:
```suggestion
* The **minimal** profile includes as few components as possible to get a working compiler (`rustc`, `rust-std`, and `cargo`). It's recommended to use this component on Windows systems if you don't use local documentation, and in CI.
```

posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+* The **default** profile includes all the components previously installed by default (`rustc`, `rust-std`, `cargo`, and `rust-docs`) plus `rustfmt` and `clippy`. This profile will be used by rustup by default, and it's the one recommended for general use.
+* The **complete** profile includes all the components available through rustup, including `miri` and IDE integration tools (`rls` and `rust-analysis`).
+
+To change the rustup profile you can use the `rustup set profile` command. For example, to select the minimal profile you can use:
+
+```
+rustup set profile minimal
+```
+
+It's also possible to choose the profile at installation time, either interactively by choosing the "Customize installation" option or programmaticaly by passing the `--profile=<name>` flag. As usual it will be possible to install individual components not included in your profile with: `rustup component add`.
```
nrc: Note that this won't actually change the components currently installed or installed at the next update, it will only affect new toolchains.

posts/2019-10-15-Rustup-1.20.0.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Improvements to `rustup doc`
+
+The `rustup doc` command opens the locally installed documentation on your browser, without any Internet connection required. rustup 1.20.0 enhances the command allowing you to open directly the API documentation of a specific item. For example to look at the documentation of [`Iterator`] you can use:
+
+```
+rustup doc std::iter::Iterator
+```
+
+This works for traits, structs/enums, macros, and modules, and can take you to the `std`, `alloc`, and `core` crates. We will be improving the command's UX over time, so if you have ideas, please do let us know!
```
nrc: Now that docs component is optional, maybe mention here that `rustup doc` will only work if docs component is installed.

posts/2019-10-15-Rustup-1.20.0.md
kinnison: :+1:

posts/2019-10-15-Rustup-1.20.0.md
kinnison: Yeah we probably should, are you okay to sort a bit of wording @pietroalbini ?  Something like:

Also, rustup will automatically fetch updates for itself at the end of a normal toolchain update:

```
rustup update
```

?

posts/2019-10-15-Rustup-1.20.0.md
kinnison: It's true on *NIX platforms with latent storage (e.g. HPC clusters with NFS homedirs - and yes, we had someone with that on the issue when we were discussing mitigation strategies)

posts/2019-10-15-Rustup-1.20.0.md
kinnison: I think "new" might still be confusing since I could see people thinking that upgrading from 1.XX to 1.XX+1 as "installing a new toolchain".  Perhaps "installing a fresh Rust toolchain" ?

kinnison: I'm also OK with it as it is.

posts/2019-10-15-Rustup-1.20.0.md
kinnison: Good point, I had that in the original wording but it got lost when I was rewriting it.

@pietroalbini something like:

Note, however, that this functionality will only work if you have the `rust-docs` component installed in your toolchain.

PR #439, opened 2019-10-28
add in 2020 call for blogs

steveklabnik:
It's that time of the year again!

cc @rust-lang/core


steveklabnik:
> Looks good to me! When do we want to post?

We had previously suggested today. I'm fine whenever.

PR #463, opened 2019-11-22
[Inside Rust] Add const prop on by default blog post

wesleywiser:
r? @nikomatsakis 

cc @oli-obk who already provided a review on a draft of this post


posts/inside-rust/2019-11-22-const-prop-on-by-default.md
```diff
@@ -0,0 +1,117 @@
...
+which is further reduced to just:
+
+```rust
+let x = 26;
+```
+
+## Compiler performance
+
+As you might have guessed, reducing the amount of control flow processed by the Rust compiler has a positive effect on compile times.
+We're seeing [2-10% improvement][perf_times] on a variety of test cases in both debug and release mode.
```
Mark-Simulacrum: note: this link will go dead quite soon; we're jettisoning old data sometime this weekend most likely.

It might re-appear later once we recollect but that's unlikely to happen quickly.

wesleywiser: Oh, good to know thanks! I guess I'll just remove the link?

nikomatsakis:
Here is a question -- should this maybe go on the main blog? Curious to hear what @rust-lang/core things

wesleywiser:
IMO Inside Rust is a good place for this post but we can certainly link to it from the 1.41 release announcement when that time comes.

nikomatsakis:
Seems fine. I don't have a strong opinion. Let's just post it on Inside Rust.

PR #466, opened 2019-11-25
Add syntax highlighting for blog posts

### LINKED ISSUE #342: Add code highlighting to code in blog posts
Just a suggestion. I think the old blog had that?

ecstatic-morse:
Resolves #342. 

Unlike the [year-old WIP PR](https://github.com/rust-lang/blog.rust-lang.org/pull/338), this uses a vendored version of the `highlightjs` library with the exact same theme as [the rust book](https://github.com/rust-lang/book). It may be preferable to use a cloudflare hosted version instead for better caching as well as transparency (I could have slipped bad stuff into the minified JS). I'm not sure what the policy is around external dependencies.

Screenshot (compare to the [original](https://blog.rust-lang.org/2019/11/07/Rust-1.39.0.html#attributes-on-function-parameters)):

image

cc @steveklabnik


alexcrichton:
r? @nikomatsakis 

or can you delegate to someone else who knows more about our blog setup?

steveklabnik:
We wanted to wait until we could make a decision that would keep consistency across the website and blog. That said, maybe it's better to move forward where we can; what does @rust-lang/core think?

Manishearth:
I think we should land this

ecstatic-morse:
cc @pietroalbini because they reviewed rust-lang/www.rust-lang.org#764.

pietroalbini:
Yeah, please avoid inline script. Also, the js can be moved at the end of the document, as we don't need to block on highlight to render the post.

static/styles/highlight.css
```diff
@@ -0,0 +1,79 @@
+/* Base16 Atelier Dune Light - Theme */
+/* by Bram de Haan (http://atelierbram.github.io/syntax-highlighting/atelier-schemes/dune) */
```
steveklabnik: https://github.com/atelierbram/syntax-highlighting/blob/master/LICENSE

static/styles/highlight.css
```diff
@@ -0,0 +1,79 @@
...
+/* Original Base16 color scheme by Chris Kempson (https://github.com/chriskempson/base16) */
```
steveklabnik: https://github.com/chriskempson/base16/blob/master/LICENSE.md

PR #468, opened 2019-11-26
[Inside Rust] adds in blog post summarizing the 11/19/2019 meeting of the Governanc…

nellshamrell:
…e WG

Signed-off-by: Nell Shamrell <nellshamrell@gmail.com>


XAMPPRocky:
Also cc @rust-lang/wg-governance

BatmanAoD:
I'm not sure the organization by person-who-brought-up-topics is helpful to anyone who wasn't actually at the meeting. Maybe just combining those sections into one giant hierarchical list would be preferable.

(If we do keep the names, then we should probably clarify in "attendees" that I'm Kyle, since otherwise that's a bit ambiguous.)

XAMPPRocky:
Yeah I think we could replace full meeting notes with a link to them in the repo. https://github.com/rust-lang/wg-governance/blob/master/minutes/2019.11.19.md

nellshamrell:
Can do!

nellshamrell:
Changes have been made and pull request has been updated!

XAMPPRocky:
👍

PR #469, opened 2019-11-26
Future of rust ides

matklad:
cc @Xankewok

r? @nikomatsakis

### DIFF
--- a/posts/inside-rust/2019-11-19-ide-future.md
+++ b/posts/inside-rust/2019-11-19-ide-future.md
@@ -24,11 +24,13 @@ The main benefits of RLS is precision (it uses `rustc` under the hood).
 Additionally, RLS is the main consumer of save-analysis infrastructure, which is a good fit for tools which need a static view of the codebase, such as [cargo-src](https://github.com/rust-dev-tools/cargo-src) or [lsif](https://code.visualstudio.com/blogs/2019/02/19/lsif).
 
 # Save-analysis
-TODO(xanewok): Intro a bit what is it and how it allows us to be 'precise'
 
 What is "save-analysis"?
-It is a currently unstable format which rustc uses to record information about the compiled code.
-`env RUSTFLAGS="-Zunstable-options -Zsave-analysis" cargo check`
+It is an unstable format which rustc uses to record information about the compiled code.
+It contains a pretty high-level information.
+For example, for each identifier in the source-crate, save-analyzer will map this identifier to a definition and list of usages.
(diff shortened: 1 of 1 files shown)


matklad:
Hm, I agree that this is a risk. I intentionally tried to keep this a short report, rather than a long post, because the stuff discussed is very much up in the air. I feel that the better place for a thorough survey of the landscape would be an RFC.

I suggest doing either of these two things:

* *expanding* the post with covering more of the history, background, etc. Basically, writing an RFC, but not calling it an RFC. Maybe it's even better to just write an RFC? Like, RFC doesn't need to be fully "complete" from the start, we can put save-analysis experiments to the unresolved questions section and shut down the RFC, should the experiments be unsuccessful. 

* tuning down the conclusions section, to make it less "sensational":

># Way forward
>Our current hypothesis is that it is possible to integrate both approaches without doubling the engineering effort. We will test this by adding an option to rust-analyzer to use save-analysis for find-usages and rename functionality. That way, we'll get precise results for the most important queries, without slowing down completion. If this works out, it will unlock new ways to merge rust-analyzer and RLS.

nikomatsakis:
@matklad yes, I approve of keeping it short. Let me see if I can play around a bit

nikomatsakis:
@matklad take a look at that last commit -- feel free to revert -- maybe just adding a bit of context helps?

matklad:
@nikomatsakis oh, yes, your edits totally make sense. It even didn't occur to me that "library-ification" is something that needs explaining, but of course it does!

nikomatsakis:
OK. I'm good to post it then, if you are.

matklad:
I am good!

nikomatsakis:
Hmm @matklad I just merged but now I am wondering -- maybe we should tweak the title?

nikomatsakis:
Oh, never mind, I was confusing the PR title for blog post title. I did go ahead and update the date.

kud1ing:
Unsolicited feedback regarding the usage of the term "IDE" as in:

> At the moment, the two IDEs developed as part of the Rust project are Rust Language Server (RLS) and rust-analyzer.

My understanding of the meaing of "IDE" matches what is described in Wikipedia: "An IDE normally consists of at least a source code editor, build automation tools and a debugger.".
Correct me if i am wrong, this is not what my understanding of what `rls`, `rust-analyzer` provide.

In order to prevent confusion i suggest:
* either explain that "IDE" in the Rust context means something different
* do not label thos tools als "IDE"s

matklad:
Hm, yeah, I think, in this context, it's better to use "language server" instead.

Issue #355, opened 2018-12-21
All posts are showing as being posted at 0:00 UTC


steveklabnik:
Given that we don't know exactly *when* down to the minute a post is published, I'm not sure how to do this. https://github.com/rust-lang/blog.rust-lang.org/blob/bdd0a9587f6b1aec4140df23345f472a570ed70f/src/main.rs#L132-L145 should be updated to fix it somehow, though.

steveklabnik:
This also leads to a validation warning, since two posts on the same day show the same time

![image](https://user-images.githubusercontent.com/27786/50360693-db74a500-052e-11e9-9747-fafd79fcdd7e.png)

Zexbe:
The problem could be solved by adding an optional hour at the end of the current format. 
2020-01-24-Blog-Post.md would be UTC 0 for compatibility with existing blogs, and convenience.
2020-01-24-05-Blog-Post.md would be at UTC 5

It could also double as a way to control the order of blog posts on the same day
2020-01-24-00-Blog-Post.md
2020-01-24-01-Blog-Post.md
2020-01-24-02-Blog-Post.md

Mark-Simulacrum:
I would prefer to avoid adding times to the posts, that just adds to the (existing) pain point of the dates. Maybe we can pull the timestamp information from git?

Zexbe:
@Mark-Simulacrum 
Possible ways to get it from get:
Using get log: downsides
1. The current dates are in the url, the git log dates don't match up with the current files, so it would break urls.
2. If two files are committed at the same time, it would probably get the same time.
3. If you wanted to show on a specific holiday, or something. you might want to commit ahead of time, but have the date in the future.

Mark-Simulacrum:
Hm, it is a good point that the two may not be in sync. Can we just generate semi-random times at buildtime? e.g. every post gets released $i minutes into the day where i increases by one for the posts assigned to that day?

Zexbe:
If you have a Part 1, and Part 2, controlling the order might be important

Mark-Simulacrum:
We almost never have part 1 and part 2, and the order really only applies to the listing, which I personally don't care about (i.e., if things are in the wrong order there, whatever). Most people access blog posts via direct links.

Zexbe:
It could support these two formats  YYYY-MM-DD and YYYY-MM-DD-HH, but fall back to using the git time if it's not in that format. So new posts don't have to specify, but the old links stay.

[...]

PR #498, opened 2020-01-25
Make it so updated time is always unique in feed.xml

Zexbe:
Resolves issue: https://github.com/rust-lang/blog.rust-lang.org/issues/355
This change separates published time, and updated time, and ensures that updated time is unique for blog posts on the same day.

### DIFF
--- a/src/posts.rs
+++ b/src/posts.rs
@@ -4,6 +4,7 @@ use regex::Regex;
 use serde_derive::{Deserialize, Serialize};
 use std::error::Error;
 use std::path::{Path, PathBuf};
+use time::{Duration, Tm};
 
 #[derive(Debug, PartialEq, Deserialize)]
 struct YamlHeader {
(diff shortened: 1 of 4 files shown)


src/posts.rs
```diff
@@ -134,11 +129,36 @@ impl Post {
...
+        tm_mon: (month as i32) - 1,    // 0-11 not 1-12
+        tm_year: (year as i32) - 1900, // from the year 1900, not the actual year
+        // these next two fields are wrong but we never use them to generate our times
+        tm_wday: 1,
+        tm_yday: 1,
+        tm_isdst: 0,
+        tm_utcoff: 0,
+        tm_nsec: 0,
+    };
+    time = time + Duration::seconds(seconds as i64);
```
Mark-Simulacrum: hm, can we assert that these seconds aren't putting us more than a day off?

Zexbe: I added a check

Mark-Simulacrum:
Thanks!

PR #506, opened 2020-01-29
Fix h3 on blog posts

### LINKED ISSUE #453: Are section headers supposed to be `inline-block`?
Most other blog templates I've used don't have this style so I was a bit confused to see that

```markdown
## Working group sync 

### [wg-polonius](link)
```

renders on one line

image

If this is intended behavior, feel free to close this issue. I tried looking at other published blog posts and couldn't find any that triggered this (or were relying on the `inline-block` behavior).

Zexbe:
The CSS was setting h3 to inline-block, this caused it to sometimes combine headers in blog posts unintentionally.

Fixes #453


alexcrichton:
Thanks!

PR #507, opened 2020-01-30
Rust 1.41.0 announcement

pietroalbini:
Mainly by me and @Centril.

cc @rust-lang/core @rust-lang/release

### DIFF
--- a/posts/2020-01-30-Rust-1.41.0.md
+++ b/posts/2020-01-30-Rust-1.41.0.md
@@ -173,7 +173,7 @@ In Rust 1.41.0, we've made the following additions to the standard library:
 - [`NonZero*` numerics now implement `From<NonZero*>` if it's a smaller integer
   width.][pr_66277] For example, `NonZeroU16` now implements `From<NonZeroU8>`.
 
-- The `weak_count` and `strong_count` methods were stabilized.
+- The `weak_count` and `strong_count` methods on `Weak` pointers were stabilized.
 
     - [`std::rc::Weak::weak_count`]
     - [`std::rc::Weak::strong_count`]
--- a/posts/2020-01-30-Rust-1.41.0.md
+++ b/posts/2020-01-30-Rust-1.41.0.md
@@ -1,0 +1,212 @@
+---
+layout: post
+title: "Announcing Rust 1.41.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.41.0. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.41.0 is as easy as:
+
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.41.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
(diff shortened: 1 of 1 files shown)


posts/2020-01-30-Rust-1.41.0.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Other changes
+
+[relnotes-cargo]: https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-141-2020-01-30
+[relnotes-clippy]: https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-141
+
+There are other changes in the Rust 1.41.0 release: check out what changed in
+[Rust][notes], [Cargo][relnotes-cargo], and [Clippy][relnotes-clippy]. We also
+have started landing MIR optimizations, which should improve compile times. For
+more, read [mir-opt].
```
Mark-Simulacrum: I don't think we ever put this link in. Should be https://blog.rust-lang.org/inside-rust/2019/12/02/const-prop-on-by-default.html - on a phone so can't push right now.

pietroalbini: Fixed.

posts/2020-01-30-Rust-1.41.0.md
```diff
@@ -0,0 +1,212 @@
...
+
+- The [`Result::map_or`] and [`Result::map_or_else`] methods were stabilized.
+
+  Similar to [`Option::map_or`] and [`Option::map_or_else`], these methods are
+  shorthands for the `.map(|val| process(val)).unwrap_or(default)` pattern.
+
+- [`NonZero*` numerics now implement `From<NonZero*>` if it's a smaller integer
+  width.][pr_66277] For example, `NonZeroU16` now implements `From<NonZeroU8>`.
+
+- The `weak_count` and `strong_count` methods were stabilized.
```
jonas-schievink: These were already stable on `Arc`/`Rc`, only the ones on `Weak` are new, so maybe:

```suggestion
- The `weak_count` and `strong_count` methods on `Weak` pointers were stabilized.
```

CryZe:
What about overriding the build settings of dependencies? That's like the biggest feature of 1.41 for me and it seems absent here.

pietroalbini:
@CryZe announcement blog posts never contain the full set of changes in a release, but a small amount of highlights the release team thinks are interesting to a large amount of users.

CryZe:
Yeah I agree, but not pointing that out and instead pointing out minor things like Box being FFI compatible (which is funny, cause I was the one pushing for Box being FFI compatible) is kind of weird. The cargo changes (also the docs documenting private items now) are much more useful to "the common developer" than obscure FFI things that only a minority of people will ever use. Being able to boost the performance of dependencies is huge for pretty much everyone working on games or other performance critical code that is way too slow to be executed in the debug profile at the moment. So I very much disagree with not mentioning that at all.

Maybe even just a small section that just talks about cargo having gained support for documenting private items and changing the build settings of dependencies without going too deep into it would be fine.

steveklabnik:
Release blog posts used to have a whole Cargo section, which is now apparently relegated to a link to its changelog.

Mark-Simulacrum:
To be clear, this was not (at least on my part) an intentional decision; I was simply not aware that this was shipping in this release. I may try to fit it in if I have time between now and the release, but I don't think that's likely to happen.

We sometimes have a Cargo section, and sometimes don't -- it varies based on what is deemed important. I think this cycle we were all pretty strapped on time and less effort/review than usual went into the blog post, too.

PR #516, opened 2020-02-14
[Inside Rust] Intro to rustc's self-profile feature

wesleywiser:
r? @nikomatsakis 

cc @michaelwoerister who gave feedback on an earlier draft of this post


nikomatsakis:
This post is great!

nikomatsakis:
I'm wondering if this ought to be promoted to the main Rust blog. It feels like the audience is more "people *using* Rust" than "people *developing* Rust". I asked this [over on Discord](https://discordapp.com/channels/442252698964721669/579961487578234880/677922092271796266), too.

nikomatsakis:
We had some brief conversation in Discord on this topic. I think the conclusion was that this could indeed be a nice fit for the main blog, but if so, it could use a bit more of an intro. @wesleywiser how would you feel about adding some sort of intro to help explain to people what this feature is all about?

I was imagining something like this:

> We often hear requests to know where compilation time is going. This is useful to use when optimizing the compiler, but also useful to users who want to refactor their crate so that it will compile faster. We've been working on an experimental feature that will help with us. This blog post gives a preview of how it works. Be warned, though: this is still experimental, and we expect the interface to change over time. We'll be adding these examples to the [Rust unstable book](https://doc.rust-lang.org/unstable-book/), as well, and we plan to keep that up to date, so you can always find the latest instructions here.

wesleywiser:
@nikomatsakis and I chatted a bit about this and we decided not to move it to the main blog at this time. While we both agree that the main blog is a perhaps underutilized, I think that this post is a bit too in-the-weeds for the main blog. 

Instead, we'd like to run a short series of posts on the main blog advertising the improvements that have been made to Rust compile times over the years as well as the contributors driving those improvements, the current state of things including profiling tools like these and anticipated in-progress or future improvements. 

As for this post, I'm going to 

- [ ] adapt the into to include a modified version of what @nikomatsakis suggested:

> Over the last year, the Self-Profile Working Group has been building tools to profile `rustc` because we often hear requests to know where compilation time is being spent.
This is useful when optimizing the compiler, but it's also useful to users who want to refactor their crate so that it will compile faster.
We've been working on an experimental feature that will help with that and this blog post gives a preview of how it works.
Be warned, though: it is still experimental and we expect the interface to change over time.
We'll be adding these examples to the Rust unstable book, as well, and we plan to keep that up to date, so you can always find the latest instructions there.

- [ ] update the Rust unstable book with these examples so we can link directly to it in the last sentence. 

- [ ] update the dates in the post to be 2020-02-18 because I do not anticipate having time today or tomorrow to make these changes and @nikomatsakis  is out on Monday.

posts/inside-rust/2020-02-18-intro-rustc-self-profile.md
```diff
@@ -0,0 +1,253 @@
...
+This creates a file called `chrome_profiler.json` in the working directory.
+To open it, we'll use the regular Chromium performance tools you might already be familiar with:
+
+1. Open Chrome
+2. Open the Developer Tools console by pressing `Ctrl` + `Shift` + `i` (Windows/Linux) or `Cmd` + `Option` + `i` (macOS)
+3. Click the Performance tab at the top of the console.
+4. Click the "Load profile" button which looks like an arrow pointing up.
+5. Select the `chrome_profiler.json` file we created.
+
+You should now see something similar to this:
```
andjo403: it is also possible to do:
1. Open Chrome
2. Go to chrome://tracing,
3. Click the “Load” button
4. Select the `chrome_profiler.json` file we created.

this will also show the trace file but gives some other possibilities that I have not found for the Developer Tools version. eg. mark one event and gets sum of all events with that name and some more statistics

wesleywiser: Yes, that's true. I've used that in the past with large trace files which caused the regular tools to choke but it seemed to work fine in this case so I thought it best to keep this as simple as possible. We should document this though probably in the Crox docs.

wesleywiser:
Updated post for today's date.

nikomatsakis:
Whoops, sorry for delay, merging!

PR #529, opened 2020-03-02
blog post: recent & future pattern matching improvements

Centril:
r? @nikomatsakis


Centril:
NOTE: the file name needs to be adjusted depending on when this is merged.

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
```diff
@@ -0,0 +1,234 @@
...
+
+Pattern matching in Rust works by checking if a [*place*][ref_place] in memory (the "data") matches a certain *pattern*. In this post, we will look at some recent improvements to patterns soon available in stable Rust as well as some more in already available in nightly.
+
+## Subslice patterns, `[head, tail @ ..]`
+
+[fixed_slice]: https://blog.rust-lang.org/2018/05/10/Rust-1.26.html#basic-slice-patterns
+[recover_attrs_no_item]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/parser/struct.Parser.html#method.recover_attrs_no_item
+[pr_subslice]: https://github.com/rust-lang/rust/pull/67712
+
+Lists, are one of the most basic and common data structures found in software. In Rust, lists are usually a contiguous sequence of elements in memory, or a *slice*.
```
Mark-Simulacrum:
```suggestion
Lists are one of the most basic and common data structures found in software. In Rust, lists are usually a contiguous sequence of elements in memory, or a *slice*.
```

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
```diff
@@ -0,0 +1,234 @@
...
+}
+```
+
+Previously, when using `|` in a `match` expression, the `|` syntax was part of `match` itelf. With `or_patterns`, this is now part of patterns themselves, so you can nest OR-patterns arbitrarily, and use them in `let` statements too:
+
+```rust
+let Ok(x) | Err(x) = foo();
+```
+
+An OR-pattern covers the *union* of all the `|`-ed ("or-ed") patterns. To ensure that whatever alternative matched, all bindings are consistent and initialized, each or-ed pattern must include the exact same set of bindings, with the same types, and the same binding modes.
```
Mark-Simulacrum: Why do we need the same binding modes? I would've expected e.g. it be fine to have `Ok(ref x) | Err(x)` with `Result<u32, &u32>` producing `x: &u32`.

Is this just a match/borrow check limitation?

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
```diff
@@ -0,0 +1,234 @@
...
+```rust
+#![feature(move_ref_pattern)]
+
+fn main() {
+    let tup = ("foo".to_string(), 0);
+    let (x, ref y) = tup;
+}
+```
+
+## How can I help?
```
Mark-Simulacrum: I wonder if linking here or something from the top is worthwhile, i.e., have a call to action immediately, and then get into the background information.

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
Centril: It's possible that there's no good reason for this limitation (which is encoded in librustc_resolve) to exist. I'm not exactly sure why it does, so my answer is the non-answer "this is how it has always been" ;)

```
error[E0409]: variable `x` is bound in inconsistent ways within the same match arm
 --> src/main.rs:2:25
  |
2 |     let Ok(ref x) | Err(x) = Ok(0);
  |                -        ^ bound in different ways
  |                |
  |                first binding
```
Maybe @petrochenkov or @matthewjasper knows why.

(Note to self: the diagnostic should not mention "match arm".)

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
```diff
@@ -0,0 +1,234 @@
...
+[ref_match]: https://doc.rust-lang.org/reference/expressions/match-expr.html
+[ref_pat]: https://doc.rust-lang.org/reference/patterns.html
+[ref_place]: https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions
+
+Much of writing software revolves around checking if some data has some shape ("pattern"), extracting information from it, and then reacting if there was a match. To facilitate this, many modern languages, Rust included, support what is known as "pattern matching".
+
+> If you are new to Rust or want to refresh your knowledge, you may first want to read chapters [6, Enums and Pattern Matching][ch_6] and [18, Patterns and Matching][ch_18] in the book, or read more about [`match` expressions][ref_match] and [patterns][ref_pat] in the reference.
+
+Pattern matching in Rust works by checking if a [*place*][ref_place] in memory (the "data") matches a certain *pattern*. In this post, we will look at some recent improvements to patterns soon available in stable Rust as well as some more in already available in nightly.
+
```
Centril:
```suggestion

If you are familiar with the nightly features discussed and would like to help out with the efforts to drive them to stable, jump ahead to [*How can I help?](#how-can-i-help?).

```
@Mark-Simulacrum Something like this? (Not sure the link is correct wrt. the `?`)

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
Centril: I like the idea. Perhaps with something like https://github.com/rust-lang/blog.rust-lang.org/pull/529#discussion_r387346540 would do?

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
Mark-Simulacrum: Sure, that'll work.

`?` won't hurt because in the worst case it's just a `?foo` starter and it'll not bother the browser I think.

posts/inside-rust/2020-03-02-recent-future-pattern-matching-improvements.md
matthewjasper: The compiler doesn't currently track binding modes separately for each binding. Since `Ok(x) | Err(mut x)` has to be an error, and `Ok(ref x) | Err(&ref x)` is allowed, there isn't much reason to change this.

PR #533, opened 2020-03-04
Match colours with main website

XAMPPRocky:
This changes the colours on the blog to match the new colours on the website. I'm also proposing changing from using the background colour for the posts feed to gray, as I feel that works better with the post list than the purple (though I don't mind the the purple). I've put in a comparison screenshots below.

### Old / New Purple
Screenshot 2020-03-04 at 10 52 31

### Old Purple / New Gray
Screenshot 2020-03-04 at 10 54 32

### New Purple / New Gray
Screenshot 2020-03-04 at 10 55 38

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -3,11 +3,11 @@
 $body-font: 'Fira Sans', Helvetica, Arial, sans-serif;
 $header-font: 'Alfa Slab One', serif;
 
-$gray: #454C52;
-$red: #C14566;
-$green: #398277;
-$purple: #403D58;
-$yellow: #FFD45E;
+$gray: #2a3439;
+$red: #a72145;
+$green: #0b7261;
+$purple: #2e2459;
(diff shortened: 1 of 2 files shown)


pietroalbini:
How would it work if we removed the background completly?

XAMPPRocky:
This is just a colour swap, this would obviously need spacing changes as well, but I'm not fan of it. Mainly I think because I don't like Fira Sans as a font enough to think it can stand alone. Maybe if we had a different font and maybe other elements it could work.

### Current / White
Screenshot 2020-03-04 at 14 35 37

### Fira Sans / System Default (San Francisco on macOS)
Screenshot 2020-03-04 at 14 47 00

XAMPPRocky:
@pietroalbini Do you approve of the new gray  for the blog feed background?

pietroalbini:
Yeah, the new grey is fine.

PR #534, opened 2020-03-05
Add initial draft of side effect post

Mark-Simulacrum:
I tried to draft the state of the world as I know it, but I'm sure I missed things or at least it feels like I am :) There's probably also some room for headings... but I'm not quite sure yet how those should look.

Happy to hear feedback!

cc @nikic 
r? @nikomatsakis

### DIFF
--- a/posts/inside-rust/2020-03-19-terminating-rust.md
+++ b/posts/inside-rust/2020-03-19-terminating-rust.md
@@ -40,9 +40,10 @@ will materialize in upstream LLVM that allows us to opt-out in the short term.
 
 However, having said that, side effect itself is likely improvable to at least
 avoid the excessive consecutive calls, as demonstrated by this [IR][IR-test]
-that occur after LLVM optimizations. It seems plausible that those improvements
-may also reduce the compile time hit that we see when enabling side effect on
-the rustc side. Having said that, how simple these improvements are is unclear.
+that occurs after LLVM optimizations. It seems plausible that those
+improvements may also reduce the compile time hit that we see when enabling
+side effect on the rustc side. Having said that, how simple these improvements
+are is unclear.
 
 We would love to hear feedback and suggestions on how to resolve this problem!
 Please leave feedback on this internals thread (link to be filled in right
--- a/posts/inside-rust/2020-03-19-terminating-rust.md
+++ b/posts/inside-rust/2020-03-19-terminating-rust.md
@@ -1,0 +1,54 @@
+---
+layout: post
+title: "Resolving Rust's forward progress guarantees"
+author: Mark Rousskov
+description: "Should side-effect be the fix?"
+team: the compiler team <https://www.rust-lang.org/governance/teams/compiler>
+---
+
+There has been a longstanding miscompilation in Rust: programs that do not make
+[forward progress]. Note that the previous link is to the C++ definition; Rust
+is not C++, but currently LLVM optimizes all LLVM IR with the assumption that a
+lack of forward progress is undefined behavior.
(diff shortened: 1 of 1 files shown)


posts/inside-rust/2020-03-19-terminating-rust.md
```diff
@@ -0,0 +1,51 @@
...
+
+Note also that Rust does not define a lack of forward progress as [undefined
+behavior], while C++ does. It is particularly common to encounter the
+miscompilation "intentionally" when writing panic handlers and other such code
+with a body of `loop {}`. Some users also report that they've unintentionally
+hit this bug in recursive code which accidentally lacks a base case.
+
+Somewhat recently, LLVM added an intrinsic which tells the optimizer that
+forward progress has been made. On nightly Rust, you can enable this with
+`-Zinsert-sideeffect`.
```
nikomatsakis: maybe give a bit more detail here -- i.e., "this will cause the compiler to automatically insert calls to the intrinsic in places that LLVM might consider to be infinite loops." or something.

posts/inside-rust/2020-03-19-terminating-rust.md
```diff
@@ -0,0 +1,51 @@
...
+opt-out of the forward progress guarantees. It seems unlikely that a solution
+will materialize in upstream LLVM that allows us to opt-out in the short term.
+
+However, having said that, side effect itself is likely improvable to at least
+avoid the excessive consecutive calls, as demonstrated by this [IR][IR-test]
+that occur after LLVM optimizations. It seems plausible that those improvements
+may also reduce the compile time hit that we see when enabling side effect on
+the rustc side. Having said that, how simple these improvements are is unclear.
+
+We would love to hear feedback and suggestions on how to resolve this problem!
```
nikomatsakis: Maybe we can make a more pointed call for help -- e.g., open an internals thread and ask people to comment there?

Mark-Simulacrum:
cc @rust-lang/compiler -- before we post this (tentatively next Thursday), would be good to get some more eyes on it to see if we can perhaps find some concrete questions to ask or if the wording anywhere is misleading.

posts/inside-rust/2020-03-19-terminating-rust.md
```diff
@@ -0,0 +1,54 @@
...
+when necessary.
+
+Upstream, in LLVM, discussion has been ongoing for some time around whether, and
+how to, adjust LLVM's model to permit frontends for languages like Rust to
+opt-out of the forward progress guarantees. It seems unlikely that a solution
+will materialize in upstream LLVM that allows us to opt-out in the short term.
+
+However, having said that, side effect itself is likely improvable to at least
+avoid the excessive consecutive calls, as demonstrated by this [IR][IR-test]
+that occur after LLVM optimizations. It seems plausible that those improvements
```
pnkfelix: nit: "occur" should be "occurs" (at least, my understanding is that "IR" is singular, not plural).

posts/inside-rust/2020-03-19-terminating-rust.md
nikomatsakis: @Mark-Simulacrum I don't think we did anything here, any thoughts? I think we should at least create an internals thread, or maybe direct people to a github issue. Otherwise who knows how they will try to give us feedback.

Mark-Simulacrum: (Discussed a bit on Zulip -- we're going to merge, post on internals, and then update this post in quick succession).

PR #548, opened 2020-03-15
Add docs.rs opt-in single target post

jyn514:
r? @pietroalbini


XAMPPRocky:
Would this be more appropiate on the main blog? Seems like something for *users* of Rust rather than *contributors* of Rust.

jyn514:
Sure, I can move it there. @pietroalbini suggested an inside-rust blog which is why I put it there.

XAMPPRocky:
@jyn514 If they already suggested it for inside rust, then it seems fine. Though I don't know if I like our current trend of putting things that aren't contributor related in Inside Rust because they're not a "big" announcement.

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
```diff
@@ -0,0 +1,73 @@
...
+# This also sets the default target to `x86_64-pc-windows-msvc`
+targets = ["x86_64-pc-windows-msvc", "i686-pc-windows-msvc"]
+```
+
+If you only need a single target, it's even simpler:
+
+```toml
+[package.metadata.docs.rs]
+# This builds only the default target (usually `x86_64-unknown-linux-gnu`)
+targets = []
```
pietroalbini: I would not recommend using the empty target list. While it's supported to prevent unnecessary build failures, we should encourage people to fill the default target in if they use `targets`.

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+See the [docs.rs documentation][metadata] for more details about how to opt-in.
+
+## How does this help docs.rs?
+
+Building all crates from crates.io can take a long time!
+Building fewer targets will allow us to reduce wait times for every crate.
+This will avoid delaying every crate by hours when e.g. rusoto publishes a new release.
```
pietroalbini: Let's not "blame" crates if possible.

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
```diff
@@ -0,0 +1,73 @@
...
+and additionally will be built for all other tier one targets.
+However, because of this, you need to opt-in in order for us to reduce our build costs.
+
+We are considering turning this on by default in the future;
+i.e. only building for one target unless multiple targets are specifically requested.
+However, we do not want to break anyone's documentation, so in the meantime the feature is opt-in instead of opt-out.
+
+## How can I learn more?
+
+- For more details about targets and what it means to be a tier-one target,
```
pietroalbini: I'd prefer to avoid bullet points here.

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
jyn514: I think this will help increase adoption by making it easier to opt-in, though. If we change it to suggest `targets = ["x86_64-unknown-linux-gnu"]`, then someone might type it by hand and spell it wrong, failing the build for no reason.

jyn514:
> I'd change the blog post to instead explain what this feature does and who can benefit from it (crates with large build times, so they can have the docs online more quickly), hinting at the future plans of allowing arbitrary targets and defaulting to just linux (as most crates don't expose a different API between platforms).

That sounds reasonable. I'll write that up.

jyn514:
That makes more sense why you would want it in Inside Rust as opposed to the main blog as well.

jyn514:
Addressed review comments

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
pietroalbini: I don't think this will prevent adoption too much, as people will copy/paste this snippet anyway. Also, your misspelling concern would apply to every other target people could add.

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
```diff
@@ -0,0 +1,76 @@
...
+
+This change will also make it easier for docs.rs to build
+for targets that are not tier one, such as embedded targets.
+
+## How can I learn more?
+
+For more details about targets and what it means to be a tier-one target, see [platform support].
+For details about the change, see the PR: [docs.rs#632].
+For details about the motivation, see [docs.rs#343].
+For details about non-tier-one targets, see [docs.rs#563].
```
pietroalbini:
```suggestion
You can learn more about the change in [the issue proposing it][docs.rs#343] and [the PR with the implementation][docs.rs#632]. Details on building non-tier-one targets are also available [in the issue requesting the feature][docs.rs#563].

More information on targets and what it means to be a tier-one target is available in the [platform support] page.
```

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
```diff
@@ -0,0 +1,76 @@
...
+Instead of building for every tier-one target, you can build for only a single target,
+reducing your documentation times by a factor of 6.
+This can especially help large crates or projects with many crates
+that take several hours to document.
+
+## How does this help docs.rs?
+
+Building all crates from crates.io can take a long time!
+Building fewer targets will allow us to reduce wait times for every crate.
+Additionally, this will decrease our fixed storage costs, allowing docs.rs to be sustainable into the future.
```
pietroalbini: Reading this, it seems like docs.rs is not sustainable atm and this change is required for it to be sustainable.

```suggestion
Additionally, this will decrease the growth of our storage costs, improving the sustainability of the project.
```

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Possible future changes
+
+We are considering turning this on by default in the future;
+i.e. only building for one target unless multiple targets are specifically requested.
+However, we do not want to break anyone's documentation, so in the meantime the feature is opt-in instead of opt-out.
```
pietroalbini:
```suggestion
However, we do not want to break anyone's documentation, so we're making this feature opt-in  while we decide the migration strategy.
```

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
jyn514: Ok, I'll change it to `targets = ["x86_64-unknown-linux-gnu"]`.

jyn514:
Addressed comments, how does this look?

posts/2020-03-15-docs-rs-opt-into-fewer-targets.md
```diff
@@ -0,0 +1,76 @@
...
+# and only builds that target
+targets = ["x86_64-unknown-linux-gnu"]
+```
+
+See the [docs.rs documentation][metadata] for more details about how to opt-in.
+
+## How does this help my crate?
+
+Instead of building for every tier-one target, you can build for only a single target,
+reducing your documentation times by a factor of 6.
```
pietroalbini:
```suggestion
reducing your documentation build times by a factor of 6.
```

PR #566, opened 2020-04-17
Blog post for 1.43.0

steveklabnik:
r? @rust-lang/release

### DIFF
--- a/posts/2020-04-23-Rust-1.43.0.md
+++ b/posts/2020-04-23-Rust-1.43.0.md
@@ -1,0 +1,103 @@
+---
+layout: post
+title: "Announcing Rust 1.41.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.43.0. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.43.0 is as easy as:
+
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.43.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1430-2020-04-23
+
+## What's in 1.43.0 stable
+
+This release is fairly minor. There are no new major features. We have some
+new stabilized APIs, some compiler performance improvements, and a small
+macro-related feature. See the [detailed release notes][notes] to learn about
+other changes not covered by this post.
+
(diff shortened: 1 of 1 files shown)


posts/2020-04-23-Rust-1.43.0.md
```diff
@@ -0,0 +1,103 @@
+---
+layout: post
+title: "Announcing Rust 1.41.0"
```
jonas-schievink, line 3:
```suggestion
title: "Announcing Rust 1.43.0"
```

pietroalbini:
I'd also mention [the new `CARGO_BIN_EXE_<name>`](https://doc.rust-lang.org/nightly/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) environment variables, as those are great for people writing integration tests for CLIs.

posts/2020-04-23-Rust-1.43.0.md
```diff
@@ -0,0 +1,103 @@
...
+## What's in 1.43.0 stable
+
+This release is fairly minor. There are no new major features. We have some
+new stabilized APIs, some compiler performance improvements, and a small
+macro-related feature. See the [detailed release notes][notes] to learn about
+other changes not covered by this post.
+
+### `item` fragments
+
+You can use `item` fragments to interpolate items into the body of traits,
```
BatmanAoD, line 35:
```suggestion
In macros, you can use `item` fragments to interpolate items into the body of traits,
```

posts/2020-04-23-Rust-1.43.0.md
```diff
@@ -0,0 +1,103 @@
...
+impls, and extern blocks. For example:
+
+```rust
+macro_rules! mac_trait {
+    ($i:item) => {
+        trait T { $i }
+    }
+}
+mac_trait! {
+    fn foo() {}
```
BatmanAoD, line 45: I don't quite understand the explanation, nor the [issue](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1430-2020-04-23) it refers to.

Is this the expansion of the macro?

```
trait T {
    fn foo() {}
}
```

In any case, the release notes say that this is "still rejected semantically"; unless there's code that now compiles that wouldn't have compiled before, I'm not sure it's worth mentioning in the blog post.

If there's an example where the error message has improved, I suppose we could approach the change from that angle.

steveklabnik: This is the expansion of the macro, I thought it was obvious but I guess it's not!

I think the release notes were out of order; the lines *before* this one are, but this is not rejected, it does compile.

steveklabnik: (I've pushed a bit showing the expansion explicitly, thank you!)

steveklabnik:
I believe all review comments are now addressed.

Issue #567, opened 2020-04-17
warning from CI when adding a new blog post with wrong date


nikomatsakis:
Since blog post filenames must include the date, and posts often stay in draft form for days, it's easy to merge with the file having the wrong name. It'd be so great if we could detect and prevent that somehow.

Mark-Simulacrum:
One question I have had historically is whether we can just drop that requirement and instead have some intelligent "when was the file added to git" support or something like that. We ~never merge things ahead of time or into the past, anyway...

PR #573, opened 2020-04-21
Port Blog Backend To Zola

XAMPPRocky:
### [Rendered](https://xampprocky.github.io/blog.rust-lang.org/)

This switches the current bespoke blog backend for Zola. This should provide us access with more features to build and customise the blog without having to build all of those features outselves. There's still a few things needed to be done before this ready to merge, but the main structure should be ready for review. The two main commits are "Make Zola Templates" and "Initial Zola Config", the rest is moving files around, changing frontmatter syntax.

For contributors of the blog there is one major change which is that new blog urls are now generated at `/<name>/` instead of `/YYYY/MM/DD/<name>/` (Previous blog entries still work at on old URL). I made this trade off as it preserves the current two directory level structure we currently have. We could also perserve the  `/YYYY/MM/DD/<name>/` url formatting but change to a more nested directory layout.

### To Do
- [x] Update CI
- [ ] Fix CSS on Inside Rust page

r? @pietroalbini  || @Manishearth


Manishearth:
I'm not sure if we should be changing the URL, even if it's done backwards compatibly. cc @steveklabnik ?

Have we diffed the output of the old and new website?

Manishearth:
I'm not sure if we should be changing the URL, even if it's done backwards compatibly. cc @steveklabnik ?

Have we diffed the output of the old and new website?

Manishearth:
I'm not sure if we should be changing the URL, even if it's done backwards compatibly. cc @steveklabnik ?

Have we diffed the output of the old and new website?

XAMPPRocky:
FWIW I'd also prefer keeping at least some of current URL scheme. This was just the most compatible solution. The one problem I would see with keeping the current scheme is having to create a directory with date (`2020/04/05/compiler-meetings.md`). Though it could be said we should move away from the date in the filesystem and move it inside the file so that someone is less likely to merge with the wrong date (#567).

XAMPPRocky:
You should now be able to view the updated version at https://xampprocky.github.io/blog.rust-lang.org/

steveklabnik:
Yes so, the custom backend here was because I needed to hit a deadline, I generally support moving it to something less bespoke.

I would prefer to not change URLs unless it's absolutely necessary. We have a *lot* of seo juice on the blog.

Manishearth:
Yeah, I think we should try to keep the URLs as is, such changes should not be made lightly

XAMPPRocky:
A change would be required to switch to Zola, since Zola doesn't currently doesn't support that kind of scheme without being inconvenient for contributors. While I do think this port merits changes in the URL scheme, I want to write more motivation for that change. In the mean time I've filed a [feature request on Zola][fr] for adding the transformation to see if the problem resolves itself.

[fr]: https://zola.discourse.group/t/add-option-to-create-directories-from-dates/392

XAMPPRocky:
I'm going to close this PR for now, as I messed up the branch, and I want to write on the motivation before revisiting this PR.

PR #577, opened 2020-04-29
Use GitHub Action for deployment.

XAMPPRocky:
This allows using the repo's `GITHUB_TOKEN` and will only push when there's actual changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,13 +17,11 @@ jobs:
           path: target
           key: cargo-build-${{ hashFiles('**/Cargo.lock') }}
           restore-key: cargo-build-
-      - name: Build Blog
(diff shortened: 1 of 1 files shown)


pietroalbini:
r? @pietroalbini 

This mostly looks fine. While reviewing the PR, I noticed both the new and the old versions don't create a [`.nojekyll`](https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/) file at the root of the deployed website, while that file should be created to avoid extra preprocessing by GitHub.

Could you add that? A simple `touch site/.nojekyll` should be enough.

XAMPPRocky:
@pietroalbini Done.

pietroalbini:
Thanks!

PR #592, opened 2020-05-15
Remove preloads of font files that do not exist

gszy:


nikomatsakis:
Seems fine to me but I'm not really familiar with this stuff

r? @XAMPPRocky

XAMPPRocky:
Yep this is fine, I already had this change in a local branch but never pushed it. We import the fonts through css.

PR #600, opened 2020-05-25
Blog post on new inline assembly syntax

joshtriplett:


joshtriplett:
cc @Amanieu for review.

posts/inside-rust/2020-05-25-new-inline-asm.md
```diff
@@ -0,0 +1,130 @@
...
+Notice that `value` and `popcnt` have registers selected for them, while
+`bits.as_mut_ptr()` must go in the `rdi` register for use with the `stosb`
+instruction.
+
+For full details on the new `asm!` syntax, see [RFC
+2873](https://github.com/Amanieu/rfcs/blob/inline-asm/text/0000-inline-asm.md).
+Please try it out, and [report any bugs via the rust issue
+tracker](https://github.com/rust-lang/rust/issues/) with the tag `F-asm`.  You
+can also [report your experiences on the tracking
+issue](https://github.com/rust-lang/rust/issues/72016).
```
Mark-Simulacrum, line 130: Should we perhaps instead direct to an internals or Zulip thread? I expect that the tracking issue is otherwise likely to relatively quickly get lots of comments, and "protecting" it seems worthwhile.

joshtriplett: I think I'll just drop the reference to the tracking issue entirely. People can already find it via the other github links if they're dedicated. We should also have a note at the top of the tracking issue saying something like "If you came here to report a bug in inline assembly, please report it as a separate issue and link to this one; don't post a comment here".

posts/inside-rust/2020-05-25-new-inline-asm.md
```diff
@@ -0,0 +1,130 @@
...
+#![feature(asm)]
+
+fn main() {
+    let buf = "Hello from asm!\n";
+    let ret: i32;
+    unsafe {
+        asm!(
+            "syscall",
+            in("rax") 1, // syscall number
+            in("rdi") 1, // fd
```
Amanieu, line 64:
```suggestion
            in("rdi") 1, // fd (stdout)
```

posts/inside-rust/2020-05-25-new-inline-asm.md
```diff
@@ -0,0 +1,130 @@
...
+                xor {v}, rax
+                tzcnt rax, rax
+                stosb
+                jmp 2b
+                1:
+                ",
+                v = inout(reg) value => _,
+                popcnt = out(reg) popcnt,
+                out("rax") _, // scratch
+                inout("rdi") bits.as_mut_ptr() => _,
```
Amanieu, line 109: `options(nostack)` could be used here.

Amanieu:
I'd like to hold the blog post back a bit while we sort out the various ICEs with the new inline asm (#72570, #72484). At least until we get fixes for those into nightly.

posts/inside-rust/2020-05-25-new-inline-asm.md
joshtriplett: This doesn't seem to make any difference in the generated code.

Along the same lines, I could use `option(readonly)` in the `syscall` example, but I don't think doing so adds any value. These are supposed to be very simple examples, and people will have the full RFC available for reference.

joshtriplett:
@Amanieu Seems reasonable to me. And thanks for the feedback; I've updated the post.

posts/inside-rust/2020-05-25-new-inline-asm.md
```diff
@@ -0,0 +1,135 @@
...
+despite being one of the most requested features.
+
+In an effort to improve `asm!` and bring it to more users, [Amanieu
+d'Antras](https://github.com/Amanieu) designed and implemented a new,
+friendlier syntax for `asm!`. This syntax has had a long road from concept to
+compiler implementation:
+- The proposal first started as a [pre-RFC on
+  internals](https://internals.rust-lang.org/t/pre-rfc-2-inline-assembly/11310).
+- Inline assembly became one of the language team's first [project
+  group](https://github.com/rust-lang/rfcs/blob/master/text/2836-project-asm.md),
```
nikomatsakis, line 37:
```suggestion
  groups](https://github.com/rust-lang/rfcs/blob/master/text/2836-project-asm.md),
```

joshtriplett: Good catch, thank you.

joshtriplett:
@nikomatsakis Thanks for reviewing and approving.

@Amanieu Please post here as soon as you're ready for this to be posted.

nikomatsakis:
Please ping me before merging this PR, thanks, =)

joshtriplett:
@Amanieu It looks like both of the issues you referenced have been fixed. Are there any other things you want to wait for before this gets posted?

Amanieu:
I think it should be fine to post now. There are a few pending issues but they are only related to diagnostics (https://github.com/rust-lang/rust/pull/72825 and https://github.com/rust-lang/rust/issues/40555) and won't affect usability.

nikomatsakis:
Nope, let's post.

PR #614, opened 2020-06-09
Mention `option(att_syntax)`

joshtriplett:
Feedback on the blog post suggests that people weren't aware they had
the option to use AT&T syntax. Mention it, particularly for folks who
may be translating existing inline assembly code to the new `asm!`
syntax.


joshtriplett:
cc @Amanieu

Based on various comments I've seen about the blog post.

Amanieu:
LGTM

PR #623, opened 2020-06-20
inline-asm: Update for new style

joshtriplett:
https://github.com/rust-lang/rust/pull/73364 implemented support for
providing multiple lines of assembly as separate arguments to `asm!`;
update the blog post to use that new syntax, so that people who find it
will use that style as an example.

### DIFF
--- a/posts/inside-rust/2020-06-08-new-inline-asm.md
+++ b/posts/inside-rust/2020-06-08-new-inline-asm.md
@@ -95,14 +95,14 @@ fn main() {
         let popcnt;
         unsafe {
             asm!(
-                "    popcnt {popcnt}, {v}",
+                "popcnt {popcnt}, {v}",
                 "2:",
-                "    blsi rax, {v}",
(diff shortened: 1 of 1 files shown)


joshtriplett:
cc @Amanieu

XAMPPRocky:
@joshtriplett LGTM, should I wait for @Amanieu's review or should I just merge it?

Amanieu:
I don't think we should indent the asm code inside the strings (i.e. keep the labels and instructions on the same indent level).

joshtriplett:
@Amanieu I don't have a strong preference, I just wanted to show the ability to do so. Happy to change it.

Amanieu:
You should update the playground link as well.

joshtriplett:
@Amanieu Good catch, done.

Amanieu:
LGTM

XAMPPRocky:
Thank you for your PR!

PR #631, opened 2020-07-01
Make the blogs highlight style inline with the website

XAMPPRocky:
Companion PR to https://github.com/rust-lang/www.rust-lang.org/pull/1165

### Production

Screenshot 2020-07-01 at 10 17 27

### Proposed

Screenshot 2020-07-01 at 10 17 15

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -300,10 +300,11 @@ header h1 {
 
 
 .highlight {
-  height: 30px;
+  height: 12px;
   position: relative;
-  top: -30px;
+  top: 0;
   left: -10px;
   width: 120%;
   max-width: 90vw;
+  border-radius: 2px;
 }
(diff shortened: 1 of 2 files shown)


lzutao:
Does the change on phone still produce the same ouptput?

XAMPPRocky:
Screenshot 2020-07-01 at 10 55 47

nikomatsakis:
I don't think I'm a good assignee here!

r? @Manishearth perhaps?

Manishearth:
As I mentioned in https://github.com/rust-lang/www.rust-lang.org/pull/1165 , I believe the highlight style chosen here was a deliberate choice. I've seen this kind of styling before across the web.

skade:
Given that there's positive review on the website topic (https://github.com/rust-lang/www.rust-lang.org/pull/1165#issuecomment-653898386), I assume this one is good, too.

XAMPPRocky:
Yep, but this shouldn’t be merged yet, the www PR needs to merge and then the website needs to be deployed so that the change happens simultaneously. I also have to make a PR to thanks.

PR #632, opened 2020-07-01
Create CODEOWNERS

XAMPPRocky:
This creates a `CODEOWNERS` file to replace highfive for assigning reviewers for the blog as currently highfive just always assigns to @nikomatsakis  which seems a bit unfair.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-* @website
-posts/* @website @nikomatsakis
+* @rust-lang/website
+posts/* @rust-lang/website @nikomatsakis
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @website
+posts/* @website @nikomatsakis
```
pietroalbini, line 2: Wouldn't this assign to the `@website` user instead of the team?

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @rust-lang/website
+posts/* @rust-lang/website @nikomatsakis
```
pietroalbini, line 2: The ownership for the posts is not supposed to be just niko, but the core team.

pietroalbini:
Other than changing niko with core, adding this won't hurt so r=me.

PR #640, opened 2020-07-08
posts: Rustup 1.22.1 release

kinnison:
This is an embarassing `.1` release as they all are.

I've pushed the `stable` build for `rustup` and asked @pietroalbini on Zulip about when we might be able to make the release, so this text needs vetting and checking, but I imagine Pietro will handle merging when we've made the release.


tshepang:
Debian folk call it *brown paper bag release*

kinnison:
> Debian folk call it _brown paper bag release_

Aye, we do :brown_heart:

PR #646, opened 2020-07-09
Blog post for 06 04 governance meeting

valgrimm:


XAMPPRocky:
Discussed in private and we're going to release a longer form post that catches up on the past three including today's meetings.

PR #649, opened 2020-07-14
Rust 1.45.0 release announcement

steveklabnik:
cc @rust-lang/release @rust-lang/core 

Also, cc @SergioBenitez since Rocket features here. Please let me know if you'd like anything changed!

### DIFF
--- a/posts/2020-07-16-Rust-1.45.0.md
+++ b/posts/2020-07-16-Rust-1.45.0.md
@@ -1,0 +1,296 @@
+---
+layout: post
+title: "Announcing Rust 1.45.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.45.0. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.45.0 is as easy as:
+
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.45.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1450-2020-07-16
+
+## What's in 1.45.0 stable
+
+There are two big changes to be aware of in Rust 1.45.0: a fix for some
+long-standing unsoundness when casting between integers and floats, and the
+stabilization of the final feature needed for one of the more popular web
+frameworks to work on stable Rust.
+
+## Fixing unsoundness in casts
+
+[Issue 10184](https://github.com/rust-lang/rust/issues/10184) was originally
+opened back in October of 2013, a year and a half before Rust 1.0. As you may
+know, `rustc` uses [LLVM](http://llvm.org/) as a compiler backend. When you
+write code like this:
+
+```rust
+pub fn cast(x: f32) -> u8 {
+    x as u8
+}
+```
+
+The Rust compiler in Rust 1.44.0 and before would produce LLVM-IR that looks
+like this:
+
+```llvm-ir
+define i8 @_ZN10playground4cast17h1bdf307357423fcfE(float %x) unnamed_addr #0 {
+start:
+  %0 = fptoui float %x to i8
+  ret i8 %0
+}
+```
+
+That `fptoui` implements the cast, it is short for "floating point to
+unsigned integer."
+
+But there's a problem here. From [the
+docs](https://llvm.org/docs/LangRef.html#fptoui-to-instruction):
+
+> The ‘fptoui’ instruction converts its floating-point operand into the
+> nearest (rounding towards zero) unsigned integer value. If the value cannot
+> fit in ty2, the result is a poison value.
+
+Now, unless you happen to dig into the depths of compilers regularly, you may
+not understand what that means. It's full of jargon, but there's a simpler
+explanation: if you cast a floating point number that's large to an integer
+that's small, you get undefined behavior.
+
+That means that this, for example, was not well-defined:
+
+```rust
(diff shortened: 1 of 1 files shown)


posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -0,0 +1,296 @@
...
+```rust
+let x: f32 = 1.0;
+let y: u8 = unsafe { x.to_int_unchecked() };
+```
+
+But as always, you should only use this method as a last resort. Just like
+with array access, the compiler can often optimize the checks away, making
+the safe and unsafe versions equivalent when the compiler can prove it.
+
+## Stabilizing function-like procedrual macros in expressions, patterns, and statements
```
Manishearth, line 150:
```suggestion
## Stabilizing function-like procedural macros in expressions, patterns, and statements
```

posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -0,0 +1,296 @@
...
+fn hello(name: String, age: u8) -> String {
+    format!("Hello, {} year old named {}!", age, name)
+}
+
+fn main() {
+    rocket::ignite().mount("/", routes![hello]).launch();
+}
+```
+
+At the same time, Rocket was only able to achieve its power and ergonomics by
```
Manishearth, line 223: the "at the same time" feels weird here

posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -0,0 +1,296 @@
...
+
+In the end, the decision was made to do this:
+
+* `as` would perform a "saturating cast".
+* A new `unsafe` cast would be added if you wanted to skip the checks.
+
+This is very similar to array access, for example:
+
+* `array[i]` will check to make sure that `array` as at least `i` elements
+* You can use `unsafe { array.get_unchecked(i) }` to skip the check
```
XAMPPRocky, line 107:
```suggestion
* `array[i]` will check to make sure that `array` has at least `i` elements.
* You can use `unsafe { array.get_unchecked(i) }` to skip the check.
```

posts/2020-07-16-Rust-1.45.0.md
XAMPPRocky: I agree, I think the sentence works just as well without it.
```suggestion
Rocket was only able to achieve its power and ergonomics by
```

posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+But as always, you should only use this method as a last resort. Just like
+with array access, the compiler can often optimize the checks away, making
+the safe and unsafe versions equivalent when the compiler can prove it.
+
+## Stabilizing function-like procedrual macros in expressions, patterns, and statements
+
+In [Rust 1.30.0](https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html), we stabilized
+"function-like procedrual macros in item position." For example, [the
```
CryZe, line 153:
```suggestion
"function-like procedural macros in item position." For example, [the
```

posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -0,0 +1,296 @@
...
+
+Being able to use macros in more places is interesting, but there's another
+reason why many Rustaceans have been waiting for this feature for a long
+time. [Rocket](https://rocket.rs/) is a popular web framework for Rust.
+Initially released in December of 2016, it has often been described as one of
+the best things the Rust ecosystem has to offer. Here's the "hello world"
+example from its homepage:
+
+```rust
+#![feature(proc_macro_hygiene, decl_macro)]
```
SergioBenitez, line 209: Would it be better to use the expected example from `0.5`?

```rust
#[macro_use] extern crate rocket;

#[get("/<name>/<age>")]
fn hello(name: String, age: u8) -> String {
    format!("Hello, {} year old named {}!", age, name)
}

#[launch]
fn rocket() -> rocket::Rocket {
    rocket::ignite().mount("/hello", routes![hello])
}
```

If so, perhaps the text above could change to:

> Being able to use macros in more places is interesting, but there's another
reason why many Rustaceans have been waiting for this feature for a long time:
[Rocket](https://rocket.rs/). Inititally released in December of 2016, Rocket is
a popular web framework for Rust often described as one of the best things the
Rust ecosystem has to offer. Here's the "hello world" example from its upcoming release:

pietroalbini: @SergioBenitez done!

SergioBenitez: The next paragraph will also need to change to accommodate this change. Drafting a suggestion now.

posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -0,0 +1,296 @@
...
+
+Rocket was only able to achieve its power and ergonomics by
+using nightly-only features. You can see two of them in the example above.
+However, as you may guess from the name of those two features, these are the
+ones that are finally stable! [This
+issue](https://github.com/SergioBenitez/Rocket/issues/19) has been tracking
+the history of all of the various features, and now they're all checked off!
+
+Rocket 0.5 has not yet been released, but when it is, many folks will be very
+happy :)
```
SergioBenitez, line 230:
```suggestion
Until today, Rocket depended on nightly-only features to deliver on its promise
of flexibility and ergonomics. In fact, as can be seen on the [project's
homepage](https://rocket.rs/v0.4), the same example above in the current version
of Rocket requires the `proc_macro_hygiene` feature to compile. However, as you
may guess from the feature's name, today it ships in stable! [This
issue](https://github.com/SergioBenitez/Rocket/issues/19) tracked the history of
nightly-only features in Rocket. Now, they're all checked off!

This next version of Rocket is still in the works, but when released, many folks
will be very happy :)
```

steveklabnik:
🎊

PR #653, opened 2020-07-16
Tweak actions config

JohnTitor:
- use actions/cache v2, it has some (performance) improvements
- `restore-key` is wrong, should be `restore-keys`


XAMPPRocky:
Thank you for your PR!

PR #654, opened 2020-07-16
Fixed wording in Rust 1.45 Blog Post

JosephTLyons:
This sentence read a little awkwardly to me, so I tried to make it more clear.

### DIFF
--- a/posts/2020-07-16-Rust-1.45.0.md
+++ b/posts/2020-07-16-Rust-1.45.0.md
@@ -87,7 +87,7 @@ fn main() {
 
 On Rust 1.44.0, this happens to print "x: 0" on my machine. But it could
 print anything, or do anything: this is undefined behavior. But the `unsafe`
-keyword has not be specified in this code block. This is what we call a
+keyword is nowhere to be found within this block of code. This is what we call a
 "soundness" bug, that is, it is a bug where the compiler does the wrong thing.
 We tag these bugs as
 [I-unsound](https://github.com/rust-lang/rust/issues?q=is%3Aissue+is%3Aopen+label%3A%22I-unsound+%F0%9F%92%A5%22)
--- a/posts/2020-07-16-Rust-1.45.0.md
+++ b/posts/2020-07-16-Rust-1.45.0.md
@@ -86,10 +86,10 @@ fn main() {
 ```
 
 On Rust 1.44.0, this happens to print "x: 0" on my machine. But it could
-print anything, or do anything: this is undefined behavior. But we have
-no `unsafe` code here. This is what we call a "soundness" bug, that is,
-it is a bug where the compiler does the wrong thing. We tag these bugs
-as
+print anything, or do anything: this is undefined behavior. But the `unsafe`
+keyword has not be specified in this code block. This is what we call a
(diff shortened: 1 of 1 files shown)


lzutao:
How about "there is not `unsafe` code" here?

posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -86,10 +86,10 @@ fn main() {
 ```
 
 On Rust 1.44.0, this happens to print "x: 0" on my machine. But it could
-print anything, or do anything: this is undefined behavior. But we have
-no `unsafe` code here. This is what we call a "soundness" bug, that is,
-it is a bug where the compiler does the wrong thing. We tag these bugs
-as
+print anything, or do anything: this is undefined behavior. But the `unsafe`
+keyword has not be specified in this code block. This is what we call a
+"soundness" bug, that is, it is a bug where the compiler does the wrong thing.
```
ntravis, line 91:
```suggestion
keyword has not been specified in this code block. This is what we call a
"soundness" bug, where the compiler does the wrong thing.
```

minor suggestion re be vs been. using that is in the middle of a sentence is pretty weird as well, so maybe we can get two fixes in one.

steveklabnik: This suggestion is good re: grammar, but using "specified" here sounds odd, because specification has a different meaning, technically. 

I can appreciate this sentence maybe sounding odd, but I think moving it in this direction makes it sound *more* odd, not less.

JosephTLyons: Oops, yeah the "be" was a typo.  Ironic.

JosephTLyons: What about

But the `unsafe` keyword is nowhere to be found within this block of code.

?

nikomatsakis: I prefer this latter formulation, fwiw.

JosephTLyons:
I pushed up the 2nd proposal for the change.  Feel free to close this PR if there are remaining objections to it.  Thanks guys.

posts/2020-07-16-Rust-1.45.0.md
```diff
@@ -86,10 +86,10 @@ fn main() {
 ```
 
 On Rust 1.44.0, this happens to print "x: 0" on my machine. But it could
-print anything, or do anything: this is undefined behavior. But we have
-no `unsafe` code here. This is what we call a "soundness" bug, that is,
-it is a bug where the compiler does the wrong thing. We tag these bugs
-as
+print anything, or do anything: this is undefined behavior. But the `unsafe`
+keyword is nowhere to be found within this block of code. This is what we call a
```
nikomatsakis, line 90:
```suggestion
keyword is not used within this block of code. This is what we call a
```
Re-reading, "nowhere to be found" seems a bit dramatic, what about this?

posts/2020-07-16-Rust-1.45.0.md
steveklabnik: @josephtlyons yeah that sounds good to me!

PR #661, opened 2020-07-30
1.45.2 post

Mark-Simulacrum:
This is targeting tomorrow right now, but that may be adjusted in the future.

### DIFF
--- a/posts/2020-07-31-Rust-1.45.2.md
+++ b/posts/2020-07-31-Rust-1.45.2.md
@@ -32,7 +32,7 @@ appropriate page on our website, and check out the [detailed release notes for
 Trait objects and `#[track_caller]` could lead to unsoundness. `#[track_caller]`
 is not yet stable on 1.45. However, the standard library makes use of this on
 some traits for better error messages. Trait objects of `SliceIndex`, `Index`,
-and `IndexMut` are affected by this bug.
+and `IndexMut` were affected by this bug.
 
 ## Bindings in tuple-patterns
 
--- a/posts/2020-07-31-Rust-1.45.2.md
+++ b/posts/2020-07-31-Rust-1.45.2.md
@@ -1,0 +1,48 @@
+---
+layout: post
+title: "Announcing Rust 1.45.2"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.45.2. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.45.2 is as easy as:
+
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install] from the
+appropriate page on our website, and check out the [detailed release notes for
+1.45.2][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1452-2020-07-31
+
+## What's in 1.45.2 stable
+
+1.45.2 contains two fixes, one to 1.45.1 and the other to 1.45.0.
+
+## `#[track_caller]` on trait objects
+
+Trait objects and `#[track_caller]` could lead to unsoundness. `#[track_caller]`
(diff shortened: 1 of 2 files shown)


posts/2020-07-31-Rust-1.45.2.md
```diff
@@ -0,0 +1,48 @@
...
+1.45.2 contains two fixes, one to 1.45.1 and the other to 1.45.0.
+
+## `#[track_caller]` on trait objects
+
+Trait objects and `#[track_caller]` could lead to unsoundness. `#[track_caller]`
+is not yet stable on 1.45. However, the standard library makes use of this on
+some traits for better error messages. Trait objects of `SliceIndex`, `Index`,
+and `IndexMut` are affected by this bug.
+
+## Bindings in tuple-patterns
```
jsgf, line 37: The bug actually manifested in a slice pattern.

posts/2020-07-31-Rust-1.45.2.md
```diff
@@ -0,0 +1,48 @@
...
+## What's in 1.45.2 stable
+
+1.45.2 contains two fixes, one to 1.45.1 and the other to 1.45.0.
+
+## `#[track_caller]` on trait objects
+
+Trait objects and `#[track_caller]` could lead to unsoundness. `#[track_caller]`
+is not yet stable on 1.45. However, the standard library makes use of this on
+some traits for better error messages. Trait objects of `SliceIndex`, `Index`,
+and `IndexMut` are affected by this bug.
```
cuviper, line 35: If we're fixing this, then I suggest past tense.

```suggestion
and `IndexMut` were affected by this bug.
```

posts/2020-07-31-Rust-1.45.2.md
cuviper: Well, the backported fix was for tuple patterns, and the regression was in a slice pattern. So we're going back to the status quo from 1.45.0, and I guess that ICE will re-emerge for tuples.

Mark-Simulacrum: Hm, I think the bug is something like "slice patterns in tuple patterns" or something like that. I'm not sure how to pinpoint it.

cuviper: Oh, because `Option::Some(_)` is a tuple variant? OK, that makes sense, and I just checked that there's no issue for slice patterns alone.

cuviper: Maybe "Nested bindings in tuple patterns"?

Mark-Simulacrum: The condition leads me to believe it needs to be a tuple struct (or tuple enum variant, I guess) specifically with a `..` pattern bound to a variable in it. Saying all of that in the heading seems too verbose, though. Perhaps "Tuple patterns containing bindings to `..`"? That also feels rather long.

posts/2020-07-31-Rust-1.45.2.md
```diff
@@ -0,0 +1,48 @@
+---
+layout: post
+title: "Announcing Rust 1.45.2"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.45.2. Rust is a
```
NieDzejkob: I'm not sure how I feel about the wording of "happy to announce". In a way, the release is a result of previous mistakes, so perhaps it's not something we should be celebrating?

Mark-Simulacrum: I think this is fine - we are in some sense happy to fix the bug at least. If you have alternative suggestions though I'd be happy to consider them.

tshepang: > The Rust team announces...

NieDzejkob: > > The Rust team announces...

I was also considering this general direction, though I had the impression that the correct tense would be "The Rust team is announcing". I'm not a native speaker, though :upside_down_face:

Mark-Simulacrum: Pushed an update.

PR #663, opened 2020-07-31
Make code ownership more fine grained

Manishearth:
Trying to reduce the amount of notifications I and core team members get.

 - Core only needs to own the toplevel posts. Inside-rust posts can and should be reviewed by the relevant team, and they can ask someone on core to merge.
 - The website team probably should not own post contents, those are reviewed by core or the relevant team.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,4 @@
-* @rust-lang/website
-posts/* @rust-lang/core
-posts/inside-rust/* @rust-lang/website @rust-lang/core
+src/* @rust-lang/website
+static/* @rust-lang/website
+templates/* @rust-lang/website
+posts/*.md @rust-lang/core


XAMPPRocky:
@Manishearth It would probably also help to enable the round robin feature for code review, so that individuals get assigned rather than the whole team. You have to set that on GitHub.

Screenshot 2020-07-31 at 21 45 31

pietroalbini:
> It would probably also help to enable the round robin code review feature for code review, so that individuals get assigned rather than the whole team. You have to set that on GitHub.

I'm not sure we want to assign any random core team member to blog posts.

nikomatsakis:
I'm not familiar with CODEOWNERS but this seems reasonable -- is this still relevant? cc @pietroalbini weren't you doing some tinkering with this?

pietroalbini:
Yeah we already did these changes.

PR #671, opened 2020-08-14
Link to RSS feeds

### LINKED ISSUE #420: Call out RSS feeds for the main and the Inside Rust blogs
I can't for the life of me find the RSS feed link on either of the blogs, but I found the main blog feed link in #307 and I figured out the Inside Rust blog using it as well:

- Main Blog: https://blog.rust-lang.org/feed.xml
- Inside Rust Blog: https://blog.rust-lang.org/inside-rust/feed.xml

Would be cool to have links to these guys on the pages itself.

camelid:
Fixes #420.

### DIFF
--- a/posts/inside-rust/2020-08-24-1.46.0-prerelease.md
+++ b/posts/inside-rust/2020-08-24-1.46.0-prerelease.md
@@ -1,0 +1,21 @@
+---
+layout: post
+title: "1.46.0 pre-release testing"
+author: Pietro Albini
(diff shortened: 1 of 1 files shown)


camelid:
I see that there's https://github.com/rust-lang/blog.rust-lang.org/issues/307#issuecomment-532761627, but in my experience many websites will say that they have an RSS feed. However, if you would rather have just the machine-readable link, #420 should be probably be closed.

PR #676, opened 2020-08-24
Add 1.46.0 pre-release post

pietroalbini:
The pre-release is not yet out, but I expect it to be available in less than 30 minutes (looking at the current logs on RCS). Once it's out we can publish this.

r? @Mark-Simulacrum


pietroalbini:
The pre-release is out, we can merge the blog post.

Mark-Simulacrum:
We received feedback last time that we should tell people where to put feedback -- I suggest a link to an internals thread or #t-release on Zulip perhaps? Could you add one of those (the internals thread requiring a dance of merging, posting internals threads, and pushing an update to the blog post)?

PR #683, opened 2020-09-01
Use the `disc` list style type

camelid:
Previously, it was the `circle` list style type, but it was determined
that the `disc` list style type is in many ways a better match for Rust.

See rust-lang/www.rust-lang.org#1252 and rust-lang/www.rust-lang.org#1254.

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -194,6 +194,10 @@ ul, ol {
   padding-left: 1.2em;
 }
 
+ul {
+  list-style-type: disc;
+}
+
 .posts {
   background-color: $gray;
   color: white;


camelid:
Was this the right place to edit the styles? It's all in one line, which made me think it might not be, but I don't see any other place.

steveklabnik:
This file is imported from http://getskeleton.com/, so it is not the right place to modify the style.

camelid:
So where should I change it if not here? Should I add a new CSS file that overrides this style?

rylev:
@camelid I guess you were waiting to learn where to put these custom styles no? Do you still have any interest in seeing this land?

camelid:
@rylev Yeah, I'm not sure where to change the style. It would be nice to see this through :)

Turbo87:
@camelid I've rebased this and moved the override to `src/style/app.scss` where we're already other styles

camelid:
@Turbo87 wow, thanks so much! Glad this change has been made at last. :)

PR #685, opened 2020-09-04
const evaluatable future compat warning

lcnr:
blocked on https://github.com/rust-lang/rust/pull/74595. Might still need some improvements :shrug: 

r? @nikomatsakis

### DIFF
--- a/posts/inside-rust/2020-09-04-Const-Evaluatable.md
+++ b/posts/inside-rust/2020-09-04-Const-Evaluatable.md
@@ -1,0 +1,51 @@
+---
+layout: post
+title: "Const evaluatable feature compatability"
+author: Bastian Kauschke
+team: the lang team <https://www.rust-lang.org/governance/teams/lang>
+---
+
+In version 1.43.0 we accidentially allowed the usage of generic parameters in the length of repeat expressions
+as long as they are not relevant for the final value of the constant. We now emit a future compatability lint
+in these cases which will get changed to a hard error in the far future. There might be a subset of affected code
+which will instead be allowed without requiring any changes, but this still needs further consideration.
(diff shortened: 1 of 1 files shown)


nikomatsakis (approved):

nikomatsakis:
This seems pretty good to me -- cc @rust-lang/lang

posts/inside-rust/2020-09-04-Const-Evaluatable.md
```diff
@@ -0,0 +1,51 @@
+---
+layout: post
+title: "Const evaluatable feature compatability"
```
therealprof, line 3: Evaluatable is a pretty strange word, did you mean evaluable?

lcnr: they apparently mean the same thing and we use "evaluatable" inside of the compiler here.

Not sure why this is the case though :thinking:

therealprof: According to typical dictionaries (e.g. Merriam-Webster) it's not a real English word.The only place where it seems to be used is Wolfram Research software.

jyn514: Maybe this should just say 'evaluation' so it's more clear?

lcnr:
do we still want to publish this post? :laughing:

posts/inside-rust/2020-09-04-Const-Evaluatable.md
lqd: BTW: 
```suggestion
title: "Const evaluatable feature compatibility"
```

PR #701, opened 2020-10-01
[inside-rust]: Update self-profile blog post for measureme 9.0

wesleywiser:
We just released measureme 0.8 which now records trace data into one
file instead of three different ones. I've updated the blog post to
reflect this since many people still use it as a getting started
tutorial.

### DIFF
--- a/posts/inside-rust/2020-02-25-intro-rustc-self-profile.md
+++ b/posts/inside-rust/2020-02-25-intro-rustc-self-profile.md
@@ -46,30 +46,30 @@ $ rustup update nightly
 ## Profiling the compiler
 
 Now we can build it and tell `rustc` to profile the build of the `regex` crate.
-This will cause three new files to be created in the working directory which contain the profling data.
+This will cause a new file to be created in the working directory which contains the profiling data.
 
 ```sh
 $ cargo rustc -- -Zself-profile
 $ ls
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (commented):
I think I might rather see us publish a new one with similar/identical contents but add a note to each saying "for version .x" and linking to the newest one.

wesleywiser:
I think I agree in the general case, but the only change here is that the file extensions are different and there is only one file instead of three. IMO having two articles that are so similar would be more confusing than helpful. 

I think it's possible make these edits in such a way that the commands would work for any version of the tools. Would that be better?

Mark-Simulacrum:
Hm. I think we should add a note at least that we've updated the article for a more recent release and point at git history for the old version.

I don't think we should try to make it compatible with both.

wesleywiser:
I've added a note that links to the current version of the post via Github.

rylev:
@wesleywiser we're on measureme 10.0.0 now. Shall we close this?

wesleywiser:
I'd be happy to spend some time updating it for 10.0 if we're fine with the idea of updating the article in general. If we'd rather just leave it as it was originally published, then we should close this.

rylev:
I am personally a fan of creating a new blog entry and only updating the old post to prominently display a link to the new entry. However, I don't hold this opinion super strongly, so just take this as my 2cents.

PR #705, opened 2020-10-07
Add post for second 1.47.0 pre-release

pietroalbini:
This adds another Inside Rust post to announce the second pre-release.


pietroalbini:
Note: the second pre-release is not out yet.

PR #714, opened 2020-10-16
backlog bonanza post

nikomatsakis:
cc @rust-lang/lang / @joshtriplett in particular -- Some folks were asking about backlog bonanza so I thought I'd write something up about what we've been up to. I am wondering if I should split this into two posts, or change the subject line, but maybe it works as is.


posts/inside-rust/2020-10-15-Backlog-Bonanza.md
```diff
@@ -0,0 +1,107 @@
...
+[weekly triage meetings]: https://lang-team.rust-lang.org/meetings.html
+[path]: https://blog.rust-lang.org/inside-rust/2020/07/09/lang-team-path-to-membership.html
+
+### Chartering a project group
+
+A "project group" is basically just a group of people working together
+completing some idea. Project groups are often pretty small, just 1 or 2
+people, though they can get significantly larger.
+
+Creating a smaller project group is meant to be lightweight. We
```
nikomatsakis, line 54: I didn't really go into the idea that we have some project groups that are more formal and some that are less. I'm wondering about this distinction and how I feel about it. =)

I am wondering if I should avoid calling things "project groups" and just talk about "projects", where sometimes we also have a group associated with it and sometimes we don't.

Or maybe we should just make it easier to create a project group on the team repo and run with it.

posts/inside-rust/2020-10-15-Backlog-Bonanza.md
```diff
@@ -0,0 +1,107 @@
...
+series of meetings in which we go through every pending RFC, one by
+one, and try to reach some sort of determination about what to do with
+it.  Once we've finished that, we can start in on categorizing other
+forms of backlog, such as tracking issues.
+
+### Possible outcomes for each RFC
+
+When we look at an RFC, we're typically deciding between one of the following outcomes:
+
+* **Close** the RFC, if we think that this problem is a good fit for the moment.
```
Mark-Simulacrum, line 19: Hm, maybe misinterpreting, but this feels like awkward phrasing? Maybe something like "solution is the wrong one and/or problem does not seem currently high priority"?

nikomatsakis: good catch, also, it was supposed to be "this problem is **not** a good fit for the moment"

nikomatsakis:
I think I'll just post this.

PR #719, opened 2020-10-30
Inside Rust: const eval roadmap

oli-obk:
I'm writing this in https://hackmd.io/7Rs5q0h4TImeOGbv9NYxcQ 

This is the blog post discussed in the language team meeting at https://github.com/nikomatsakis/lang-team/blob/master/design-meeting-minutes/2020-06-24-const-eval-unsafe-and-skill-tree.md

cc @rust-lang/wg-const-eval @joshtriplett

### DIFF
--- a/posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
+++ b/posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
@@ -1,0 +1,28 @@
+# Const eval roadmap
+
+Y'all probably noticed that const eval finally got loops, mutable local variables and branches. With those features existing, we were able to make a lot of basic arithmetic functions `const fn `. Things like `checked_div` needs to be able to check that the divisor is not zero and return `None` in that case. These features have been asked for a lot, and while it may not appear so, these were the low hanging fruit.
+
+So you may now be thinking "If the low hanging fruit took you 5 years" *(yes oli-obk has been working on const eval for that long, and in the end ecstaticmorse implemented const control flow)*, "will we see any of the more fancy features in this new decade?".
+
+TLDR: Yes, and we now have a system for making things go fast (brrr?) while still making sure we don't accidentally bypass the decision processes.
+
+Instead of creating more and more [RFCs](https://github.com/rust-lang/rfcs/) or [MCPs](https://github.com/rust-lang/lang-team/) and DDOSing the language team, we made one roadmap that doesn't actually have the details on each feature. The plan is a tree-graph that lists the dependencies between features. The language team then signs off on the roadmap, giving the const-eval team the permission to experiment with unstable features without checking back with the language team. Stabilization of any of the features will still work by having a language team sign off, but that can then be done with the full knowledge about the final product instead of hypotheticals. If new features that are not yet on the plan are desired to get implemented, the roadmap is adjusted and the language team signs off on the roadmap changes.
+
+This new scheme makes it much easier to keep an overview over how features interact with other features. So without further adue (and hoping that the previous sentence's statement is true), let's introduce the
+
+[Const. Eval. Skill. Tree.](https://rust-lang.github.io/const-eval/)
+
+All the nodes are hyperlinks, pointing to the tracking issue for a specific feature. The edges turn bold and pink when you hover over them, making it easy to track which nodes it connects. If you want to see the source that generates this graph, you can find it at [in the const-eval team repository](https://github.com/rust-lang/const-eval/blob/master/src/skill_tree.md).
+
+Nodes further on the right are higher level features depending on features whose nodes are further left. Arrows point from low level features to high level features. So an Arrow is basically pointing from a feature to the features that are enabled by it.
+
+Looking at the current state of the roadmap, you can see that there are three major low level features:
+
+* trait bounds on `const fn`
+    * is being worked on, and a preliminary subset of it has been implemented. 
+* mutable references
+    * is being worked on, basically works, we just need to hammer out some kinks
+* unsafe code
+    * RalfJung is in the process of writing up a document explaining how we can reliably detect most undefined behaviour in const eval code (and the kind of UB we cannot detect).
+
+So, as you can see, things are moving along nicely and we'll likely be able to give you new const eval features to play on nightly with soon.
--- a/posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
+++ b/posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
@@ -5,7 +5,7 @@ author: Oli Scherer
 team: the const eval working group <https://github.com/rust-lang/const-eval/>
 ---
 
-Y'all probably noticed that const eval finally got loops, mutable local variables and branches. With those features stabilized, we were able to make a lot of basic arithmetic functions `const fn` on the stable compiler. These features have been asked for a lot, and while it may not appear so, these were the low hanging fruit.
+Y'all probably noticed that const eval finally got loops, mutable local variables and branches. With those features stabilized, we were able to make a lot of checked and saturating arithmetic functions `const fn` on the stable compiler. These features have been asked for a lot, and while it may not appear so, these were the low hanging fruit.
 
 So you may now be thinking "If the low hanging fruit took you 5 years" *(yes oli-obk has been working on const eval for that long, and in the end ecstaticmorse implemented const control flow)*, "will we see any of the more fancy features in this new decade?".
 
--- a/posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
+++ b/posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
@@ -1,12 +1,17 @@
-# Const eval roadmap
+---
+layout: post
+title: "Constant evaluation roadmap"
+author: Oliver Scherer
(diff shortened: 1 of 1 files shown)


posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
```diff
@@ -0,0 +1,28 @@
...
+
+This new scheme makes it much easier to keep an overview over how features interact with other features. So without further adue (and hoping that the previous sentence's statement is true), let's introduce the
+
+[Const. Eval. Skill. Tree.](https://rust-lang.github.io/const-eval/)
+
+All the nodes are hyperlinks, pointing to the tracking issue for a specific feature. The edges turn bold and pink when you hover over them, making it easy to track which nodes it connects. If you want to see the source that generates this graph, you can find it at [in the const-eval team repository](https://github.com/rust-lang/const-eval/blob/master/src/skill_tree.md).
```
lqd, line 15:
```suggestion
All the nodes are hyperlinks, pointing to the tracking issue for a specific feature. The edges turn bold and pink when you hover over them, making it easy to track which nodes it connects. If you want to see the source that generates this graph, you can find it [in the const-eval team repository](https://github.com/rust-lang/const-eval/blob/master/src/skill_tree.md).
```

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
```diff
@@ -0,0 +1,28 @@
+# Const eval roadmap
```
lqd, line 1:
```suggestion
---
layout: post
title: "Const eval roadmap"
author: Oli Scherer
team: the const eval working group <https://github.com/rust-lang/const-eval/>
---

```

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
```diff
@@ -0,0 +1,28 @@
...
+
+TLDR: Yes, and we now have a system for making things go fast (brrr?) while still making sure we don't accidentally bypass the decision processes.
+
+Instead of creating more and more [RFCs](https://github.com/rust-lang/rfcs/) or [MCPs](https://github.com/rust-lang/lang-team/) and DDOSing the language team, we made one roadmap that doesn't actually have the details on each feature. The plan is a tree-graph that lists the dependencies between features. The language team then signs off on the roadmap, giving the const-eval team the permission to experiment with unstable features without checking back with the language team. Stabilization of any of the features will still work by having a language team sign off, but that can then be done with the full knowledge about the final product instead of hypotheticals. If new features that are not yet on the plan are desired to get implemented, the roadmap is adjusted and the language team signs off on the roadmap changes.
+
+This new scheme makes it much easier to keep an overview over how features interact with other features. So without further adue (and hoping that the previous sentence's statement is true), let's introduce the
```
wesleywiser, line 11: Nit

```suggestion
This new scheme makes it much easier to keep an overview over how features interact with other features. So without further ado (and hoping that the previous sentence's statement is true), let's introduce the
```

wesleywiser (approved):
Exciting! 🎉

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
```diff
@@ -0,0 +1,33 @@
+---
+layout: post
+title: "Constant evaluation roadmap"
+author: Oli Scherer
+team: the const eval working group <https://github.com/rust-lang/const-eval/>
+---
+
+Y'all probably noticed that const eval finally got loops, mutable local variables and branches. With those features existing, we were able to make a lot of basic arithmetic functions `const fn `. Things like `checked_div` needs to be able to check that the divisor is not zero and return `None` in that case. These features have been asked for a lot, and while it may not appear so, these were the low hanging fruit.
```
RalfJung, line 8: I don't understand the purpose of the 2nd sentence here ("Things like..."). It seems unrelated to the first and third sentence, and just cutting it doesn't leave a gap as far as I can see.

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
```diff
@@ -0,0 +1,33 @@
...
+
+TLDR: Yes, and we now have a system for making things go ~~fast~~ brrr while still making sure we don't accidentally bypass the decision processes.
+
+Instead of creating more and more [RFCs](https://github.com/rust-lang/rfcs/) or [MCPs](https://github.com/rust-lang/lang-team/) and DDOSing the language team, we made one roadmap that doesn't contain the actual details of each feature. The plan is a tree-graph that lists the dependencies between features. The language team then signs off on the roadmap, giving the const-eval team the permission to experiment with unstable features without checking back with the language team. Stabilization of any of the features will still work by having a language team sign off, but that can then be done with the full knowledge about the final product instead of hypotheticals. If new features that are not yet on the plan are desired to get implemented, the roadmap is adjusted and the language team signs off on the roadmap changes.
```
RalfJung, line 14: What's a "tree-graph"?^^

RalfJung:
One question about the skill tree itself: So far, I think we removed (or planned to remove) nodes when they got stabilized. Would it make more sense to instead change their color or so? That would (a) keep the useful dependency information around, and (b) give a nice sense of progress as the entire graph slowly changes color. :D

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
camelid, line 8: I think you have an extra space here.

```suggestion
Y'all probably noticed that const eval finally got loops, mutable local variables and branches. With those features existing, we were able to make a lot of basic arithmetic functions `const fn`. Things like `checked_div` needs to be able to check that the divisor is not zero and return `None` in that case. These features have been asked for a lot, and while it may not appear so, these were the low hanging fruit.
```

oli-obk: I didn't know where to put an example... Maybe I could just say

> basic arithmetic functions (e.g. `checked_div`) `const fn`

?

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
oli-obk: well... just a tree I guess XD

oli-obk:
> That would (a) keep the useful dependency information around, and

Is it still useful? There's no dependency anymore. It's more historical information. Maybe I could extend the skill tree generator to output two graphs? One with all nodes (and outdated ones grayed out or something) and one with just the still active nodes.

> (b) give a nice sense of progress as the entire graph slowly changes color. :D

I like progress being the graph having less parts ^^ Like the entire graph is just the things we still have to work at without keeping parts in it that are done.

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
RalfJung: Yes, that seems much better. Also `checked_div` seems rather niche, isn't there something "bigger"?

RalfJung:
> I like progress being the graph having less parts ^^ Like the entire graph is just the things we still have to work at without keeping parts in it that are done.

I expect more nodes will be added as we discover more things we want to do during CTFE. So we might make progress without the graph getting smaller. Or is that not how you intended this to go?

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
jyn514: I think `unwrap_or` works now and didn't before, anything with Option or Result really.

oli-obk: I chose `checked_div` because it's pretty obvious what it does. the `Option`/`Result` methods may still have some blockers somewhere, not sure if we fixed/stabilized the drop fixes needed for that.

oli-obk:
> So we might make progress without the graph getting smaller. Or is that not how you intended this to go?

No, that is indeed how I intended this to work.

If we keep around old nodes the graph will grow indefinitely, even if it contains things that have been stable for a long time and some ppl on the project may not even know that this didn't work some time in the past. If we remove things (maybe after a grace period), the graph will only be the (future?) roadmap, and not a general overview over the past and future features of const eval.

What do you think about going with both our preferred systems and just making two graphs out of the same skill tree file?

posts/inside-rust/0000-00-00-const-eval-roadmap-and-skill-tree.md
RalfJung: What about "(e.g. checked and saturating arithmetic)"? That makes it less specific to "div".
(I am looking at https://blog.rust-lang.org/2020/08/27/Rust-1.46.0.html to see what got stabilized there.)

RalfJung:
If you don't think keeping the old nodes around is useful, I am okay with dropping them. We have the git history for historic information. I just felt like it could be an interesting thing to present, but I might just be wrong about it being interesting. ;)

rylev:
@oli-obk I'm guessing this has gone pretty stale. Shall we close this?

oli-obk:
Yea, I guess

PR #737, opened 2020-12-12
Add release flag to prior-1.32 blog posts

Folyd:


.gitignore
```diff
@@ -5,3 +5,4 @@ site
 static/styles/vendor.css
 static/styles/app.css
 static/styles/fonts.css
+.idea
```
Mark-Simulacrum, line 8: We generally prefer not to include editor-specific files in the project .gitignore (I see .bk and .DS_Store here which seem like they could get removed by a separate PR, FWIW) -- can you drop this change?

src/main.rs
```diff
@@ -1,18 +1,20 @@
...
 use chrono::Timelike;
 use handlebars::{handlebars_helper, Handlebars};
 use sass_rs::{compile_file, Options};
 use serde_derive::Serialize;
 use serde_json::json;
-use std::convert::AsRef;
-use std::error::Error;
-use std::fs::{self, File};
-use std::io::Write;
-use std::path::{Path, PathBuf};
```
Mark-Simulacrum, line 15: Can we avoid whatever reformatting or shuffling was done here? It's going to make it harder to review and accept this PR -- and it shouldn't be necessary?

Mark-Simulacrum (commented):
Can you say more about your intent with this PR? It looks like it's aiming to add the version number for releases, but currently Rust infra doesn't consume that. I am not see any optimizations either, so not quite sure what you meant there.

I would prefer to drop the clippy and .gitignore changes from this PR regardless.

.gitignore
bjorn3: If you don't know it yet, you can have `$XDG_CONFIG_HOME/git/ignore` for global ignores and `.git/info/exclude` for not committed per-repo ignores.

Folyd:
> Can you say more about your intent with this PR? It looks like it's aiming to add the version number for releases, but currently Rust infra doesn't consume that. I am not see any optimizations either, so not quite sure what you meant there.
> 
> I would prefer to drop the clippy and .gitignore changes from this PR regardless.

Thanks for reviewing. I rectified the title of this PR and dropt the clippy and .gitignore changes.

Mark-Simulacrum (approved):

PR #738, opened 2020-12-14
Cache the target directory based on the compiler version

pietroalbini:
While Travis CI dynamically updated the caches after builds finished (with a different cache for each branch), GitHub Actions identifies each cache with a key and never updates a cache with the same key, even if the contents of the cached directory changed.

Before this PR, we cached the target directory based on the hash of the `Cargo.lock` file, which means the cache was never updated as long as the `Cargo.lock` didn't change. Unfortunately this was not enough, as the existing target directory is useless after a Rust version update. We also did not cache the crates.io index nor the downloaded dependencies.

This PR changes the CI configuration to also include the Rust version in the cache key. This also adds caching for the crates.io index and the downloaded crates. This should speed up CI from 2.5 minutes to 30 seconds.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,13 +11,35 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Cache cargo
+
+      - name: Detect the installed Rust version
+        id: rustc
+        run: echo "::set-output name=version::$(rustc -V)"
+
+      - name: Cache Cargo's registry cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.cargo/registry/cache
+          key: cargo-registry-cache-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            cargo-registry-cache-
+
+      - name: Cache Cargo's registry index
+        uses: actions/cache@v2
+        with:
+          path: ~/.cargo/registry/index
+          key: cargo-registry-index-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            cargo-registry-index-
+
+      - name: Cache Cargo's target directory
         uses: actions/cache@v2
         with:
           path: target
-          key: cargo-build-${{ hashFiles('**/Cargo.lock') }}
+          key: cargo-build-${{ steps.rustc.outputs.version }}-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
-            cargo-build-
+            cargo-build-${{ steps.rustc.outputs.version }}-
+
       - run: cargo run
       - run: cp CNAME ./site/
       - run: touch site/.nojekyll


.github/workflows/main.yml
```diff
@@ -11,13 +11,35 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Cache cargo
+
+      - name: Detect the installed Rust version
+        id: rustc
+        run: echo "::set-output name=version::$(rustc -V)"
```
jyn514, line 17: Was this deprecated as part of https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/ ? Or just `set-env` and set-output is still fine?

.github/workflows/main.yml
```diff
@@ -11,13 +11,35 @@ jobs:
...
+          path: ~/.cargo/registry/cache
+          key: cargo-registry-cache-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            cargo-registry-cache-
+
+      - name: Cache Cargo's registry index
+        uses: actions/cache@v2
+        with:
+          path: ~/.cargo/registry/index
+          key: cargo-registry-index-${{ hashFiles('**/Cargo.lock') }}
```
jyn514, line 31: Why key this on Cargo.lock? Even if the index is partially out of date, it will still be useful, right? Or is the idea to cache the latest version of the index possible?

.github/workflows/main.yml
```diff
@@ -11,13 +11,35 @@ jobs:
...
+      - name: Cache Cargo's target directory
         uses: actions/cache@v2
         with:
           path: target
-          key: cargo-build-${{ hashFiles('**/Cargo.lock') }}
+          key: cargo-build-${{ steps.rustc.outputs.version }}-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
-            cargo-build-
+            cargo-build-${{ steps.rustc.outputs.version }}-
+
```
jyn514, line 42: What does not including `Cargo.lock` in the restore key get you?

.github/workflows/main.yml
pietroalbini: Only `set-env` was deprecated.

.github/workflows/main.yml
pietroalbini: Yes, the goal is to cache the most recent clone of the index, to avoid having to clone it at all during a build. If the `Cargo.lock` is updated, the `restore-keys` will download the most recent cache available.

.github/workflows/main.yml
pietroalbini: GitHub Actions's caching works roughly this way when it tries to restore something:

* If a cache with the exact key is present (in this case, when both the Rust version and the `Cargo.lock` are the same) it will download that specific cache.
* If an exact cache match is not present, it will try downloading the latest cache matching the prefix in the `restore-keys`.

So, what this snippet does is it tries to download the cached target directory for the right combination of Rust version and `Cargo.lock`, and if that fails it will try downloading the latest cached target directory for that Rust version, regardless of the contents of the `Cargo.lock`

Manishearth (approved):

PR #743, opened 2020-12-17
Inside Rust: FFI-unwind group's exploration of `longjmp`-like functions

BatmanAoD:
The FFI-Unwind project group has started discussing how Rust can handle "`longjmp`-like functions". This blog post explains the problem space and our current plan for approaching it.

A less jokey title would probably be an improvement; I am very open to suggestions.


posts/inside-rust/2020-12-XX-ffi-unwind-longjmp.md
```diff
@@ -0,0 +1,214 @@
...
+which sounds superficially similar to functions that can be "canceled".
+However, cancelation is more restrictive: `-> !` usually indicates an infinite
+loop. This cannot cause a resource leak, so it is safe for non-POFs.
+
+## Join us!
+
+If you would like to help us create this specification and write an RFC for it,
+please join us in [zulip][proj-group-zulip]!
+
+#### Footnotes
```
BatmanAoD, line 202: Does the blog have built-in footnote support?

nikomatsakis: I...don't remember :)

posts/inside-rust/2020-12-XX-ffi-unwind-longjmp.md
```diff
@@ -0,0 +1,214 @@
...
+}
+```
+
+## Difference from the "never" type
+
+Above, a Rust call to C's `longjmp` function was declared with the "never"
+return type, `!`. A function with this return type cannot return normally,
+which sounds superficially similar to functions that can be "canceled".
+However, cancelation is more restrictive: `-> !` usually indicates an infinite
+loop. This cannot cause a resource leak, so it is safe for non-POFs.
```
nikomatsakis, line 195: Nit, pehaps? A `-> !` function can also panic.

nikomatsakis (approved):
Left one nit but this looks good

nikomatsakis:
Did you see this, @BatmanAoD ?

```
  Compiling blog v0.1.0 (/home/runner/work/blog.rust-lang.org/blog.rust-lang.org)
    Finished dev [unoptimized + debuginfo] target(s) in 13.67s
     Running `target/debug/blog`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', src/posts.rs:49:56
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jyn514:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', src/posts.rs:49:56

This happens when it can't parse a date from the file name, I've been meaning to give a better error. You'll have to change `12-XX` to be an actual day I think.

nikomatsakis (approved):

PR #745, opened 2020-12-28
Blog for Rust 1.49.0

steveklabnik:

### DIFF
--- a/posts/2020-12-31-Rust-1.49.0.md
+++ b/posts/2020-12-31-Rust-1.49.0.md
@@ -1,6 +1,6 @@
 ---
 layout: post
-title: "Announcing Rust 1.40.0"
+title: "Announcing Rust 1.49.0"
 author: The Rust Release Team
 release: true
 ---
--- a/posts/2020-12-31-Rust-1.49.0.md
+++ b/posts/2020-12-31-Rust-1.49.0.md
@@ -29,20 +29,48 @@ For this release, we have some new targets and an improvement to the test
 framework. See the [detailed release notes][notes] to learn about other
 changes not covered by this post.
 
-### aarch64 targets
+### 64-bit ARM Linux reaches Tier 1
 
-We have two big changes to three targets:
+The Rust compiler supports [a wide variety of targets][platform-support], but
+the Rust Team can't provide the same level of support for all of them. To
+clearly mark how supported each target is we created a tiering system:
 
-* [`aarch64-unknown-linux-gnu` is now Tier 1](https://github.com/rust-lang/rust/pull/78228)
-* [`aarch64-apple-darwin` is now Tier 2](https://github.com/rust-lang/rust/pull/75991)
-* [`aarch64-pc-windows-msvc` is now Tier 2](https://github.com/rust-lang/rust/pull/75914)
+* Tier 3 targets are technically supported by the compiler, but we don't check
+  whether their code build or passes the tests, and we don't provide any
+  prebuilt binaries as part of our releases.
+* Tier 2 targets are guaranteed to build and we provide prebuilt binaries, but
+  we don't execute the test suite on those platforms: the produced binaries
+  might not work or might have bugs.
+* Tier 1 targets provide the highest support guarantee, and we run the full
+  suite on those platforms for every change merged in the compiler. Prebuilt
+  binaries are also available.
 
-If you're not familiar with our "tiers" of support, you can read about that
-on our [platform support
-page](https://doc.rust-lang.org/stable/rustc/platform-support.html). In
-brief, Tier 1 is "guaranteed to work," Tier 2 is "guaranteed to build," and
-Tier 3 is "code exists but is not tested." If you're a 64-bit ARM user, these
-changes mean that Rust programs will work better on your platforms!
+Rust 1.49.0 promotes the `aarch64-unknown-linux-gnu` target to Tier 1 support,
+bringing our highest guarantees to users of 64-bit ARM systems running Linux!
+We expect this change to benefit workloads spanning from embedded to desktops
+and servers.
+
+This is an important milestone for the project: it's the first time a non-x86
+target reaches the highest level of Quality Assurance and support provided by
+the Rust Team, and we hope this will pave the way for more targets to reach
+Tier 1 in the future!
+
+Note that Android is not affected by this change as it uses a different Tier 2
+target.
+
+[platform-support]: https://doc.rust-lang.org/stable/rustc/platform-support.html
+
+### 64-bit ARM macOS and Windows reach Tier 2
+
+Rust 1.49.0 also features two targets reaching Tier 2 support:
+
+* The `aarch64-apple-darwin` target brings support for Rust on Apple M1 systems!
+* The `aarch64-pc-windows-mvc` target brings support for Rust on 64-bit ARM
+  devices running Windows on ARM!
+
+Developers can expect both of those targets to have prebuilt binaries
+installable with `rustup` from now on! The Rust Team is not running the test
+suite on those platforms though, so there might be bugs or instabilities.
 
 ### Test framework captures output in threads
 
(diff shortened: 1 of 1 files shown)


posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,169 @@
...
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1490-2020-12-31
+
+## What's in 1.49.0 stable
+
+For this release, we have some new targets and an improvement to the test
+framework. See the [detailed release notes][notes] to learn about other
+changes not covered by this post.
+
+### aarch64 targets
```
pietroalbini, line 32: Hmm, I feel like we should expand this. Having the first non-x86 platform reaching Tier 1 is a huge deal, and a ton of people are interested in support for the new Apple M1. Possible sketch of what content to put in:

* 64-bit ARM on Linux reached Tier 1 support
  * Every change done to the compiler now requires the full test suite to pass on the target.
  * First non-x86 target to be promoted to Tier 1!
  * Maybe: where the target could be used
* 64-bit ARM on Windows and Apple Silicon reached Tier 2 support!
  * Pre-built compilers are now guaranteed to be available through rustup
  * We're not running the compiler's test suite on those platforms, so there could be bugs in the compiler
  * Maybe: where the targets could be used

posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,169 @@
+---
+layout: post
+title: "Announcing Rust 1.40.0"
```
jonas-schievink, line 3:
```suggestion
title: "Announcing Rust 1.49.0"
```

CryZe:
What about the ability to bind by both move and ref now? Seems like a language change that people may want to know about.

posts/2020-12-31-Rust-1.49.0.md
steveklabnik: I checked FIVE TIMES, hahaha.

steveklabnik:
@CryZe I think it was a good change, but @pietroalbini didn't mention it when I asked him what he thought mattered, so I doubted myself and left it out, hehe.

posts/2020-12-31-Rust-1.49.0.md
steveklabnik: You had mentioned really wanting these to be two sections, is there a reason why? they feel like the same thing to me, though I do think expanding on it would be good.

pietroalbini:
> What about the ability to bind by both move and ref now?

I thought it was kinda niche, but if people want it featured I'm all for it!

posts/2020-12-31-Rust-1.49.0.md
pietroalbini: Mostly because the guarantees provided by Tier 1 and Tier 2 are so different that merging them together might confuse people. Also I kinda want to give the first non-x86 target the spotlight on its own :D

posts/2020-12-31-Rust-1.49.0.md
pietroalbini: Pushed a commit fixing this.

posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,197 @@
...
+
+For this release, we have some new targets and an improvement to the test
+framework. See the [detailed release notes][notes] to learn about other
+changes not covered by this post.
+
+### 64-bit ARM Linux reaches Tier 1
+
+The Rust compiler supports [a wide variety of targets][platform-support], but
+the Rust Team can't provide the same level of support for all of them. To
+clearly mark how supported each target is we created a tiering system:
```
jyn514, line 36:
```suggestion
clearly mark how supported each target is, we created a tiering system:
```

posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,197 @@
...
+
+* Tier 3 targets are technically supported by the compiler, but we don't check
+  whether their code build or passes the tests, and we don't provide any
+  prebuilt binaries as part of our releases.
+* Tier 2 targets are guaranteed to build and we provide prebuilt binaries, but
+  we don't execute the test suite on those platforms: the produced binaries
+  might not work or might have bugs.
+* Tier 1 targets provide the highest support guarantee, and we run the full
+  suite on those platforms for every change merged in the compiler. Prebuilt
+  binaries are also available.
```
jyn514, line 46: This is already in the platform-support page, right? Does it make sense to repeat it in the blog? I think the following paragraph ('highest guarantees') is good enough to understand why this is a big deal.

posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,197 @@
...
+
+Rust 1.49.0 promotes the `aarch64-unknown-linux-gnu` target to Tier 1 support,
+bringing our highest guarantees to users of 64-bit ARM systems running Linux!
+We expect this change to benefit workloads spanning from embedded to desktops
+and servers.
+
+This is an important milestone for the project: it's the first time a non-x86
+target reaches the highest level of Quality Assurance and support provided by
```
jyn514, line 54:
```suggestion
target reaches the highest level of support provided by
```

Mark-Simulacrum, line 54: I think my only nit is that we currently still limit crater to x86_64-unknown-linux-gnu; maybe we can just say "It's the first non-x86 target that reaches tier 1"?

posts/2020-12-31-Rust-1.49.0.md
pietroalbini: I'd prefer to keep it to provide more context for people who are not already familiar with our policy and who are going to miss the link.

pietroalbini:
Addressed all the feedback got so far.

posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,196 @@
...
+  suite on those platforms for every change merged in the compiler. Prebuilt
+  binaries are also available.
+
+Rust 1.49.0 promotes the `aarch64-unknown-linux-gnu` target to Tier 1 support,
+bringing our highest guarantees to users of 64-bit ARM systems running Linux!
+We expect this change to benefit workloads spanning from embedded to desktops
+and servers.
+
+This is an important milestone for the project, since it's the first time a
+non-x86 target reaches Tier 1 support: we hope this will pave the way for more
```
Mark-Simulacrum, line 54:
```suggestion
non-x86 target has reached Tier 1 support: we hope this will pave the way for more
```

pietroalbini: Fixed.

posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,196 @@
...
+
+For this release, we have some new targets and an improvement to the test
+framework. See the [detailed release notes][notes] to learn about other
+changes not covered by this post.
+
+### 64-bit ARM Linux reaches Tier 1
+
+The Rust compiler supports [a wide variety of targets][platform-support], but
+the Rust Team can't provide the same level of support for all of them. To
+clearly mark how supported each target is, we created a tiering system:
```
nagisa, line 36: Nit: we haven't created the system this release. Perhaps "we utilize a tiering system" would be a better way to phrase this?

posts/2020-12-31-Rust-1.49.0.md
```diff
@@ -0,0 +1,196 @@
...
+target.
+
+[platform-support]: https://doc.rust-lang.org/stable/rustc/platform-support.html
+
+### 64-bit ARM macOS and Windows reach Tier 2
+
+Rust 1.49.0 also features two targets reaching Tier 2 support:
+
+* The `aarch64-apple-darwin` target brings support for Rust on Apple M1 systems.
+* The `aarch64-pc-windows-mvc` target brings support for Rust on 64-bit ARM
```
nagisa, line 67: msvc

PR #771, opened 2021-01-28
Blog for Rust 1.50.0

cuviper:

### DIFF
--- a/posts/2021-02-11-Rust-1.50.0.md
+++ b/posts/2021-02-11-Rust-1.50.0.md
@@ -55,8 +55,8 @@ fn main() {
 ### `const` value repetition for arrays
 
 Arrays in Rust can be written either as a list `[a, b, c]` or a repetition `[x; N]`.
-For lengths `N` greater than one, repetition has only been allowed for `Copy`
-types, and [RFC 2203] sought to allow any `const` expression there. However,
+For lengths `N` greater than one, repetition has only been allowed for `x`s that are `Copy`,
+and [RFC 2203] sought to allow any `const` expression there. However,
 while that feature was unstable for arbitrary expressions, its implementation
 since Rust 1.38 accidentally allowed stable use of `const` _values_ in array
 repetition.
--- a/posts/2021-02-11-Rust-1.50.0.md
+++ b/posts/2021-02-11-Rust-1.50.0.md
@@ -75,8 +75,7 @@ fn main() {
 }
 ```
 
-In Rust 1.50, that stabilization is formally acknowledged, so you may use this
-construct without fear. In the future, to avoid such "temporary" named
+In Rust 1.50, that stabilization is formally acknowledged. In the future, to avoid such "temporary" named
 constants, you can look forward to inline `const` expressions per [RFC 2920].
 
 [RFC 2203]: https://rust-lang.github.io/rfcs/2203-const-repeat-expr.html
--- a/posts/2021-02-11-Rust-1.50.0.md
+++ b/posts/2021-02-11-Rust-1.50.0.md
@@ -25,7 +25,7 @@ from the appropriate page on our website, and check out the
 
 ## What's in 1.50.0 stable
 
-For this release, we have improved array indexing, expanded union field safety, refined file descriptors, and added to the standard library.
+For this release, we have improved array indexing, expanded safe access to union fields, and added to the standard library.
 See the [detailed release notes][notes] to learn about other changes
 not covered by this post.
 
@@ -67,7 +67,7 @@ Some types in Rust have specific limitations on what is considered a
 valid value, which may not cover the entire range of possible memory
 values. We call any remaining invalid value a [niche], and this space
 may be used for type layout optimizations. For example, in Rust 1.28
-we introduced `NonZero` types where `0` is a niche, and this allowed
+we introduced `NonZero` integer types (like `NonZeroU8`) where `0` is a niche, and this allowed
 `Option<NonZero>` to use `0` to represent `None` with no extra memory.
 
 On Unix platforms, Rust's `File` is simply made of the system's integer
--- a/posts/2021-02-11-Rust-1.50.0.md
+++ b/posts/2021-02-11-Rust-1.50.0.md
@@ -1,0 +1,140 @@
+---
+layout: post
+title: "Announcing Rust 1.50.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.50.0. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.50.0 is as easy as:
+
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install]
+from the appropriate page on our website, and check out the
+[detailed release notes for 1.50.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1500-2021-02-11
+
+## What's in 1.50.0 stable
+
+For this release, we have improved array indexing, expanded union field safety, refined file descriptors, and new library additions.
+See the [detailed release notes][notes] to learn about other changes
+not covered by this post.
+
+### Const-generic array indexing
+
+Continuing the march toward stable `const` generics, this release adds
+implementations of `ops::Index` and `IndexMut` for arrays `[T; N]` for
+_any_ length of `const N`. The indexing operator `[]` already worked on
+arrays through built-in compiler magic, but at the type level, arrays
+didn't actually implement the library traits until now.
+
+```rust
+fn second<C>(container: &C) -> &C::Output
+where
+    C: std::ops::Index<usize> + ?Sized,
+{
+    &container[1]
+}
+
+fn main() {
+    let array: [i32; 3] = [1, 2, 3];
+    assert_eq!(second(&array[..]), &2); // slices worked before
+    assert_eq!(second(&array), &2); // now it also works directly
+}
+```
+
+### Safe assignments to `ManuallyDrop<T>` union fields
+
+Rust 1.49 made it possible to add `ManuallyDrop<T>` fields to a `union`
+as part of allowing `Drop` for unions at all; however, unions don't know
+which variant is valid (they don't drop any old value on assignment), so
+safe Rust has limited this to `Copy` types only, which never `Drop`. Of
+course, that's also true of `ManuallyDrop`, so now Rust 1.50 allows safe
+assignments to these fields as well.
+
+### A niche for `File` on Unix platforms
+
+Some types in Rust have specific limitations on what is considered
+valid value, which may not cover the entire range of possible memory
+values. We call any remaining invalid value a [niche], and this space
+may be used for type layout optimizations. For example, in Rust 1.28
+we introduced `NonZero` types where `0` is a niche, and this allowed
+`Option<NonZero>` to use `0` to represent `None` with no extra memory.
+
+On Unix platforms, Rust's `File` is simply made of the system's integer
+file descriptor, and this happens to have a possible niche
+as well because it can never be `-1`! System calls which return a file
+descriptor use `-1` to indicate that an error occurred (check `errno`)
+so it's never possible for `-1` to be a real file descriptor. Starting
+in Rust 1.50 this niche is represented to the compiler so it can be
+use in layout optimizations too. It follows that `Option<File>` will
+now have the same size as `File` itself!
+
+[niche]: https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#niche
+
+### Library changes
+
+In Rust 1.50.0, there are nine new stable functions:
+
+- [`bool::then`]
+- [`btree_map::Entry::or_insert_with_key`]
+- [`f32::clamp`]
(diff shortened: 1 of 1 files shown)


CryZe:
What about `You can now use const values for x in [x; N] array expressions.`? Seems like a pretty big deal considering it has always been kinda difficult to initialize arrays.

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+may be used for type layout optimizations. For example, in Rust 1.28
+we introduced `NonZero` types where `0` is a niche, and this allowed
+`Option<NonZero>` to use `0` to represent `None` with no extra memory.
+
+On Unix platforms, Rust's `File` is simply made of the system's integer
+file descriptor, and this happens to have a possible niche
+as well because it can never be `-1`! System calls which return a file
+descriptor use `-1` to indicate that an error occurred (check `errno`)
+so it's never possible for `-1` to be a real file descriptor. Starting
+in Rust 1.50 this niche is represented to the compiler so it can be
```
camelid, line 78: "represented" sounds a bit funny to me. However, I can't think of a better word to use ;)

Does someone have an idea of a better word? If not, it's probably fine as-is though.

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+Rust 1.49 made it possible to add `ManuallyDrop<T>` fields to a `union`
+as part of allowing `Drop` for unions at all; however, unions don't know
+which variant is valid (they don't drop any old value on assignment), so
+safe Rust has limited this to `Copy` types only, which never `Drop`. Of
+course, that's also true of `ManuallyDrop`, so now Rust 1.50 allows safe
+assignments to these fields as well.
+
+### A niche for `File` on Unix platforms
+
+Some types in Rust have specific limitations on what is considered
```
camelid, line 66: You're missing "a":

```suggestion
Some types in Rust have specific limitations on what is considered a
```

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+If you don't have it already, you can [get `rustup`][install]
+from the appropriate page on our website, and check out the
+[detailed release notes for 1.50.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1500-2021-02-11
+
+## What's in 1.50.0 stable
+
+For this release, we have improved array indexing, expanded union field safety, refined file descriptors, and new library additions.
```
lzutao, line 28:
```suggestion
For this release, we have improved array indexing, expanded union field safety, refined file descriptors, and ? new library additions.
```

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+we introduced `NonZero` types where `0` is a niche, and this allowed
+`Option<NonZero>` to use `0` to represent `None` with no extra memory.
+
+On Unix platforms, Rust's `File` is simply made of the system's integer
+file descriptor, and this happens to have a possible niche
+as well because it can never be `-1`! System calls which return a file
+descriptor use `-1` to indicate that an error occurred (check `errno`)
+so it's never possible for `-1` to be a real file descriptor. Starting
+in Rust 1.50 this niche is represented to the compiler so it can be
+use in layout optimizations too. It follows that `Option<File>` will
```
lzutao, line 79:
```suggestion
used in layout optimizations too. It follows that `Option<File>` will
```

posts/2021-02-11-Rust-1.50.0.md
camelid: Maybe?

```suggestion
For this release, we have improved array indexing, expanded union field safety, refined file descriptors, and added to the standard library.
```

posts/2021-02-11-Rust-1.50.0.md
programmerjake: how about "this niche is added to Rust's type system so it can be used in layout optimizations too"

camelid: IIUC, it's not that it's added to the type system – that implies it was hard-coded. It's more like it's *exposed* to the type system, so maybe this?

```suggestion
in Rust 1.50 information about this niche is exposed to the compiler so it can be
```

programmerjake: how about "this niche is added to the type's definition so it can be used in layout optimizations too"

programmerjake: since the niche didn't exist before

posts/2021-02-11-Rust-1.50.0.md
cuviper: I guess I was thinking "we have ... new library additions", but that doesn't agree with verb tense of the others. I like @camelid's suggestion.

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+If you don't have it already, you can [get `rustup`][install]
+from the appropriate page on our website, and check out the
+[detailed release notes for 1.50.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1500-2021-02-11
+
+## What's in 1.50.0 stable
+
+For this release, we have improved array indexing, expanded union field safety, refined file descriptors, and added to the standard library.
```
Mark-Simulacrum, line 28:
```suggestion
For this release, we have improved array indexing, expanded safe access to union fields, and added to the standard library.
```

* Not sure why, but "expanded union field safety" was not quite making sense to me - maybe this looks better?
* I don't thinking refining file descriptors parses for me, seems easier to cut it.

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+### Safe assignments to `ManuallyDrop<T>` union fields
+
+Rust 1.49 made it possible to add `ManuallyDrop<T>` fields to a `union`
+as part of allowing `Drop` for unions at all; however, unions don't know
+which variant is valid (they don't drop any old value on assignment), so
+safe Rust has limited this to `Copy` types only, which never `Drop`. Of
+course, that's also true of `ManuallyDrop`, so now Rust 1.50 allows safe
+assignments to these fields as well.
```
Mark-Simulacrum, line 62: Can we split the first sentence of this paragraph up? I think it's carrying a lot of weight right now in a relatively short space and would be nice to add some breaks for readers.

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+
+### A niche for `File` on Unix platforms
+
+Some types in Rust have specific limitations on what is considered a
+valid value, which may not cover the entire range of possible memory
```
Mark-Simulacrum, line 67: Maybe worth linking to https://www.ralfj.de/blog/2019/07/14/uninit.html as well

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+values. We call any remaining invalid value a [niche], and this space
+may be used for type layout optimizations. For example, in Rust 1.28
+we introduced `NonZero` types where `0` is a niche, and this allowed
```
Mark-Simulacrum, line 70:
```suggestion
we introduced `NonZero` integer types (like `NonZeroU8`) where `0` is a niche, and this allowed
```

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,140 @@
...
+In Rust 1.50.0, there are nine new stable functions:
+
+- [`bool::then`]
+- [`btree_map::Entry::or_insert_with_key`]
+- [`f32::clamp`]
+- [`f64::clamp`]
+- [`hash_map::Entry::or_insert_with_key`]
+- [`Ord::clamp`]
+- [`RefCell::take`]
+- [`slice::fill`]
```
Mark-Simulacrum, line 95: Hm, I am now wondering if this would be a better thing to call out than the File niche - it seems more likely to be notable to most users and is pretty nice as "safe Rust API" where the C ones is pretty annoying to use in a safe way.

posts/2021-02-11-Rust-1.50.0.md
cuviper: I think that's off-topic? Uninitialized memory is sort of an abstract fuzzy state, outside of merely "bit patterns that are invalid for this type" that we use for niches.

Mark-Simulacrum: Yeah - I just see "entire range of memory values" and think of uninit immediately :)

posts/2021-02-11-Rust-1.50.0.md
cuviper: I split and rearranged this a bit.

posts/2021-02-11-Rust-1.50.0.md
cuviper: My muse fails me on this one. I agree `slice::fill` is a useful method, but even the documentation has only a terse sentence and short example. If you're feeling inspired, would you (or anyone else) care to try writing something for this?

jyn514: > is pretty nice as "safe Rust API" where the C ones is pretty annoying to use in a safe way.

This is the same as `memset`, right? Why is this annoying in C? Memset takes the length of the buffer.

Mark-Simulacrum: memset is also used to e.g. zero out arrays of structs, where it's finicky (IME) to get the length and such right in a guaranteed way. this API avoids those problems. but I don't feel strongly.

jyn514: Ah right, because people tend to put `sizeof(T*)` instead of `sizeof(T)`. Ok, makes sense.

XAMPPRocky: > Hm, I am now wondering if this would be a better thing to call out than the File niche

FWIW, I don't think it needs to be an either or situation, the blog post as it stands is about 600 words, adding one more paragraph isn't going to significantly increase the length. It takes roughly 3 minutes to read. I don't think I've ever seen someone complain about the blog post being too long, but I have seen people lament it being too short.

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,170 @@
...
+    assert_eq!(second(&array[..]), &2); // slices worked before
+    assert_eq!(second(&array), &2); // now it also works directly
+}
+```
+
+### `const` value repetition for arrays
+
+Arrays in Rust can be written either as a list `[a, b, c]` or a repetition `[x; N]`.
+For lengths `N` greater than one, repetition has only been allowed for `Copy`
+types, and [RFC 2203] sought to allow any `const` expression there. However,
```
camelid, line 59: This is a little unclear about *what* needs to be `Copy`: is it `x` or `N`?

```suggestion
For lengths `N` greater than one, repetition has only been allowed for `x`s that are `Copy`,
and [RFC 2203] sought to allow any `const` expression there. However,
```

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,170 @@
...
+    const EMPTY: Option<Vec<i32>> = Some(Vec::new());
+
+    // However, repeating a `const` value is allowed!
+    let nones = [NONE; 10];
+    let empties = [EMPTY; 10];
+}
+```
+
+In Rust 1.50, that stabilization is formally acknowledged, so you may use this
+construct without fear. In the future, to avoid such "temporary" named
```
Mark-Simulacrum, line 79: Hm, I'd prefer to avoid this "without fear" language - I think it's unlikely we would have dropped support for such a long-term stabilized feature regardless, and it indicates some worry about backwards compatibility in my eyes.

Mark-Simulacrum (commented):
Looking good modulo one small nit - I think maybe just cutting the sentence early is best.

posts/2021-02-11-Rust-1.50.0.md
Mark-Simulacrum, line 79:
```suggestion
In Rust 1.50, that stabilization is formally acknowledged. In the future, to avoid such "temporary" named
```

posts/2021-02-11-Rust-1.50.0.md
```diff
@@ -0,0 +1,169 @@
...
+
+[RFC 2203]: https://rust-lang.github.io/rfcs/2203-const-repeat-expr.html
+[RFC 2920]: https://rust-lang.github.io/rfcs/2920-inline-const.html
+
+### Safe assignments to `ManuallyDrop<T>` union fields
+
+Rust 1.49 made it possible to add `ManuallyDrop<T>` fields to a `union` as part
+of allowing `Drop` for unions at all. However, unions don't drop old values
+when a field is assigned, since they don't know which variant was formerly
+valid, so safe Rust has limited this to `Copy` types only, which never `Drop`.
```
Mark-Simulacrum, line 89:
```suggestion
valid, so safe Rust previously limited this to `Copy` types only, which never `Drop`.
```

PR #786, opened 2021-02-27
Style: Add left margin for columns on all screen sizes

rbuch:
Previously, `.column` and `.columns` only had a `margin-left` of `4%` when the viewport satisfied `min-width:550px`. On mobile, this led to the columns in the footer abutting each other with no space in between (see "Before" screenshot, in particular the "Contact the Rust Team" and "Licenses" lines). This change gives `.column` and `.columns` a left margin of `4%` at all screen widths, giving the columns a bit of breathing room. Before and after screenshots are given below, as rendered by Firefox's Responsive Design Mode using the Pixel 2 template.

Before:
before

After:
after

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -163,6 +163,11 @@ ul.nav, ul.nav li {
   margin-bottom: 0;
 }
 
+// TODO:  remove when switching to Tachyons -- these are overrides for Skeleton.
+.column, .columns {
+  margin-left: 4%;
+}
+
 .nav a {
   color: var(--gray);
   color: var(--nav-links-a);


camelid:
It seems we're not supposed to modify `skeleton.css` directly—see https://github.com/rust-lang/blog.rust-lang.org/pull/683#issuecomment-689624894.

rbuch:
> It seems we're not supposed to modify `skeleton.css` directly—see [#683 (comment)](https://github.com/rust-lang/blog.rust-lang.org/pull/683#issuecomment-689624894).

Ah, I see, thanks for pointing that out. It doesn't seem like your questions from that PR regarding where to put new CSS were ever answered, but do you know where new style changes should go? It seem like `src/styles/app.scss` has some overrides for Skeleton already, perhaps that's the place.

rbuch:
I've moved my change over to `src/styles/app.scss` rather than modifying `skeleton.css` directly, hopefully this is a better place for it.

camelid:
> It doesn't seem like your questions from that PR regarding where to put new CSS were ever answered, but do you know where new style changes should go?

Yeah, I don't know unfortunately.

> I've moved my change over to `src/styles/app.scss` rather than modifying `skeleton.css` directly, hopefully this is a better place for it.

If that works out for you, I might try that in my own PR. Thanks :)

rylev:
@rbuch sorry for letting this sit for so long. Is this just waiting for approval?

rbuch:
> @rbuch sorry for letting this sit for so long. Is this just waiting for approval?

No worries. Yeah, it should be good to go, just needs approval.

corneliusroemer:
I opened an issue for this not knowing there was a PR already! I think this would be good to get back on track :)

senekor (approved):

PR #799, opened 2021-03-16
Add new error handling blog post

yaahc:
* [Rendered](https://github.com/yaahc/blog.rust-lang.org/blob/master/posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md)

### DIFF
--- a/posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
+++ b/posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
@@ -178,7 +178,7 @@ Assuming this works as well as we hope it does, and doesn't introduce worse prob
 
 ### Disclaimers
 
-Note, we almost certainly won't _start_ with adding these types to std, and will instead trial this approach in 3rd party libraries, probably in a fork of `eyre` or `anyhow`. We're particularly worried about backwards compatilibity so its easy to switch to the new libraries that properly implement `Error`. We're also worried about teachability, particularly with respect to knowing which `Result` type to use in each situation, and we're looking into what kind of errors get produced when you pick the wrong Error or Result type, and how we can guide users to the correct types.
+Note, we almost certainly won't _start_ with adding these types to std, and will instead trial this approach in 3rd party libraries, probably in a fork of `eyre` or `anyhow`. We're particularly worried about backwards compatibility so its easy to switch to the new libraries that properly implement `Error`. We're also worried about teachability, particularly with respect to knowing which `Result` type to use in each situation, and we're looking into what kind of errors get produced when you pick the wrong Error or Result type, and how we can guide users to the correct types.
 
 Also worth noting, this doesn't change how `std::result::Result` works. If you're happy with enum based error handling and `From` impls as necessary then you can still stick with the current `Result` type and `?` as you currently do. These changes only apply to types like `Box<dyn Error>` that implement From for all error types so they can still implement error while using `?` for convenience.
 
--- a/posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
+++ b/posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
@@ -1,0 +1,199 @@
+---
+layout: post
+title: "What the Error Handling Project Group is Working Towards"
+author: Jane Lusby
+team: the library team <https://www.rust-lang.org/governance/teams/library>
+---
+
+This blog post is a follow up of our [previous](https://blog.rust-lang.org/inside-rust/2020/11/23/What-the-error-handling-project-group-is-working-on.html) post detailing what we're working on now. We figure it's time to take a step back and answer some of the bigger questions of why we're making these changes and how they're going to fit into the bigger picture of error-handling provided by Rust.
+
+> Disclaimer: *This post is equal parts plan and aspiration. There are technical challenges here to sort out so the final outcome may look rather different from our initial vision, so please don't assume any of this is final.*
+
+## Guidelines for implementing `Display::fmt` and `Error::source`
+
+But before we get into the plans we first need to share an announcement. Project Error Handling recently created a guideline for [how to implement `Display::fmt` and `Error::source`](https://github.com/rust-lang/project-error-handling/issues/27#issuecomment-763950178). To create this guideline we first had to answer a fundamental question, _What is the primary role of the error trait?_.
+
+As we saw it there are two possible answers.
+
+* The error trait is primarily an interface for reporting errors.
+* The error trait is primarily an interface for reacting to specific errors.
+
+In the first case, "an interface for reporting errors" means it provides access to pieces of information about the error such as the error message, optional backtrace, optional source, and so on. A reporter then takes all this information and stitches it together into a full error report, explaining what went wrong.
+
+If we decide to prefer the former viewpoint then we would want to prioritize not duplicating information in the reporting interfaces of the `Error` trait. That would then turn into the guidance: _Return source errors via `Error::source` unless the source error's message is included in your own error message in `Display`._
+
+In the second case, "an interface for reacting to specific errors" is referring to the various `downcast` methods associated with the error trait. These methods can be used to test if a type erased error is a specific error variant and they're functionally equivalent to `match`.
+
+If we decide to prefer the latter then we would want to prioritize exposing every error type in the chain of errors to ensure users can always correctly downcast to specific error variants they want to handle. In this situation the guidance would be: _Always return source errors via `Error::source`._
+
+After some discussion we concluded that the reporting focused viewpoint was the ideal one. We prefer splitting our errors into multiple error messages with no duplication of information as opposed to having each error print itself followed by all others. This split moves the responsibility of formatting to the end user, rather than relying on many library authors concatenating error messages together with a consistent format.
+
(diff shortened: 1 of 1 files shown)


posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,201 @@
...
+
+
+### Disclaimers
+
+Note, we almost certainly won't _start_ with adding these types to std, and will instead trial this approach in 3rd party libraries, probably in a fork of `eyre` or `anyhow`. We're particularly worried about backwards compatilibity so its easy to switch to the new libraries that properly implement `Error`. We're also worried about teachability, particularly with respect to knowing which `Result` type to use in each situation, and we're looking into what kind of errors get produced when you pick the wrong Error or Result type, and how we can guide users to the correct types.
```
kw-fn, line 181:
```suggestion
Note, we almost certainly won't _start_ with adding these types to std, and will instead trial this approach in 3rd party libraries, probably in a fork of `eyre` or `anyhow`. We're particularly worried about backwards compatibility so its easy to switch to the new libraries that properly implement `Error`. We're also worried about teachability, particularly with respect to knowing which `Result` type to use in each situation, and we're looking into what kind of errors get produced when you pick the wrong Error or Result type, and how we can guide users to the correct types.
```

posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,201 @@
...
+   1: source's error message
+   2: source's source's ...
+
+Backtrace:
+   ...
+```
+
+This should provide reasonable defaults for simple error reporting cases that are encountered most of the time. Applications that need a more complex error report format can then provide their own reporter by wrapping errors in a `Display` type the same way the `Path::display` method does for paths.
+
+## Panics cannot be made from `dyn Error`s, and instead stringifies the error
```
flip1995, line 63: The first heading starts with `Problem #1:`, but then no `Problem #2` heading follows in the headings, so I would either remove it from the first heading or add it to the following.

flip1995 (commented):
Personally, I prefer to avoid the word "Problem" and rather call it "Challenge" :) Great write-up either way though.

posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,201 @@
+---
+layout: post
+title: "What the Error Handling Project Group is Working Towards"
+author: Jane Lusby
+team: the library team <https://www.rust-lang.org/governance/teams/library>
```
m-ou-se, line 5: Shouldn't this be the error handling group?
```suggestion
team: The Error Handling Project Group <https://www.rust-lang.org/governance/teams/library#project-error-handling>
```

posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,201 @@
...
+        t
+    }
+}
+```
+
+These `From` impls interfere with our `Error` impl due to the "Overlap Rule", which states that no two generic implementations of a Trait can apply to the same type. If `Box<dyn Error>` implements the `Error` trait both of the `From` impls above would apply, and `BoxError::from(other_box_error)` becomes ambiguous. The compiler doesn't know whether it should pass the Box through directly or wrap it in another `Box`.
+
+In theory we could fix this in the future if we ever get specialization finished, but as far as we know its highly unlikely that it would be useful in this specific case. We'd need to support [lattice specialization](https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md#the-lattice-rule) which is, by our most optimistic understanding, very far from happening if ever.
+
+### Hope, AKA `Try` trait v2 RFC
```
bstrie, line 150: Now that this RFC is in its final revision, just wanted to confirm that everything in this document is still true (bikeshedded names notwithstanding).

posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,201 @@
...
+
+Note, we almost certainly won't _start_ with adding these types to std, and will instead trial this approach in 3rd party libraries, probably in a fork of `eyre` or `anyhow`. We're particularly worried about backwards compatibility so its easy to switch to the new libraries that properly implement `Error`. We're also worried about teachability, particularly with respect to knowing which `Result` type to use in each situation, and we're looking into what kind of errors get produced when you pick the wrong Error or Result type, and how we can guide users to the correct types.
+
+Also worth noting, this doesn't change how `std::result::Result` works. If you're happy with enum based error handling and `From` impls as necessary then you can still stick with the current `Result` type and `?` as you currently do. These changes only apply to types like `Box<dyn Error>` that implement From for all error types so they can still implement error while using `?` for convenience.
+
+## Summary
+
+To summarize on the above our plan is to:
+
+1. Move the error trait into core
```
Amanieu, line 189: This comes a bit out of nowhere considering this is the first item on the plan. There was no other mention of core in the rest of the blog post.

posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,201 @@
...
+---
+
+This blog post is a follow up of our [previous](https://blog.rust-lang.org/inside-rust/2020/11/23/What-the-error-handling-project-group-is-working-on.html) post detailing what we're working on now. We figure it's time to take a step back and answer some of the bigger questions of why we're making these changes and how they're going to fit into the bigger picture of error-handling provided by Rust.
+
+> Disclaimer: *This post is equal parts plan and aspiration. There are technical challenges here to sort out so the final outcome may look rather different from our initial vision, so please don't assume any of this is final.*
+
+## Guidelines for implementing `Display::fmt` and `Error::source`
+
+But before we get into the plans we first need to share an announcement. Project Error Handling recently created a guideline for [how to implement `Display::fmt` and `Error::source`](https://github.com/rust-lang/project-error-handling/issues/27#issuecomment-763950178). To create this guideline we first had to answer a fundamental question, _What is the primary role of the error trait?_.
```
therealprof, line 14: That renders funny with the period following the questionmark.

posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,201 @@
...
+
+In the second case, "an interface for reacting to specific errors" is referring to the various `downcast` methods associated with the error trait. These methods can be used to test if a type erased error is a specific error variant and they're functionally equivalent to `match`.
+
+If we decide to prefer the latter then we would want to prioritize exposing every error type in the chain of errors to ensure users can always correctly downcast to specific error variants they want to handle. In this situation the guidance would be: _Always return source errors via `Error::source`._
+
+After some discussion we concluded that the reporting focused viewpoint was the ideal one. We prefer splitting our errors into multiple error messages with no duplication of information as opposed to having each error print itself followed by all others. This split moves the responsibility of formatting to the end user, rather than relying on many library authors concatenating error messages together with a consistent format.
```
therealprof, line 29: Why is this is considered the "ideal" approach? It feels as if the mission of the "The Error Handling Project Group" was to prioritise the presentation of errors rather than the handling of errors (as indicated in the project group name). For a systems programming language I feel that the handling of errors should have priority over the presentation of errors; there're systems which do not even have the ability to present error messages to a user (e.g. embedded devices) but where error handling is of utmost importance (and rather poorly supported by Rust at the moment).

nikomatsakis (commented):
Hmm, @yaahc, I'd love to discuss this post a bit over voice sometime. To be quite honest, reading it, I had a hard time following it. I'd like to get a better idea of who the audience is and what the top-level goals are that you're trying to communicate. It might be that it wants to be structured into a few separate posts, even.

posts/inside-rust/2021-03-15-What-the-error-handling-project-group-is-working-towards.md
yaahc: > Why is this considered the "ideal" approach?

Perhaps ideal isn't the best vocab here, since there isn't really a right answer, just a best one. The reasoning is that we evaluated the trade-offs and costs of the various options and found that this was the least bad option, given that it doesn't prevent reacting to source errors via downcasting, where-as the alternative completely prevents configurable and consistent error reporting formats by duplicating information throughout the `Error` interface. If you're interested in a more in-depth analysis of the trade offs and why we felt it was best to prioritize deduplicating information please checkout this issue: https://github.com/rust-lang/project-error-handling/issues/27

> It feels as if the mission of the "The Error Handling Project Group" was to prioritise the presentation of errors rather than the handling of errors (as indicated in the project group name). For a systems programming language I feel that the handling of errors should have priority over the presentation of errors; there're systems which do not even have the ability to present error messages to a user (e.g. embedded devices) but where error handling is of utmost importance (and rather poorly supported by Rust at the moment).

"Error Handling" a superset of responsibilities, which in my mind includes creating, propagating, matching/reacting, discarding, and reporting. This is more accurately a tension between how you can programmatically react to specific errors and how you represent and report them, as opposed to handling in general. I think its well within the mission of the error handling project group to analyze the trade offs and make a recommendation one direction or the other.

yaahc:
I'm closing this PR for now because I'm working on rewriting this blog post based on the feedback I received from @nikomatsakis. I'll do my best to include all the feedback given in the updated version of the blog post where applicable.

PR #801, opened 2021-03-17
Add tmandry director intro post

tmandry:
I'm not exactly sure what the title should be. I also added a small bit of text introducing the series. @ashleygwilliams feel free to change these completely.

Today I make some very tiny wording tweaks, and added a sub-header to break up the paragraphs a bit.


tmandry:
Wrong repo :)

PR #836, opened 2021-05-07
1.52.1 - Blog post explaining the fingerprint issue.

pnkfelix:


Mark-Simulacrum (dismissed):
I will likely aim to merge this in roughly an hour, but it looks good to me.

Mark-Simulacrum:
Moving publish date to Monday morning.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
...
+thread 'rustc' panicked at 'found unstable fingerprints for predicates_of(<massive text describing rustc internals elided>)', /rustc/88f19c6dab716c6281af7602e30f413e809c5974/compiler/rustc_query_system/src/query/plumbing.rs:593:5
+```
+
+They all arise from inconsistencies when comparing the incremental-compilation cache stored on disk against the values computed during a current `rustc` invocation, which means they all arise from using incremental compilation.
+
+There are three ways that you may have incremental compilation turned on: You may have set the [environment variable][env-vars] `CARGO_INCREMENTAL=1`, or you may have enabled the `build.incremental` [setting in your Cargo.toml][cargo-toml], or you may be building with the `dev` or `test` [profiles][], which default to having incremental compilation enabled.
+
+[env-vars]: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads
+[cargo-toml]: https://doc.rust-lang.org/cargo/reference/config.html#buildincremental
+[profiles]: (https://doc.rust-lang.org/cargo/reference/profiles.html
```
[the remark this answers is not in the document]
pnkfelix: omg thank you!

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
...
+## What should a Rust programmer do in response
+
+The Internal Compiler Error asks you to report a bug, and if you can do so, we still want that information. We *want* to know about the cases that are failing.
+
+But regardless of whether or not you file a bug, the problem here can be resolved by either:
+
+ 1. deleting your incremental compilation cache (e.g. by running `cargo clean`), or
+ 2. force incremental compilation to be disabled, by setting `CARGO_INCREMENTAL=0` in your environment or `build.incremental` to `false` in the `config.toml`.
+
+We recommend that users of 1.52.0 disable incremental compilation, to avoid running into this problem.
```
pnkfelix, line 101: Is this language too strong? I'm reflecting on the feedback from @carols10cents [here](https://github.com/rust-lang/rust/issues/84970#issuecomment-834771043).

Should we be using the softer language of "We recommend users of 1.52.0 disable incremental compilation *if* they are encountering unstable fingerprints repeatedly" ? We still don't have a good understanding of how often people are running into this, and if they run into it repeatedly even after their first `cargo clean`...

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
...
+
+We do *not* recommend that users of 1.52.0 downgrade to an earlier version of Rust in response to this problem. As noted above, there is at least one instance of a silent [miscompilation][issue-82920] caused by incremental compilation that was not caught until we added the fingerprint checking.
+
+## What is the Rust project going to do to fix this
```
carols10cents, line 105: I'm not sure if yinz have had time to do this analysis yet, but I'd love to see some reflection in this section on "how did this get missed in nightly/beta and how would we prevent similar mistakes in the future", or an indication that this analysis will be done and published after the fixes are out.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
+---
+layout: post
+title: "1.52.0, fingerprints and compiler error forensics"
+author: Felix Klock
+team: the compiler team <https://www.rust-lang.org/governance/teams/compiler>
+---
+
+The Rust teams are always excited to report on new features offered with each release. Sometimes, however, an important change that is not yet "fully baked" gets accidentally included in a release.
```
carols10cents, line 8: I think it would be really helpful to have a Summary/TL;DR section at the top that says something along the lines of:

- There are bugs in incremental compilation in 1.52.0 stable. 
- Incremental compilation is off by default for release builds, so no production builds should be affected. 
- If you are encountering these bugs during development, you can disable incremental compilation by using the environment variable `CARGO_INCREMENTAL=0` or editing your `$HOME/.cargo/config.toml` or `$HOME/.cargo/config.toml` file to add `build.incremental = 'false'`.
- We're expecting to have a 1.52.1 release on [whatever date]
- We [do|do not] want more bug reports (and if you do want them, what's the most useful information to include)

carols10cents (changes requested):
Haven't finished reading yet but wanted to submit comments so far

posts/2021-05-07-caught-red-handed.md
est31: > There are bugs in incremental compilation in 1.52.0 stable.

*and prior versions*

posts/2021-05-07-caught-red-handed.md
Mark-Simulacrum: I'll provide some background and maybe you can help indicate which of this is useful/helpful information - it may be published in a retrospective later, but I don't know yet if that'll happen.

On the day of the release I submitted https://github.com/rust-lang/rust/issues/84970 to resolve the question of what to do on 1.53 and 1.54 releases. The verification change landed in 1.52, but I had mistakenly assumed that it had landed in 1.53 and we had an additional 6 weeks to either resolve it or improve our messaging; so, ultimately, the fact that this ICE existed *was* known for 12 weeks now, and indeed a few people have been driving the fixing. I think it was not as well understood how common it was - to my knowledge, we had not received reports until after the stable release of it occurring on *every* build, for example. So it was a priority, but not too strongly so.

Another part of the reason for the lack of prioritization is that this is a longstanding bug - to my knowledge, perhaps even to the initial addition of incremental (at least the red/green system, added a few years back), I don't think I'm aware of work since then that would have caused this. So, while it is true that the ICEs are definitely annoying, they merely reveal underlying problems - and in some sense, are better than nothing. I think everyone is in agreement that they are not the best user experience, and that we should avoid them if at all possible, but I think no one has come up with a straightforward solution to do so. It's possible that an alternative to deploying them would be to disable incremental as a default-on option, and indeed that is under active consideration for next week's point release to mitigate the problem.

In hindsight, it's possible that when we discovered this bug we should have immediately moved to do so. However, disabling such a milestone feature that *does* contribute significant performance improvements to many builds would be a hard sell, especially in light of some builds not triggering the verification panics - and as such being able to compile smoothly, even with the added verification. The tradeoff here is not clear.

I will also share that the release team discussed this last Friday and is expecting to invest in better tooling for tracking bugs (regressions in particular, but more generally too) as they propagate to stable, so that we can have a better awareness and hopefully avoid mistakes of the "woops, I thought this would be in 1.53" variety. This should help with a number of other challenges we've encountered as well. However, I expect this to be a long-term project, not one we can quickly address.

posts/2021-05-07-caught-red-handed.md
Mark-Simulacrum: > There are bugs in incremental compilation in 1.52.0 stable.

Note that there are bugs in incremental compilation ... in essentially all known versions, in 1.52 they're just very visible. I think we should be very clear that downgrading to 1.51 is likely *not* the path users want to take.

posts/2021-05-07-caught-red-handed.md
carols10cents: If it would be helpful for folks to provide you with more information about what they're seeing, this would be a great place to give more details on the most helpful information and ways to provide that. I think some folks would be happy to put up with some inconvenience if they can help to identify and fix the problems sooner.

Some folks will definitely just want to get on with their work though, and it sucks that the way to do that will make compile times slower, so I think it's important to acknowledge that. Something like:

```suggestion
To completely avoid running into this problem when using 1.52.0, we recommend disabling incremental compilation via one of the methods described above until we have had a chance to identify and fix the problems and issue a new release. We know this will make compile times slower and we know that's not an ideal workaround, but we are working to make this as temporary a situation as possible.
```

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
...
+The Internal Compiler Error asks you to report a bug, and if you can do so, we still want that information. We *want* to know about the cases that are failing.
+
+But regardless of whether or not you file a bug, the problem here can be resolved by either:
+
+ 1. deleting your incremental compilation cache (e.g. by running `cargo clean`), or
+ 2. force incremental compilation to be disabled, by setting `CARGO_INCREMENTAL=0` in your environment or `build.incremental` to `false` in the `config.toml`.
+
+We recommend that users of 1.52.0 disable incremental compilation, to avoid running into this problem.
+
+We do *not* recommend that users of 1.52.0 downgrade to an earlier version of Rust in response to this problem. As noted above, there is at least one instance of a silent [miscompilation][issue-82920] caused by incremental compilation that was not caught until we added the fingerprint checking.
```
carols10cents, line 103: And we don't know how far back these bugs are present.

posts/2021-05-07-caught-red-handed.md
est31: > "how did this get missed in nightly/beta and how would we prevent similar mistakes in the future"

There has been bug filing going on basically ever since the ICEs were turned on.

If you check the [list of issues](https://github.com/rust-lang/rust/issues?q=is%3Aissue+is%3Aopen+unstable+fingerprints) linked by @Mark-Simulacrum 's issue, you'll see the [oldest](https://github.com/rust-lang/rust/issues/83115) [issues](https://github.com/rust-lang/rust/issues/83126) having been filed on March 14 to the nightly compiler, with the [PR](https://github.com/rust-lang/rust/pull/83007) that enabled the checking having been merged in March 13, so right after that PR merged. With increased exposure to users, the number of reports increased.

But the bugs existed previously too, just caused silent corruptions and weird bugs instead, which is arguably worse than an explicit error message. To the contrary, turning on this error is "making sure the errors won't be missed any more".

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
...
+
+## What is the Rust project going to do to fix this
+
+### Short-term plan
+
+Based on the number of bug reports that have already come in, we know we need to do something quickly.
+
+We are going to be issuing a point release, 1.52.1. The point release will improve the diagnostic output from `verify-ich`, so that a programmer can more effectively act in response to the message (i.e., they will be told that the problem is due to the use of incremental compilation, and that they need to delete their cache, at the very least).
```
carols10cents, line 111: Committing to a date for the point release here, if that's at all possible, would go a long way. Even if you know it won't fix any problems, or will only fix a small number of problems, saying "we'll have a release with whatever fixes we have ready out" for people to try, and at that time saying whether we think there's going to be another point release or not, would make people feel like Something is Being Done.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Long-term plan
+
+The long-term plan is to fix the bugs! Incremental compilation is the only realistic way for the Rust compiler to be able to provide a fast edit-compile-run cycle for all of its programmers, and so we need to address [all of the issues][issue-list] that have been identified thus far via `verify-ich`. (There are 32 such issues as of this writing, though many may be duplicates.)
+
+[issue-list]: https://github.com/rust-lang/rust/issues?q=is%3Aissue+is%3Aopen+unstable+fingerprints
+
+If you want to come help us do it, we would love for you to [join us][] with the effort!
+
+[join us]: https://www.rust-lang.org/community
```
carols10cents, line 125: I think a more specific place for people to go who want to help with this specific problem would be better than rust-lang.org/community-- would the T-compiler stream on zulip be appropriate? WG-incr-comp? Maybe have a transient stream just for this issue so that the more general streams can be about other topics?

carols10cents (changes requested):
Ok, thoughts on the post done. I'm happy to read revisions later today or tomorrow if that would be helpful!

posts/2021-05-07-caught-red-handed.md
carols10cents: Is "There are bugs in incremental compilation in all Rust stable versions that were made into compilation errors in 1.52.0 stable." more accurate? I just edited my post to that

posts/2021-05-07-caught-red-handed.md
carols10cents: Ok, I think the start of a TL;DR on root causes sounds to me like:

- We discovered a silent bug(s) with incremental compilation that have probably existed since incremental compilation was added
- We made the bug into a loud compiler error rather than a silent bug
- While that change was in nightly and beta, we got some bug reports but didn't realize exactly how often this bug would be hit in normal development with stable (most people/projects probably only run night/beta in CI where incremental compilation is either irrelevant or disabled)
- We thought this change was only going into beta this cycle; longer term, we will be looking into tooling improvements to provide more clarity of what version a change is in and possibly how many bugs are being reported about a change as a change switches trains
- It also sounds like we should do some thinking on how to get more user testing of parts of Rust only used during development while they're on nightly and beta?

Please feel free to edit this if I've misunderstood or if you disagree with my conclusions or save this stuff for a retro, but as a Rust user reading this, I mostly want to know that the team is going to take a bigger view of this as well as fixing the immediate issue.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
+---
+layout: post
+title: "1.52.0, fingerprints and compiler error forensics"
+author: Felix Klock
+team: the compiler team <https://www.rust-lang.org/governance/teams/compiler>
+---
+
+The Rust teams are always excited to report on new features offered with each release. Sometimes, however, an important change that is not yet "fully baked" gets accidentally included in a release.
+
+There was an instance of this in the most recent release, 1.52.0, which added a new bit of internal-consistency checking, called "incremental compilation hash verification" (abbreviated `verify-ich`). This check is also called an "unstable fingerprint" check, because the diagnostic it currently prints look [like this](https://github.com/rust-lang/rust/issues/84336):
```
camelid, line 10: I think you're missing an "s" here:

```suggestion
There was an instance of this in the most recent release, 1.52.0, which added a new bit of internal-consistency checking, called "incremental compilation hash verification" (abbreviated `verify-ich`). This check is also called an "unstable fingerprint" check, because the diagnostic it currently prints looks [like this](https://github.com/rust-lang/rust/issues/84336):
```

Mark-Simulacrum:
Just pushed an update in light of my expectation that we'll be releasing a point release on Monday - see https://github.com/rust-lang/rust/pull/85097 - but I expect to continue working on the language tomorrow, thank you everyone for the comments so far.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -1,11 +1,36 @@
...
-author: Felix Klock
+title: "Rust 1.52.1"
+author: Felix Klock, Mark Rousskov
 team: the compiler team <https://www.rust-lang.org/governance/teams/compiler>
+release: true
 ---
 
-The Rust teams are always excited to report on new features offered with each release. Sometimes, however, an important change that is not yet "fully baked" gets accidentally included in a release.
+The Rust team has prepared a new release, 1.52.1, working around a bug
+introduced in 1.52.0. We recommend all Rust users, including those currently
```
carols10cents, line 10: From what I understand of previous comments yinz have made, the bug(s) weren't *introduced* in 1.52.0 so much as *surfaced* or *made into a compiler error*. I'd also mention incremental compilation in the first sentence. So perhaps something like:

> The Rust team has prepared a new release, 1.52.1, working around a bug in incremental compilation that was made into a compiler error in 1.52.0.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -1,11 +1,36 @@
...
+using stable versions prior to 1.52.0 upgrade to 1.52.1.
```
carols10cents, line 11: This sentence needs another comma to end the appositive phrase:

> We recommend all Rust users, including those currently using stable versions prior to 1.52.0, upgrade to 1.52.1.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -1,11 +1,36 @@
...
+
+If you don't have it already, you can [get `rustup`][install]
+from the appropriate page on our website.
+
+[install]: https://www.rust-lang.org/install.html
+
+# What's in 1.52.1 stable
+
+This point release contains a single change: it disables incremental
+compilation. Read on for more details as to why, and the next steps the Rust
```
carols10cents, line 28: I would say "disables incremental compilation by default" here.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,125 @@
...
+thread 'rustc' panicked at 'assertion failed: `(left == right)`
+  left: `Some(Fingerprint(4565771098143344972, 7869445775526300234))`,
+   right: `Some(Fingerprint(14934403843752251060, 623484215826468126))`: found unstable fingerprints for <massive text describing rustc internals elided>
+
+error: internal compiler error: unexpected panic
+
+note: the compiler unexpectedly panicked. this is a bug.
+```
+
+This internal-consistency check, as stated in the diagnostic, yields an "Internal Compiler Error" (or ICE). In other words, it represents a bug in the internals of the Rust compiler itself. In *this* case, though, the ICE is revealing a bug that 1.) is very likely to predate the 1.52.0 release and 2.) could result in miscompilation if it had not been caught by `verify-ich`.
```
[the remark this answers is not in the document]
Mark-Simulacrum: @carols10cents It might be helpful to clarify what the distinctions between these things are - the *compiler* itself has no UB, but other than that the miscompilations could include any of these. The known case, for example, caused the wrong pointer to be loaded from a vtable (due to one piece of code expecting one layout for it, and another a different layout, IIRC) and called, leading to memory unsafety (and eventually a crash).

I think we could say more about what we mean by miscompilation but I'd like to better understand what you think the distinctions between these cases are so we can better explain it.

Mark-Simulacrum:
Hey @carols10cents I applied a bunch of feedback from your comments I think, though some of them are also a bit stale after some movement - I would really love a review if you have a chance to do so :)

posts/2021-05-07-caught-red-handed.md
```diff
@@ -22,44 +24,51 @@ from the appropriate page on our website.
...
-
-```
-thread 'rustc' panicked at 'assertion failed: `(left == right)`
-  left: `Some(Fingerprint(4565771098143344972, 7869445775526300234))`,
-   right: `Some(Fingerprint(14934403843752251060, 623484215826468126))`: found unstable fingerprints for <massive text describing rustc internals elided>
-
-error: internal compiler error: unexpected panic
+This release works around broken builds on 1.52.0, which are caused by newly
+added verification. The bugs this verification detects are present in all Rust
+versions, and can trigger miscompilations in incremental builds, so downgrading
```
carols10cents, line 31: I would say "all Rust versions since 1.24.0, when incremental compilation was turned on by default", according to here: https://github.com/rust-lang/rust/blob/master/RELEASES.md#cargo-27, I can't find quickly where incremental compilation was first shipped in stable. Effectively, this is "all Rust versions" because people usually upgrade, but I think it helps to be more specific if we can.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,214 @@
...
+
+# Summary
+
+This release works around broken builds on 1.52.0, which are caused by newly
+added verification. The bugs this verification detects are present in all Rust
+versions, and can trigger miscompilations in incremental builds, so downgrading
+to a prior stable version is not a fix.
+
+Users are encouraged to upgrade to 1.52.1 or disable incremental in their local
+environment if on a prior versions: please see the [what you should do][part3]
```
carols10cents, line 35:
```suggestion
environment if on a prior version: please see the [what you should do][part3]
```

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,214 @@
...
+thread 'rustc' panicked at 'assertion failed: `(left == right)`
+  left: `Some(Fingerprint(4565771098143344972, 7869445775526300234))`,
+  right: `Some(Fingerprint(14934403843752251060, 623484215826468126))`: found unstable fingerprints for <massive text describing rustc internals elided>
+
+error: internal compiler error: unexpected panic
+
+note: the compiler unexpectedly panicked. this is a bug.
+```
+
+This is the error caused by the internal consistency check, and as stated in the diagnostic, it yields an "Internal Compiler Error" (or ICE). In other words, it represents a bug in the internals of the Rust compiler itself. In *this* case, the ICE is revealing a bug that predates the 1.52.0 release and could result in miscompilation if it had not been caught by `verify-ich`.
```
carols10cents, line 70: I don't think it hurts to stress repeatedly that this is a bug in *incremental compilation*, not *all* compilation:

```suggestion
This is the error caused by the internal consistency check, and as stated in the diagnostic, it yields an "Internal Compiler Error" (or ICE). In other words, it represents a bug in the internals of the Rust compiler itself. In *this* case, the ICE is revealing a bug in incremental compilation that predates the 1.52.0 release and could result in miscompilation if it had not been caught by `verify-ich`.
```

posts/2021-05-07-caught-red-handed.md
```diff
@@ -110,20 +125,27 @@ Essentially, for some crates, certain sequences of edit-compile cycles will caus
...
 
 They all arise from inconsistencies when comparing the incremental-compilation cache stored on disk against the values computed during a current `rustc` invocation, which means they all arise from using incremental compilation.
 
-There are three ways that you may have incremental compilation turned on: You may have set the [environment variable][env-vars] `CARGO_INCREMENTAL=1`, or you may have enabled the `build.incremental` [setting in your Cargo.toml][cargo-toml], or you may be building with the `dev` or `test` [profiles][], which default to having incremental compilation enabled.
+There are several ways that you may have incremental compilation turned on:
+
+1. You may be building with the `dev` or `test` [profiles][] which default to having incremental compilation enabled.
+2. You may have set the [environment variable][env-vars] `CARGO_INCREMENTAL=1`
+3. You may have enabled the `build.incremental` [setting in your Cargo config][cargo-config]
+4. You may have enabled the `incremental` [setting in your Cargo.toml][cargo-toml] for a given profile
```
carols10cents, line 138: 👍🏻 👍🏻 👍🏻 👍🏻

posts/2021-05-07-caught-red-handed.md
```diff
@@ -22,44 +24,51 @@ from the appropriate page on our website.
...
 
+[part0]: #what-does-the-error-look-like
 [part1]: #what-are-fingerprints-why-are-we-checking-them
 [part2]: #how-does-this-show-up
 [part3]: #what-should-a-rust-programmer-do-in-response
 [part4]: #what-is-the-rust-project-going-to-do-to-fix-this
 
+## What does the error look like?
+
+The error message looks something like this, with the key piece being the "found
```
carols10cents, line 57: 👍🏻 👍🏻 👍🏻 👍🏻

carols10cents (changes requested):
This is lots, lots better!!!

I think the only bits of my first round of feedback that I don't see incorporated are:

- At least a short statement that we will be investigating root causes of how this change got to stable before we expected it to and how to prevent similar situations in the future
- At least a short definition of what "miscompilation" means to give a sense of what kind of "wrong" the compiler may have been silently generating

Without that information, I would bet money that those two topics will be the ones that get the most comments/questions when this is posted.

posts/2021-05-07-caught-red-handed.md
Mark-Simulacrum: Rust 1.24 is correct - I'm going to add this as a footnote, so we're not distracting too much from the central message. It was all the way back in February 2018, so my guess is most users have migrated.

Mark-Simulacrum:
Hopefully addressed all remaining points. Thanks!

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,230 @@
...
+If your project has not adjusted the defaults, then when running `cargo build
+--release` or otherwise in the `release` profile configuration incremental is
+disabled on all Rust versions[^1], and these issues should not affect your release
+builds.
+
+## What should a Rust programmer do in response
+
+The Internal Compiler Error asks you to report a bug, and if you can do so, we still want that information. We *want* to know about the cases that are failing.
+
+But regardless of whether or not you file a bug, the problem here can be resolved by either:
```
est31, line 163:
```suggestion
But regardless of whether or not you file a bug, the problem can be worked around on your end by either:
```

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,230 @@
...
+
+ 1. upgrading to 1.52.1, if you have not yet done so (which will disable
+    incremental for you).
+ 2. deleting your incremental compilation cache (e.g. by running `cargo clean`), or
+ 3. force incremental compilation to be disabled, by setting `CARGO_INCREMENTAL=0` in your environment or `build.incremental` to `false` in the `config.toml`.
```
est31, line 168:
```suggestion
 3. forcing incremental compilation to be disabled, by setting `CARGO_INCREMENTAL=0` in your environment or `build.incremental` to `false` in the `config.toml`.
```

upgrading, deleting, forcing :)

carols10cents (approved):
This looks great now (with the few small edits @est31 pointed out that I agree with). Thank you so much for working on this ❤️

steveklabnik (dismissed):
Thanks for doing such a thorough writeup on this. Looks good to me.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,230 @@
...
+
+We recommend that users of 1.52.0 upgrade to 1.52.1, which disables incremental
+compilation.
+
+We do *not* recommend that users of 1.52.0 downgrade to an earlier version of Rust in response to this problem. As noted above, there is at least one instance of a silent [miscompilation][issue-82920] caused by incremental compilation that was not caught until we added the fingerprint checking.
+
+If a user is willing to deal with the incremental verification ICE's, and wishes
+to opt back into the 1.52.0 behavior, they may set `RUSTC_FORCE_INCREMENTAL` to
```
wesleywiser, line 176: For these users who are willing to put up with some ICEs, would it make sense to ask them to install the 1.53 beta? We have fixed a number in beta although not all of them.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,230 @@
...
+This is intended to be a mitigation that helps the majority of Rust users have
+an upgrade path to a safe Rust compiler which does not have the risk of
+miscompiling their code, but also provide the option for users willing to deal
+with the errors to do so.
+
+We expect to continue to actively invest in fixing the bugs, and depending on
+our confidence in the fixes, may issue a 1.52.2 point release which backports
+those fixes to the stable channel. Users wishing to help us test can use the
+nightly channel, and report bugs to rust-lang/rust with any ICEs they are
+seeing. We do not expect at this time to disable incremental by default on the
```
wesleywiser, line 212: I think this raises the question "will incremental be disabled on beta?"

It might be good to preemptively address that question.

wesleywiser (dismissed):

Mark-Simulacrum:
@wesleywiser I think I'm currently inclined to leave beta and nightly in the non-disabled incremental states, but it's definitely a good idea to call out their status. I think it's not unlikely 1.53 will need to also ship without incremental, though.

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,235 @@
...
+done under the assumption that the new verification would ship in 1.53, not
+1.52.
+
+[issue-84970]: https://github.com/rust-lang/rust/issues/84970
+
+It turns out `verify-ich` was turned on in version 1.52.0, which was [released recently][].
+
+[released recently]: /2021/05/06/Rust-1.52.0.html
+
+This release, 1.52.1, works around the breakage caused by the newly added
```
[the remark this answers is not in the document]
pnkfelix: (This change is to emphasize that the the subject of the sentence is *not* version 1.52.0, which was referenced in the previous sentence.)

posts/2021-05-07-caught-red-handed.md
pnkfelix: mm, good catch; i hadn't thought of the scenario of adjusting the settings for individual profiles in a local manner.

posts/2021-05-07-caught-red-handed.md
pnkfelix: (The text a few paragraphs down does suggest that option, of trying out the 1.53 beta since a number of bugs are fixed there.)

posts/2021-05-07-caught-red-handed.md
pnkfelix: (I think the text in the summary does a good job of explaining that aspect of things. I don't think we need to reiterate the point here.)

wesleywiser (approved):

posts/2021-05-07-caught-red-handed.md
```diff
@@ -0,0 +1,235 @@
+---
+layout: post
+title: "Announcing Rust 1.52.1"
+author: Felix Klock, Mark Rousskov
+team: the compiler team <https://www.rust-lang.org/governance/teams/compiler>
```
pietroalbini, line 5: Should this be the release team as this is a release?

pietroalbini (changes requested):
Looks good!

We need to rename the file to `posts/2021-05-10-Rust-1.52.1.md`, as the date is wrong and the blog post slug is not the one we usually use for releases.

pietroalbini (approved):
Ready to merge as soon as we release!

PR #840, opened 2021-05-13
Add second version of new error handling blog post

yaahc:
This is a follow up from https://github.com/rust-lang/blog.rust-lang.org/pull/799

* [Rendered](https://github.com/yaahc/blog.rust-lang.org/blob/master/posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md)

cc @bstrie @kw-fn @therealprof @rust-lang/libs @nikomatsakis @flip1995 who all gave feedback on the previous version


posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
+---
+layout: post
+title: "What the Error Handling Project Group is Working Towards"
+author: Jane Lusby
+team: the library team <https://www.rust-lang.org/governance/teams/library>
+---
+
+This blog post is a follow up of our [previous](https://blog.rust-lang.org/inside-rust/2020/11/23/What-the-error-handling-project-group-is-working-on.html) post detailing what we're working on now. We've been iterating for a while now on some of the problems that we see with error handling today and have reached the point where we want to describe some of the new changes we're working towards. But first we need to describe the main problems we've identified.
```
yaahc, line 8:
```suggestion
This blog post is a follow up of our [previous](https://blog.rust-lang.org/inside-rust/2020/11/23/What-the-error-handling-project-group-is-working-on.html) post detailing what we're working on now. We've been iterating for a while now on some of the challenges that we see with error handling today and have reached the point where we want to describe some of the new changes we're working towards. But first we need to describe the main challenges we've identified.
```

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+> Disclaimer: *This post is equal parts plan and aspiration. There are technical challenges here to sort out so the final outcome may look rather different from our initial vision, so please don't assume any of this is final.*
+
+## Error Handling Today
+
+The first problem we'd like to solve is that it's easy to lose context accidentally when reporting errors. There are a couple of places this can happen, either when printing an error and forgetting to print sources, when returning an error from main, or when converting a recoverable error into a non recoverable error.
```
yaahc, line 14:
```suggestion
The first challenge we'd like to solve is that it's easy to lose context accidentally when reporting errors. There are a couple of places this can happen, either when printing an error and forgetting to print sources, when returning an error from main, or when converting a recoverable error into a non recoverable error.
```

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+That's what we would like, at least in the version of rust that the error handling project group wants to see, but what we actually get is this...
+
+```
+$ cargo run
+thread 'main' panicked at 'config is always valid and exists: Error(SourceError)', main.rs:4:33
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+```
+
+Now, I definitely don't think this is what we want as a default when promoting recoverable errors to non-recoverable errors! `unwrap` and `expect` work by stringifying the error variant using it's `Debug` impl, but this is often the wrong operation for types that implement the `Error` trait. By converting the `Error` to a `String` we lose access to the pieces of context we carefully split up via the `Error` trait, and in all likelihood the `derive(Debug)` output of our error types won't even include the error messages in our `Display` impls. Rust's panic infrastructure doesn't provide a method for converting an `Error` type into a panic, it only supports converting `Debug` types into panics, and we feel that this is a major problem.
```
yaahc, line 88:
```suggestion
Now, I definitely don't think this is what we want as a default when promoting recoverable errors to non-recoverable errors! `unwrap` and `expect` work by stringifying the error variant using it's `Debug` impl, but this is often the wrong operation for types that implement the `Error` trait. By converting the `Error` to a `String` we lose access to the pieces of context we carefully split up via the `Error` trait, and in all likelihood the `derive(Debug)` output of our error types won't even include the error messages in our `Display` impls. Rust's panic infrastructure doesn't provide a method for converting an `Error` type into a panic, it only supports converting `Debug` types into panics, and we feel that this is a major issue.
```

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+Here we can see the header we provided to indicate we're printing an error, followed by each error message in the chain of sources separated by colons.
+
+But instead all we get is this:
+
+```
+$ cargo run
+Error: invalid config
+```
+
+By default all of the source's error messages are lost. This problem arises from the fact that we used `Display` as the interface to an individual error message. If we could go back we'd currently propose instead adding `fn message(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result'` to the `Error` trait, but that ship has sailed.
```
yaahc, line 121:
```suggestion
By default all of the source's error messages are lost. This arises from the fact that we used `Display` as the interface to an individual error message. If we could go back we'd currently propose instead adding `fn message(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result'` to the `Error` trait, but that ship has sailed.
```

nikomatsakis:
This is great 🎉

BurntSushi (commented):
This looks great! I think the only feedback I have is about the "Guidelines for implementing Display::fmt and Error::source" section. The recommendation itself sounds great, but I wonder if it's worth acknowledging that this will require some kind of migration in the ecosystem. e.g., If I switched my crates to match this recommendation, then I believe it's likely that folks just printing the error without handling the causal chain will end up losing information.

Now maybe this isn't a problem that's terribly common in practice, but it seems good to have on our radar and acknowledge it maybe? To be honest, I don't know if it belongs in this specific blog post, but thought I would throw it out there.

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+### 2. Error Reporter
+
+We would also like to provide a basic error reporter in `std`, and some facilities for making it easy to use, or easy to replace with your own preferred error reporter. Printing an error and its sources is a fundamental operation in Rust, so we want the language to provide a pit of success for reporting, where the easiest thing to do is the right thing. We can't get there completely because we use `Display` for individual error messages, and we can't change that in a backwards compatible fashion, but we hope that adding a convenient method for printing a full chain of errors and some clever lints will relieve most of the pressure.
+
+We plan on fixing this by first adding a `Report` type to the standard library that wraps a `&dyn Error` and implements `Display` such that it prints each source as desired. We would like the output of `Report`'s display method to support the styles of error concatenation that are most common in the rust ecosystem.
```
bstrie, line 147: For my own curiosity, I suppose this subsumes the `{:!}` format specifier I've seen proposed before?

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+So how do we avoid this? We adopt a consistent separation for `Display` and `source` implementations.
+
+## Guidelines for implementing `Display::fmt` and `Error::source`
+
+To resolve this issue, project error handling recently created a guideline for [how to implement `Display::fmt` and `Error::source`](https://github.com/rust-lang/project-error-handling/issues/27#issuecomment-763950178). In it we make the following recommendation:
+
+**An error type with a source error should either return that error via `source` or include that source's error message in it's own `Display` output, but never both.**
+
+We figure the default will be to return error's via source, so that source errors can be reacted to via `downcast` when appropriate. In coming up with this recommendation we had to figure out what the `Error` trait's primary role is in Rust. After discussing it with the library team we concluded that reporting should be treated as the primary role, and that reacting via `downcast` should come second when designing error types.
```
bstrie, line 227:
```suggestion
We figure the default will be to return errors via source, so that source errors can be reacted to via `downcast` when appropriate. In coming up with this recommendation we had to figure out what the `Error` trait's primary role is in Rust. After discussing it with the library team we concluded that reporting should be treated as the primary role, and that reacting via `downcast` should come second when designing error types.
```

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+This recommendation only applies for error types that are exposed as part of library APIs. Internal errors in libraries or and applicatoins can do whatever they want, but as soon as they need to be integrated into other crates by 3rd party users it's important that errors follow a consistent style. If you're interested in our rational or have any comments please check out our github issue on the topic: [rust-lang/project-error-handling#27](https://github.com/rust-lang/project-error-handling/issues/27).
```
bstrie, line 229:
```suggestion
This recommendation only applies for error types that are exposed as part of library APIs. Internal errors in libraries or and applications can do whatever they want, but as soon as they need to be integrated into other crates by 3rd party users it's important that errors follow a consistent style. If you're interested in our rationale or have any comments please check out our github issue on the topic: [rust-lang/project-error-handling#27](https://github.com/rust-lang/project-error-handling/issues/27).
```

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Conclusion
+
+We hope that these changes will significantly improve error handling experience provided by Rust. Error reporting will be more consistent and flexible and let the final application developer define how to format error reports for their specific use-case. It will be a lot harder to accidentally lose information when reporting errors. The tools for reporting errors will be more tightly integrated into the standard library and the language itself and we hope this will have extra benefits to the embedded ecosystem by more universally standardizing on the `Error` trait.
+
+So that's the plan for now, it's not the full plan of all the changes we'd like to make, but we think its the best first step. However, this is by no means set in stone and we're interested in getting feedback from the rest of the community so we can refine our design. So if you have thoughts please let us know, our project group repo is https://github.com/rust-lang/project-error-handling. Please feel free to open an issue or hop in our zulip stream and create a new topic to let us know what you think of this plan.
```
bstrie, line 235: Link to Zulip stream?

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
yaahc: not quite, we just didn't want to distract from the core of the proposal so we're focusing on getting the MVP in, but the format stuff is still on the table as something we could add later that desugars into using `Report` or something.

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+But instead all we get is this:
+
+```
+$ cargo run
+Error: invalid config
+```
+
+By default all of the source's error messages are lost. This arises from the fact that we used `Display` as the interface to an individual error message. If we could go back we'd currently propose instead adding `fn message(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result` to the `Error` trait, but that ship has sailed.
+
+The way that libraries work around this today is by abusing the `Debug` trait. Types like [`eyre`](https://docs.rs/eyre/0.6.5/eyre/trait.EyreHandler.html#tymethod.debug), [`anyhow`](https://docs.rs/anyhow/1.0.40/src/anyhow/fmt.rs.html#19), and even sometimes [`custom error enums`]() use their `Debug` output to print the full chain of errors in a human readable report.
```
yaahc, line 123:
```suggestion
The way that libraries work around this today is by abusing the `Debug` trait. Types like [`eyre`](https://docs.rs/eyre/0.6.5/eyre/trait.EyreHandler.html#tymethod.debug), [`anyhow`](https://docs.rs/anyhow/1.0.40/src/anyhow/fmt.rs.html#19), and even sometimes [`custom error enums`](https://www.lpalmieri.com/posts/error-handling-rust/#error_chain_fmt) use their `Debug` output to print the full chain of errors in a human readable report.
```

yaahc:
> This looks great! I think the only feedback I have is about the "Guidelines for implementing Display::fmt and Error::source" section. The recommendation itself sounds great, but I wonder if it's worth acknowledging that this will require some kind of migration in the ecosystem. e.g., If I switched my crates to match this recommendation, then I believe it's likely that folks just printing the error without handling the causal chain will end up losing information.
> 
> Now maybe this isn't a problem that's terribly common in practice, but it seems good to have on our radar and acknowledge it maybe? To be honest, I don't know if it belongs in this specific blog post, but thought I would throw it out there.

@BurntSushi I feel like this is a really good point and belongs in this blog post, and more specifically I think we should also include guidance on how libraries should manage the migration of their error types to follow the guidance.

As for the nature of the guidance, my guess is that we're going to have to state that (if?) we feel that the change we're recommending constitutes a breaking change. Then we should point out that the break is purely on runtime behavior so it won't be caught by the compiler, and that just incrementing the version number is likely insufficient for helping dependencies upgrade smoothly without breaking any code that used to rely on `Display` or `downcast` behavior. My feeling is that we should push people towards removing the source error message from their `Display` rather than not returning their error via `source`. Between only printing the outermost error message and making it so an error that was previously reacted to via `downcast` now sneaks through, I'd much rather break the display output.

Also, thinking about it more I think it will be easier to add lints to clean up the old `Display` based printing across a codebase than it will be to handle changes in `downcast` behavior. I'm thinking we might want to add a restriction lint against printing `Error`s directly via their `Display` or `Debug` traits, which we can then include in our guidance as a tool to be used for the migration.

BurntSushi:
@yaahc I _think_ we're on the same page. Might be good to have an issue for this on the error handling repo where can dig into it a bit more. Maybe for this blog post all we need to say is that 1) it will require some kind of migration and 2) we are working out the details for that HERE, where "HERE" is a link to an issue seeded with your most recent comment. How does that sound?

yaahc:
> @yaahc I _think_ we're on the same page. Might be good to have an issue for this on the error handling repo where can dig into it a bit more. Maybe for this blog post all we need to say is that 1) it will require some kind of migration and 2) we are working out the details for that HERE, where "HERE" is a link to an issue seeded with your most recent comment. How does that sound?

That sounds like a great idea. I'll create that issue right now.

nikomatsakis (approved):

yaahc:
@nikomatsakis fwiw I don't think this is ready to merge yet. I still want to decide on and fill out a suggested migration plan in https://github.com/rust-lang/project-error-handling/issues/44 before we post this.

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+That's what we would like, at least in the version of rust that the error handling project group wants to see, but what we actually get is this...
+
+```
+$ cargo run
+thread 'main' panicked at 'config is always valid and exists: Error(SourceError)', main.rs:4:33
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+```
+
+Now, I definitely don't think this is what we want as a default when promoting recoverable errors to non-recoverable errors! `unwrap` and `expect` work by stringifying the error variant using it's `Debug` impl, but this is often the wrong operation for types that implement the `Error` trait. By converting the `Error` to a `String` we lose access to the pieces of context we carefully split up via the `Error` trait, and in all likelihood the `derive(Debug)` output of our error types won't even include the error messages in our `Display` impls. Rust's panic infrastructure doesn't provide a method for converting an `Error` type into a panic, it only supports converting `Debug` types into panics, and we feel that this is a major issue.
```
JDuchniewicz, line 88: a newline here would make this more expressive

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+Error:
+    0: invalid config
+    1: config file does not exist
+
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+```
+
+In this error message we can see that we exited because of a panic. We can see what invariant we violated that was supposed to prevent this panic from happening. We can see the location where the panic was produced. And we can see the error message of each error in the chain of errors accessible via `source`.
+
+That's what we would like, at least in the version of rust that the error handling project group wants to see, but what we actually get is this...
```
JDuchniewicz, line 80: capital Rust or lowercase rust? Is there a consensus for using either naming?

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+
+// We have a program that loads a config and expects that
+// loading the config will always succeed.
+fn main() {
+    let _config = load_config()
+        .expect("config is always valid and exists");
+}
+
+// We have a dummy implementation of load_config which
+// always errors, since we're just focusing on diagnostics
```
JDuchniewicz, line 29: errors or errs 😄 ?

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
```diff
@@ -0,0 +1,235 @@
...
+```rust
+fn report(&self) -> impl Display + '_
+where
+    Self: Sized,
+{
+    Report::from(self)
+}
+```
+
+We want the return type here to be generic rather than hard coded to `Report` so that individual error types can provide their own report format if desired. We expect there will be some fun things derive macros can do here to customize error reporting formats. This will work well with composition because the reporter from the outermost type will be used to format the full chain of errors.
```
JDuchniewicz, line 187: replace
> We expect there will be some fun things derive macros can do here to customize error reporting formats.

with
>We expect there will be some fun things, that derive macros can do here to customize error reporting formats.

?

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
yaahc: I think errors is appropriate here, pretty sure it's valid as a verb.

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
yaahc: should be capital everywhere, :+1:

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
yaahc: I don't think this is grammatically correct in English.

posts/inside-rust/2021-05-15-What-the-error-handling-project-group-is-working-towards.md
yaahc: I'm not sure exactly where you were referring to but I took a guess, let me know if it's better.

JDuchniewicz: much better :)

yaahc:
Okay, I think this may be ready now.

yaahc:
Bumping, still ready for approval / publication @rust-lang/core

Manishearth (approved):

PR #847, opened 2021-05-19
Note on using `--bind`.

pnkfelix:
(I don't actually know if the number of people who might need this warrants having it in this readme. But **I** would have benefited from such a note, so here we are. :smile:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,15 @@ any web server to check it out in your browser:
 
 The site is now available at <http://0.0.0.0:8000>.
 
+If you need to preview the site from another machine, pass the host's address via `--bind`. For example, if running on 192.168.72.73:
+
+```console
+> cd site
+> python3 -m http.server --bind 192.168.72.73
+```
+
+The site is now available at <http://192.168.72.73:8000>.
+
 ## Contributing
 
 First of all, thank you!
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ If you need to preview the site from another machine, pass the host's address vi
 
 The site is now available at <http://192.168.72.73:8000>.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,6 +29,15 @@ any web server to check it out in your browser:
 
 The site is now available at <http://0.0.0.0:8000>.
 
+If you need to preview the site from another machine, pass the host's address via `--bind`. For example, if running on 192.168.72.73:
```
Mark-Simulacrum, line 32: Huh. I would have expected this to work with 0.0.0.0...

pnkfelix: Are you saying I should try doing `python3 -m http.server --bind 0.0.0.0` ?

/me tries it.

Okay, you're right, that works too.

(The crucial thing is that it *doesn't* work if you don't pass `--bind` at all)

README.md
```diff
@@ -29,6 +29,15 @@ any web server to check it out in your browser:
...
+
+```console
+> cd site
+> python3 -m http.server --bind 192.168.72.73
```
pnkfelix, line 36:
```suggestion
> python3 -m http.server --bind 0.0.0.0
```

README.md
```diff
@@ -29,6 +29,15 @@ any web server to check it out in your browser:
...
+```
+
+The site is now available at <http://192.168.72.73:8000>.
```
pnkfelix, line 39:
```suggestion
The site is now available at <http://192.168.72.73:8000>.

(As a short-cut: If you pass `--bind 0.0.0.0`, the server will be reachable at all of the host's IP addresses.)
```

Manishearth (approved):

PR #854, opened 2021-06-17
added note that incr-comp is still off by default in 1.53 release

pnkfelix:


posts/2021-06-17-Rust-1.53.0.md
```diff
@@ -116,6 +116,10 @@ Cargo no longer assumes the default `HEAD` of git repositories is named `master`
 This means you no longer need to specify `branch = "main"` for git dependencies
 from a repository where the default branch is called `main`.
 
+### Incremental Compilation remains off by default
+
+As previously discussed on the [blog post for version 1.52.1](/2021/05/10/Rust-1.52.1.html), incremental compilation has been turned off by default on the stable Rust release channel. The feature remains available on the beta and nightly release channels. For the stable release channel, you can explicitly opt into re-enabling incremental compilation via the `RUSTC_FORCE_INCREMENTAL=1` environment variable.
+
```
Mark-Simulacrum, line 122: I think we should say something explicit like "behavior with regards to enabling incremental is unchanged from 1.52.1" to avoid the impression that it's enough to set the environment variable (when in practice you need to pass -Cincremental too).

pnkfelix: Okay, attempted to address this via 8b8a8b9ea583ab8e450efe551261c9d9cb22ef0a

Mark-Simulacrum (approved):

PR #856, opened 2021-06-23
post announcing this fridays steering meeting.

pnkfelix:


pnkfelix:
better late than never!

Mark-Simulacrum (approved):

PR #870, opened 2021-07-22
1.54 release blog post

Mark-Simulacrum:
I have not had a chance to fill in the stabilized APIs and some feedback from the Zulip [thread](https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/1.2E54.20blog.20post) is still unaddressed; it would be great for concrete suggestions on how to improve things. General review is of course appreciated as well.

r? @pietroalbini 

cc @rust-lang/release


posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,74 @@
...
+
+A number of intrinsics for the wasm32 platform have been stabilized, which gives access to the SIMD instructions in WebAssembly.
+
+Notably, unlike the previously stabilized `x86` and `x86_64` intrinsics, these do not have a safety requirement to only be called when the appropriate target feature is enabled. This is because WebAssembly was written from the start to validate code safely before executing it, so instructions are guaranteed to be decoded correctly (or not at all).
+
+This means that we can expose some of the intrinsics as entirely safe functions, for example [`v128_bitselect`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_bitselect.html). However, there are still some intrinsics which are unsafe, such as [`v128_load`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_load.html).
+
+### Incremental Compilation is XXX by default
+
+TODO - Wesley is going to draft text here.
```
Mark-Simulacrum, line 57: cc @wesleywiser as discussed from meeting today, I believe you'll provide the text for this section. Would be great to have it by Tuesday to give some time for review.

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,74 @@
...
+[detailed release notes for 1.54.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: XXX
+
+## What's in 1.54.0 stable
+
+### Attributes can call macros!()
+
+Rust 1.54 supports invoking function-like macros inside attributes. One notable use case for this is including documentation from other files into Rust doc comments. For example, if your project's README represents a good documentation comment, you can use `include_str!` to directly incorporate the contents. Previously, various workarounds allowed similar functionality, but from 1.54 this is much more ergonomic.
```
ehuss, line 31: I'm uncertain if other people will find this important, but I think it might be good to briefly mention that this only works for built-in attributes of the `=EXPRESSION` form.  I don't believe it can be used for proc-macro attributes, or inputs for other styles of attributes.  Maybe something like this?  I'm not sure if "built-in" is jargon the average rust developer will understand, though.

```suggestion
Rust 1.54 supports invoking function-like macros inside built-in attributes. One notable use case for this is including documentation from other files into Rust doc comments. For example, if your project's README represents a good documentation comment, you can use `include_str!` to directly incorporate the contents. Previously, various workarounds allowed similar functionality, but from 1.54 this is much more ergonomic.
```

Mark-Simulacrum: I admit I was unable to find any concise user-intelligible description of what was stabilized, but comments like https://github.com/rust-lang/rust/pull/83366#issuecomment-823628221 and the PR description there (which may be up to date, not sure) suggest any expression or macro would work.

I think https://github.com/rust-lang/rust/pull/83366#issuecomment-831487787 indicates that:

* built-in attributes can get access to the expanded form (e.g., `#[path = foo!()]` gets the expansion of foo!() as input, rather than the expression `foo!()`)
* external attributes get access solely to the non-expanded form (i.e., `#[foo(bar = baz!())]` would give foo baz!() as the input token stream.
  * It's possible that external attributes also now gain access to *any* expression here -- not just macro calls -- e.g., `#[foo = 5]` would now be potentially usable, whereas before it had to be written inside a string. But I'm not quite sure on this. (No builtin macro currently takes non-string values for it's key/value pairs, so this does not apply there).

cc @rust-lang/lang @petrochenkov

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,74 @@
...
+
+Read [here](https://github.com/rust-lang/rust/pull/83366) for more details.
+
+### wasm32 intrinsics stabilized
+
+A number of intrinsics for the wasm32 platform have been stabilized, which gives access to the SIMD instructions in WebAssembly.
+
+Notably, unlike the previously stabilized `x86` and `x86_64` intrinsics, these do not have a safety requirement to only be called when the appropriate target feature is enabled. This is because WebAssembly was written from the start to validate code safely before executing it, so instructions are guaranteed to be decoded correctly (or not at all).
+
+This means that we can expose some of the intrinsics as entirely safe functions, for example [`v128_bitselect`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_bitselect.html). However, there are still some intrinsics which are unsafe, such as [`v128_load`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_load.html).
```
joshtriplett, line 53:
```suggestion
This means that we can expose some of the intrinsics as entirely safe functions, for example [`v128_bitselect`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_bitselect.html). However, there are still some intrinsics which are unsafe because they use raw pointers, such as [`v128_load`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_load.html).
```

Mark-Simulacrum: I intentionally omitted this in case there were other constraints -- the stable docs likely won't be up to date in time for the release with safety comments. There was some discussion on https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/wasm32.20intrinsics.20safety.20docs and https://github.com/rust-lang/stdarch/pull/1184 has been merged, but we're not expecting backports to 1.54 right now.

Mark-Simulacrum: It looks like from a quick skim this is indeed accurate in broad strokes (the more precise thing needed is valid memory for N bytes, doesn't need to be aligned), though, so accepting.

calebcartwright:
Realize rustfmt info isn't typically included in the release notes, but would it be worth including a footnote (not sure where) for https://github.com/rust-lang/rust/pull/86424?

Maybe something like:

> There was a bug in rustfmt on 1.53 where rustfmt wasn't seeing certain files and was unable to format them. That rustfmt bug has been fixed so users may see formatting diffs on 1.54 since rustfmt will be seeing some files for the first time since 1.52

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -67,6 +67,10 @@ TODO
 There are other changes in the Rust 1.54.0 release:
 check out what changed in [Rust](XXX), [Cargo](XXX), and [Clippy](XXX).
 
+rustfmt has also been fixed in the 1.54.0 release to properly format nested
+out-of-line modules. This may cause changes in formatting to files that were
+being ignored by the 1.53.0 rustfmt. See details [here](https://github.com/rust-lang/rust/pull/86424).
```
calebcartwright, line 72: Thank you :+1:

posts/2021-07-29-Rust-1.54.0.md
ehuss: Yea, I think I may have dropped the ball a little here in terms of getting the documentation squared away.  The reference was updated (see [metaitem](https://doc.rust-lang.org/nightly/reference/attributes.html#meta-item-attribute-syntax)), but it isn't very explicit about the implications.  I didn't pursue getting it updated in detail because the reference doesn't cover expansion at all, so it can be difficult to add a small piece like this without the rest of expansion described.  

> It's possible that external attributes also now gain access to any expression here 

To clarify this statement, it's "built-in attributes also now gain". I'm not sure what is meant by "external" here, but proc-macros do not allow name=EXPR syntax.  And no built-in attribute supports non-strings.

But otherwise, yea, I think you have the gist of it.  I think the only attributes where this is really useful is `#[doc]` and `#[path]`.  The others don't seem too useful. If anyone's curious, the stable attributes that are affected are `crate_type`, `deprecated`, `doc`, `export_name`, `ignore`, `link`, `link_name`, `link_section`, `must_use`, `path`, `recursion_limit`, `should_panic`, `type_length_limit`, `windows_subsystem`.  Most of those don't benefit from early expansion.

Mark-Simulacrum: Niko's comment in that thread suggests otherwise - and so does the reference here https://github.com/rust-lang/reference/pull/1029/files#diff-a1074bf6cead5b27ef8c88162d6282ba3c3af77497646a7f22ac614255b26906R113

> In particular, if you have a derive (or whatever) with a custom attribute like #[foo = "bar"], you can't do:

```
#[derive(MyDerive)]
struct Foo {
    #[foo = include_str!(...)]
    x: u32
}
```

> unless you are able to pipe the include_str! through to your output.

ehuss: I think those are referring to derive macro helper attributes, which have allowed accepting arbitrary expressions since 1.30. Or maybe I'm misunderstanding what you mean?

ehuss: Or, maybe you're referring to helper attributes now allowing a macro call syntax in that position, which is new, though I'm not sure how useful that is.

Mark-Simulacrum: I'm not sure to be honest :) I think it'd be good to update the language in the blog post to be accurate, but I don't feel confident myself I understand enough quite yet to do so. Certainly it sounds like my interpretation of comments on the stabilization PR didn't quite match reality from what you say....

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,78 @@
...
+A number of intrinsics for the wasm32 platform have been stabilized, which gives access to the SIMD instructions in WebAssembly.
+
+Notably, unlike the previously stabilized `x86` and `x86_64` intrinsics, these do not have a safety requirement to only be called when the appropriate target feature is enabled. This is because WebAssembly was written from the start to validate code safely before executing it, so instructions are guaranteed to be decoded correctly (or not at all).
+
+This means that we can expose some of the intrinsics as entirely safe functions, for example [`v128_bitselect`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_bitselect.html). However, there are still some intrinsics which are unsafe because they use raw pointers, such as [`v128_load`](https://doc.rust-lang.org/beta/core/arch/wasm32/fn.v128_load.html).
+
+### Incremental Compilation is XXX by default
+
+TODO - Wesley is going to draft text here.
+
```
wesleywiser, line 58:
```suggestion
### Incremental Compilation is re-enabled by default

In Rust 1.52, additional validation was added when loading incremental compilation data from the on-disk cache.
This resulted in a number of pre-existing issues being uncovered as the validation changed these silent bugs into internal compiler errors (ICEs).
In response, the Compiler Team decided to disable incremental compilation in the 1.52.1 patch, allowing users to avoid encountering the ICEs. [^1]

Since then, we've conducted a [series of retrospectives][retros] and contributors have been hard at work resolving the reported issues with some fixes landing in 1.53 and the majority landing in this release. [^2]
There are currently still two known issues which can result in an ICE.
However, based on the feedback we received from users affected by the 1.52 release, we believe these issues to be rare.

Therefore, incremental compilation has been re-enabled in this release!

[^1]: The [1.52.1 release notes] contain a more detailed description of these events.
[^2]: The tracking issue for the issues is [#84970].

[#84970]: https://github.com/rust-lang/rust/issues/84970
[1.52.1 release notes]: https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html
[retros]: https://github.com/rust-lang/compiler-team/issues/435

```

cc @pnkfelix

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,78 @@
...
+### Stabilized APIs
+
+The following methods and trait implementations were stabilized.
+
+TODO
```
XAMPPRocky, line 63:
```suggestion
- [`BTreeMap::into_keys`]
- [`BTreeMap::into_values`]
- [`HashMap::into_keys`]
- [`HashMap::into_values`]
- [`arch::wasm32`]
- [`VecDeque::binary_search`]
- [`VecDeque::binary_search_by`]
- [`VecDeque::binary_search_by_key`]
- [`VecDeque::partition_point`]

[`BTreeMap::into_keys`]: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.into_keys
[`BTreeMap::into_values`]: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.into_values
[`HashMap::into_keys`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.into_keys
[`HashMap::into_values`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.into_values
[`arch::wasm32`]: https://doc.rust-lang.org/core/arch/wasm32/index.html
[`VecDeque::binary_search`]: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.binary_search
[`VecDeque::binary_search_by`]: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.binary_search_by
[`VecDeque::binary_search_by_key`]: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.binary_search_by_key
[`VecDeque::partition_point`]: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.partition_point
```

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Other changes
+
+There are other changes in the Rust 1.54.0 release:
+check out what changed in [Rust](XXX), [Cargo](XXX), and [Clippy](XXX).
```
XAMPPRocky, line 68:
```suggestion
check out what changed in [Rust](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1540-2021-07-29), [Cargo](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-154-2021-07-29), and [Clippy](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-154).
```

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,78 @@
...
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][install]
+from the appropriate page on our website, and check out the
+[detailed release notes for 1.54.0][notes] on GitHub.
+
+[install]: https://www.rust-lang.org/install.html
+[notes]: XXX
```
XAMPPRocky, line 25:
```suggestion
[notes]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1540-2021-07-29
```

posts/2021-07-29-Rust-1.54.0.md
pnkfelix: Nit: I'd avoid passive voice. E.g., instead of "This resulted in a number of pre-existing issues being uncovered ...", I'd say "This uncovered a number of pre-existing issues ..."

pnkfelix: In the compiler team meeting, we touched on the topic that the lack of telemetry data from the compiler impedes the feedback cycle: It makes it impossible for us to measure true impact, and also any feedback we get is relying on a human being to decide to report the issue. Did we want to make any mention of that here, as a way to explain how we are operating somewhat in the dark at this moment?

estebank: I would lead with the good news.

```suggestion
### Incremental Compilation is re-enabled by default

Incremental compilation has been re-enabled by default in this release, after it being disabled by default in 1.52.

In Rust 1.52, additional validation was added when loading incremental compilation data from the on-disk cache.
This resulted in a number of pre-existing potential soundness issues being uncovered as the validation changed these silent bugs into internal compiler errors (ICEs).
In response, the Compiler Team decided to disable incremental compilation in the 1.52.1 patch, allowing users to avoid encountering the ICEs and the underlying unsoundness, at the expense of longer compile times. [^1]

Since then, we've conducted a [series of retrospectives][retros] and contributors have been hard at work resolving the reported issues, with some fixes landing in 1.53 and the majority landing in this release. [^2]

There are currently still two known issues which can result in an ICE.
Due to the lack of automated crash reporting, we can't be certain of the full extent of impact of the outstanding issues. However, based on the feedback we received from users affected by the 1.52 release, we believe these issues to be rare.

Therefore, incremental compilation has been re-enabled in this release!

[^1]: The [1.52.1 release notes] contain a more detailed description of these events.
[^2]: The tracking issue for the issues is [#84970].
[#84970]: https://github.com/rust-lang/rust/issues/84970
[1.52.1 release notes]: https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html
[retros]: https://github.com/rust-lang/compiler-team/issues/435
```

posts/2021-07-29-Rust-1.54.0.md
```diff
@@ -0,0 +1,113 @@
...
+This resulted in a number of pre-existing potential soundness issues being uncovered as the validation changed these silent bugs into internal compiler errors (ICEs).
+In response, the Compiler Team decided to disable incremental compilation in the 1.52.1 patch, allowing users to avoid encountering the ICEs and the underlying unsoundness, at the expense of longer compile times. [^1]
+
+Since then, we've conducted a [series of retrospectives][retros] and contributors have been hard at work resolving the reported issues, with some fixes landing in 1.53 and the majority landing in this release. [^2]
+
+There are currently still two known issues which can result in an ICE.
+Due to the lack of automated crash reporting, we can't be certain of the full extent of impact of the outstanding issues. However, based on the feedback we received from users affected by the 1.52 release, we believe the remaining issues to be rare in practice.
+
+Therefore, incremental compilation has been re-enabled in this release!
+
```
wesleywiser, line 69:
````suggestion
The Rust compiler has also been updated to include a more detailed error message with instructions to help the user work around the issue if it were to occur:

```
   Compiling example v0.1.0 (/path/to/example)
error: internal compiler error: encountered incremental compilation error with query(hash)
  |
  = help: This is a known issue with the compiler. Run `cargo clean -p example` or `cargo clean` to allow your project to compile
  = note: Please follow the instructions below to create a bug report with the provided information
  = note: See <https://github.com/rust-lang/rust/issues/84970> for more information

thread 'rustc' panicked at 'Found unstable fingerprints for query(hash): QueryResult', /rustc/compiler/rustc_query_system/src/query/plumbing.rs:619:9
```

````

pnkfelix: Should the note take the further step and actively emphasize that we really do want people to follow through and file bugs when they see this message, since, as alluded to above, that is our main avenue for feedback here?

I'm torn about this, because I don't want to turn the release blog post into a post just about the fingerprint bug. After all, our hope here is that this really will be rare now...

wesleywiser: I had a similar thought while writing this part and decided against it because the error message says to report the issue. I'll go ahead and add that and we can see if we like that better.

wesleywiser: Should we make it longer than that? I think short and to the point is better here and we want to get reports of all ICEs, not just this one.

pietroalbini: I'm leaning towards not including the error message: it's great that the diagnostics improved for this, but focusing on the error message seems... wrong for the release announcement :)

posts/2021-07-29-Rust-1.54.0.md
jyn514: The PR description is up to date, the relevant bit is

> Accepted syntax
The parser accepts arbitrary Rust expressions in this position, but any expression other than a macro invocation will ultimately lead to an error because it is not expected by the built-in expression forms (e.g., #[doc]). Note that decorators and the like may be able to observe other expression forms.

A more concise version is "In key-value attributes, arbitrary expressions are now allowed syntactically but not semantically; to be accepted semantically, they must expand to literal values".

> Niko's comment in that thread suggests otherwise

Niko was asking how this interacts with proc macros, in particular whether the attributes are eagerly- or lazily- expanded. The answer is they are lazily expanded.

> external attributes get access solely to the non-expanded form (i.e., #[foo(bar = baz!())] would give foo baz!() as the input token stream.
It's possible that external attributes also now gain access to any expression here -- not just macro calls -- e.g., #[foo = 5] would now be potentially usable, whereas before it had to be written inside a string.

I don't know what "external" means here, but otherwise you're correct, this only applies to top-level key-value attributes, not nested meta items. The parser doesn't treat the `x = y` in `foo(x = y)` as a key-value attribute but as an arbitrary stream of tokens. (This was discussed in more detail in the internals blog post.)

Mark-Simulacrum: A more concise version is "In key-value attributes, arbitrary expressions are now allowed syntactically but not semantically; to be accepted semantically, they must expand to literal values".

To be clear, this is more so saying that the new stabilization does not change the behavior of any built-in macros to accept non-string-literal values, right? Otherwise, there's not actually a restriction in some sense.

And this also means that it *is* true that proc macros are fully able to accept arbitrary Rust expressions. However, IIRC, we do validate that they must be valid Rust expressions (rather than random tokens).

jyn514: > To be clear, this is more so saying that the new stabilization does not change the behavior of any built-in macros to accept non-string-literal values, right? Otherwise, there's not actually a restriction in some sense.

This does not change the behavior of any built-in macros, but it's now checked at the macro definition site instead of in the parser itself. You could define a macro that accepted `#[x = 5]`, there just doesn't happen to be anything in the standard library that accepts that.

> And this also means that it is true that proc macros are fully able to accept arbitrary Rust expressions. However, IIRC, we do validate that they must be valid Rust expressions (rather than random tokens).

Correct.

posts/2021-07-29-Rust-1.54.0.md
estebank: I think I agree with @pietroalbini here. The error already has a call to action... Anyone who should care will see it and hopefully that is enough.

pietroalbini (approved):

PR #872, opened 2021-07-24
Add post on x.py and submodules

jyn514:
r? @Mark-Simulacrum

### DIFF
--- a/posts/inside-rust/2021-07-29-speeding-up-submodules-for-fun-and-profit.md
+++ b/posts/inside-rust/2021-07-29-speeding-up-submodules-for-fun-and-profit.md
@@ -1,0 +1,83 @@
+---
+layout: post
+title: "Speeding up submodules for fun and profit"
+author: Joshua Nelson
+description: "What are submodules, how did they get faster, and why did it take so long?"
+---
+
+`x.py` now has faster handling of submodules! We need your help to make x.py faster :)
+
+## How can I opt-in to these changes?
+
+If you're cloning the rust repository for the first time, no action is necessary. If you have an
+existing clone of `rust-lang/rust`, you can almost double the speed of running `git fetch` by
+running the following command:
+
+```sh
+git submodule deinit src/doc/book src/doc/edition-guide src/doc/embedded-book src/doc/nomicon src/doc/reference src/doc/rust-by-example src/doc/rustc-dev-guide src/llvm-project src/tools/rust-analyzer
+```
+
+You should see something similar to the following output:
+
+```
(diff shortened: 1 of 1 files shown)


jyn514:
What team do you think I should put for this? T-infra doesn't seem right since bootstrap isn't really owned by them, and T-compiler doesn't seem right either. I guess T-core works? But this is pretty far outside their responsibilities.

pietroalbini:
Hmm, x.py is currently maintained by Mark, I'm not sure whether the hat is infra or release.

Mark-Simulacrum:
x.py is not really maintained with a team hat. I don't have a particular perspective on what team it *should* be under.

I haven't had a chance to take a look at this yet.

jyn514:
@Mark-Simulacrum do you know when you'll get a chance to look at this?

jyn514:
ping @Mark-Simulacrum

posts/inside-rust/2021-07-29-speeding-up-submodules-for-fun-and-profit.md
```diff
@@ -0,0 +1,83 @@
...
+    - Tarballs with the source code don't have `.gitmodules` available, so it's not possible to tell what the submodules are in the first place. [82653]
+    - The CI for `rust-lang/rust` actually bypasses git altogether and just downloads a zip archive, but still runs `touch .git` for each submodule. [81601]
+
+These were especially hard to fix because it's very difficult to test changes to submodule updating
+- the expected behavior is unclear and usually isn't known until people complain that you've broken
+their workflow. However, all known issues have now been fixed. There are still some remaining issues -
+in particular, `x.py` is only smart for submodules which aren't in the Cargo workspace. Unfortunately,
+Cargo requires that we have all the Cargo.toml files available whenever it's run, even if it's not yet
+compiling those projects. I hope to [extend this lazy-cloning to members of the Cargo workspace][76653] in
+the future.
```
Mark-Simulacrum, line 76: I feel like this is a little bit too pessimistic -- I wonder if you'd be up for adjusting this to be a little more optimistic about the future?

In particular, I think reframing this as "here's what went into this and why it was hard", rather than focusing on it taking a long time, would be better -- and also lets us spend a little time educating on submodules and the use cases we learned about, which may be nice.

I think the core section on various workflows is good (bulleted list), and just adjusting some of the paragraphs around it will help with framing it in a more positive light. Alternatively, we can also drop this section.

jyn514:
I don't have the energy to be less pessimistic here.

PR #897, opened 2021-10-19
Announcing Rust 1.56.0 and Rust 2021

cuviper:


cuviper:
cc @rust-lang/release

PR #904, opened 2021-11-07
Try using real bidi in CVE-2021-42574.md

SamB:
This wouldn't really matter, except that the syntax highlighting is different.
Alternative: just disable the highlighting.

[I can certainly see why you wouldn't want to include the bidi characters in the email :-).]

### DIFF
--- a/posts/2021-11-01-cve-2021-42574.md
+++ b/posts/2021-11-01-cve-2021-42574.md
@@ -38,7 +38,13 @@ codepoint `NNNN`):
 if access_level != "user{U+202E} {U+2066}// Check if admin{U+2069} {U+2066}" {
 ```
 
-...would be rendered by bidirectional-aware tools as:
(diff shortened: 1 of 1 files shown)


SamB:
I basically did this because, as I tried to read <https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html>, I thought to myself "but surely the syntax highlighter wouldn't be fooled by those bidi control characters" ...

Turns out I was at least partly right.

pietroalbini:
I would prefer not to change this, mostly to ensure consistent rendering: we don't know whether all the programs displaying the blog post correctly support bidi (browsers do, but for example do all feed readers?), so if someone doesn't use a tool properly implementing bidi they might not get what the security advisory is about.

pietroalbini:
Still, thanks for taking the time to send the PR!

PR #906, opened 2021-11-16
Announce std::simd is on nightly

workingjubilee:
Not 100% sure if this should go on the main blog or Inside Rust.

### DIFF
--- a/posts/inside-rust/2021-11-16-std-simd-on-nightly.md
+++ b/posts/inside-rust/2021-11-16-std-simd-on-nightly.md
@@ -1,0 +1,114 @@
+---
+layout: post
+title: "std::simd available in nightly"
+author: Caleb Zulawski and Jubilee Young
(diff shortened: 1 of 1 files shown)


rylev:
@workingjubilee sorry for letting this sit for so long. Would it still make sense for this to be updated and released or is it too late, and this should be closed?

programmerjake:
imho this is still useful, though it may need to be updated to the latest portable-simd API.

Turbo87:
it's been over two years, if this is actually still relevant then it's probably best to open a new PR with an updated text 😉

workingjubilee:
@Turbo87 What are you winking at? Our discords?

And 3, actually.

Having to answer "this should be closed" for a reason that is no fault of my own is demoralizing, which is half of why I simply did not. The other half is the lack of reason to believe that I would not get a response, if I opened something now, sooner than 2025, unless I did something to further piss everyone off.

Turbo87:
I was just assuming that the content is probably not up-to-date anymore and since there hasn't been any movement for a long while I closed it, but feel free to reopen if you still want to land it :)

PR #931, opened 2022-02-16
Blog post for Rust 1.59.0

tmandry:


posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,242 @@
...
+    );
+}
+assert_eq!(x, 4 * 6);
+```
+
+The format string syntax used to name registers in the `asm!` and `global_asm!`
+macros is the same used in Rust format strings, so it should feel quite familiar
+to Rust programmers.
+
+The syntax and instructions available in the assembly code itself vary according
```
Mark-Simulacrum, line 56:
```suggestion
The assembly language and instructions available in the assembly code itself vary according
```

Maybe? Otherwise it might be interpreted as 'format string syntax' varies, which is not the case, I believe? Maybe we can also reword to something simpler, like "Each target has a fairly different set of instructions, so usage would often be cfg-gated in practice" or something like that?

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+You can run the `cargo report` command mentioned in the warning to see a full
+report of the code that will be rejected. This gives you time to upgrade your
+dependency before it breaks your build.
+
+### Creating stripped binaries
+
+It's often useful to strip unnecessary information like debuginfo from binaries
+you distribute, making them smaller.
```
Mark-Simulacrum, line 151: Maybe add something about std shipping with debuginfo = 1 by default or something, which is why release builds may benefit from this even if they don't themselves turn on debug?

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,242 @@
...
+The format string syntax used to name registers in the `asm!` and `global_asm!`
+macros is the same used in Rust format strings, so it should feel quite familiar
+to Rust programmers.
+
+The syntax and instructions available in the assembly code itself vary according
+to the target architecture. Today, the following architectures are supported:
+
+* x86 and x86-64
+* ARM
+* AArch64
```
CryZe, line 61: Considering for the first time aarch64 neon intrinsics are stable, would it maybe make sense to highlight that in this section or another some more?

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,242 @@
...
+
+## What's in 1.59.0 stable
+
+### Inline assembly
+
+The Rust language now supports inline assembly. This enables many applications
+of Rust that need very low-level control over their execution, or access to
+specialized machine instructions.
+
+When compiling for x86 and x86-64 targets, for instance, you can now write:
```
joshtriplett, line 32:
```suggestion
When compiling for x86-64 targets, for instance, you can now write:
```

The asm as written doesn't work on x86-32 because it wants a 64-bit register. Changing the `u64` to a `u32` would work on x86-32, but require some additional distracting changes to work on x86-64 (to use an `e` register rather than an `r` register). So, just narrow the example to x86-64.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,242 @@
...
+}
+assert_eq!(x, 4 * 6);
+```
+
+The format string syntax used to name registers in the `asm!` and `global_asm!`
+macros is the same used in Rust format strings, so it should feel quite familiar
+to Rust programmers.
+
+The syntax and instructions available in the assembly code itself vary according
+to the target architecture. Today, the following architectures are supported:
```
joshtriplett, line 57:
```suggestion
to the target architecture. Today, the stable Rust compiler supports inline assembly on the following architectures:
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,242 @@
...
+
+(a, b) = (1, 2);
+[c, .., d, _] = [1, 2, 3, 4, 5];
+Struct { e, .. } = Struct { e: 5, f: 3 };
+
+assert_eq!([1, 2, 1, 4, 5], [a, b, c, d, e]);
+```
+
+This makes assignment more consistent with let bindings, which allow you to do
+the same thing. Note that assignments with operators such as `+=` are not
```
joshtriplett, line 86:
```suggestion
the same thing. Note that destructuring assignments with operators such as `+=` are not
```

posts/2022-02-24-Rust-1.59.0.md
Mark-Simulacrum: I think these are noted in the stabilized APIs section -- I definitely wouldn't add them to the asm section, but I think we could add a separate block. I'm not sure how significant their stabilization is.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,242 @@
...
+enable this, add the following to your `Cargo.toml`:
+
+```toml
+[profile.release]
+strip = "debuginfo"
+```
+
+This causes debuginfo to be stripped from release binaries. You can also supply
+`"symbols"` or just `true` to strip all symbol information where supported.
+
```
joshtriplett, line 164:
```suggestion

The standard library ships with debug symbols by default, so Rust binaries built without debug symbols enabled still by default include the debug symbols of the standard library. The `strip` allows you to remove those symbols, producing smaller Rust binaries.

```

posts/2022-02-24-Rust-1.59.0.md
joshtriplett: Done in a suggestion below.

pietroalbini:
We also need to mention that we disabled incremental compilation (https://github.com/rust-lang/rust/issues/94124).

estebank:
It should also include information on [how to enable it](https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html#short-term-plan).

Mark-Simulacrum:
I pushed an initial section on incr comp -- would appreciate feedback, not sure how it reads just yet.

I also put it pretty far down, but we may want to raise it up higher in the post, not sure which is best.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+compilation, new bugs may get surfaced which were previously hidden; due to a
+reluctance to ship unknown problems we have avoided aggressively backporting
+incremental-related changes. This is in part due to having limited ability to
+exercise incremental in automated tests, which means that we currently must rely
+on user-submitted issues to determine the likely level of impact and make
+decisions to enable or disable incremental for a given release cycle.
+
+This particular bug is already fixed on beta and should roll out to stable 1.60.
+We have not made a decision yet as to whether 1.60 will re-enable incremental
+compilation by default.
```
estebank, line 221: Although being noncommittal is ok, I wouldn't want us to give the wrong impressions that we wouldn't reenable it if we don't find other cases in the wild.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+### Incremental compilation off by default
+
+The 1.59.0 release disables incremental by default (unless explicitly asked for
+by an environment variable, with `RUSTC_FORCE_INCREMENTAL=1`). This mitigates a
+[known bug] which can cause deserialization errors (and panics) during compilation
+with incremental compilation turned on.
+
+This particular bug is preexisting, and was exposed due to other bugs fixed in
+incremental in the 1.59 release cycle. As we continue to fix bugs in incremental
+compilation, new bugs may get surfaced which were previously hidden; due to a
```
estebank, line 212:
```suggestion
compilation, new bugs may get surfaced which were previously hidden. Due to a
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+- [`ffi::CStr::from_bytes_with_nul_unchecked`][cstr_from_bytes]
+
+### Incremental compilation off by default
+
+The 1.59.0 release disables incremental by default (unless explicitly asked for
+by an environment variable, with `RUSTC_FORCE_INCREMENTAL=1`). This mitigates a
+[known bug] which can cause deserialization errors (and panics) during compilation
+with incremental compilation turned on.
+
+This particular bug is preexisting, and was exposed due to other bugs fixed in
```
estebank, line 210:
```suggestion
This particular bug is preexisting and had already been fixed on the nightly and beta corresponding to 1.60. It was exposed due to other bugs fixed in
```

estebank:
For what is worth, the wording in the release notes isn't bad: https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-1590-2022-02-24

Mark-Simulacrum:
Yeah, maybe we can copy that. That's my yesterday-night take on it :)

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+        "add {x}, {tmp}",
+        x = inout(reg) x,
+        tmp = out(reg) _,
+    );
+}
+assert_eq!(x, 4 * 6);
+```
+
+The format string syntax used to name registers in the `asm!` and `global_asm!`
+macros is the same used in Rust format strings, so it should feel quite familiar
```
pierwill, line 53: Maybe a link on  `format strings` to https://doc.rust-lang.org/std/fmt?

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+to Rust programmers.
+
+The assembly language and instructions available in the assembly code itself vary according
```
pierwill, line 56:
```suggestion
The assembly language and instructions available with inline assembly vary according
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+
+You can see more examples of inline assembly in [Rust By Example][asm-example],
+and find more detailed documentation in the [reference][asm-reference].
+
+[asm-example]: https://doc.rust-lang.org/nightly/rust-by-example/unsafe/asm.html
+[asm-reference]: https://doc.rust-lang.org/nightly/reference/inline-assembly.html
+
+### Destructuring assignments
+
+You can now use tuple, slice, and struct patterns as the left hand side of an
```
pierwill, line 72:
```suggestion
You can now use tuple, slice, and struct patterns as the left-hand side of an
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+This causes debuginfo to be stripped from release binaries. You can also supply
+`"symbols"` or just `true` to strip all symbol information where supported.
+
+The standard library typically ships with debug symbols, so Rust binaries built without debug symbols enabled still include the debug symbols of the standard library by default. Using `strip` allows you to remove those symbols, producing smaller Rust binaries.
+
+### Stabilized APIs
+
+The following methods and trait implementations were stabilized.
```
pierwill, line 169:
```suggestion
The following methods and trait implementations are now stabilized:
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+- [`NonZeroU64::is_power_of_two`][is_power_of_two64]
+- [`NonZeroU128::is_power_of_two`][is_power_of_two128]
+- [`DoubleEndedIterator for ToLowercase`][lowercase]
+- [`DoubleEndedIterator for ToUppercase`][uppercase]
+- [`TryFrom<&mut [T]> for [T; N]`][tryfrom_ref_arr]
+- [`UnwindSafe for Once`][unwindsafe_once]
+- [`RefUnwindSafe for Once`][refunwindsafe_once]
+- [armv8 neon intrinsics for aarch64][stdarch/1266]
+
+The following previously stable functions are now `const`.
```
pierwill, line 196:
```suggestion
The following previously stable functions are now `const`:
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+
+This particular bug is already fixed on beta and should roll out to stable 1.60.
+We have not made a decision yet as to whether 1.60 will re-enable incremental
+compilation by default.
+
+[known bug]: https://github.com/rust-lang/rust/issues/94124
+
+### Other changes
+
+There are other changes in the Rust 1.59.0 release: check out what changed in
```
pierwill, line 227:
```suggestion
There are other changes in the Rust 1.59.0 release. Check out what changed in
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -200,6 +200,26 @@ The following previously stable functions are now `const`.
...
+compilation, new bugs may get surfaced which were previously hidden; due to a
+reluctance to ship unknown problems we have avoided aggressively backporting
+incremental-related changes. This is in part due to having limited ability to
+exercise incremental in automated tests, which means that we currently must rely
+on user-submitted issues to determine the likely level of impact and make
+decisions to enable or disable incremental for a given release cycle.
+
+This particular bug is already fixed on beta and should roll out to stable 1.60.
+We have not made a decision yet as to whether 1.60 will re-enable incremental
+compilation by default.
```
tmandry, line 221: I think this paragraph should go second, or at least the first sentence should be moved up. The wording "we have avoided aggressively backporting incremental-related changes" above makes more sense given this context. You could potentially pull that into this paragraph as well, which frames the situation more positively – we already have a fix, but want to ensure that we're shipping high quality releases by following the release process. This feature only affects build times so it's safest to disable for this release.

Also, "This particular bug is" is repeated at the beginning of both paragraphs.

posts/2022-02-24-Rust-1.59.0.md
estebank: Make it clear that there's high likelihood that incremental will be enabled in 1.60, because *this* bug has already been fixed there. The wording here is present of course because *other* bugs might crop up.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,266 @@
...
+
+```toml
+[profile.release]
+strip = "debuginfo"
+```
+
+This causes debuginfo to be stripped from release binaries. You can also supply
+`"symbols"` or just `true` to strip all symbol information where supported.
+
+The standard library typically ships with debug symbols, so Rust binaries built without debug symbols enabled still include the debug symbols of the standard library by default. Using `strip` allows you to remove those symbols, producing smaller Rust binaries.
```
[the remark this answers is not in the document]
Mark-Simulacrum: The option should be capable of stripping symbols, yes. I've adjusted the wording to hopefully make that clearer.

Mark-Simulacrum:
Ok, I've updated the post to reflect feedback provided so far, and refactored the incremental section once more. I'm still unsure whether putting in essentially at the end of the post vs. the beginning makes sense -- I can see pros and cons to either strategy. I think it'll likely remain as-is unless there's strong feedback otherwise.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,274 @@
...
+
+- [`mem::MaybeUninit::as_ptr`][muninit_ptr]
+- [`mem::MaybeUninit::assume_init`][muninit_init]
+- [`mem::MaybeUninit::assume_init_ref`][muninit_init_ref]
+- [`ffi::CStr::from_bytes_with_nul_unchecked`][cstr_from_bytes]
+
+### Incremental compilation off by default
+
+The 1.59.0 release disables incremental by default (unless explicitly asked for
+by via an environment variable: `RUSTC_FORCE_INCREMENTAL=1`). This mitigates a
```
pierwill, line 215:
```suggestion
by via an environment variable: `RUSTC_FORCE_INCREMENTAL=1`). This mitigates
```

matklad:
I feel we probably should tone down the announcement, the way we did for 1.44.0.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,274 @@
...
+
+- [`mem::MaybeUninit::as_ptr`][muninit_ptr]
+- [`mem::MaybeUninit::assume_init`][muninit_init]
+- [`mem::MaybeUninit::assume_init_ref`][muninit_init_ref]
+- [`ffi::CStr::from_bytes_with_nul_unchecked`][cstr_from_bytes]
+
+### Incremental compilation off by default
+
+The 1.59.0 release disables incremental by default (unless explicitly asked for
+by via an environment variable: `RUSTC_FORCE_INCREMENTAL=1`). This mitigates
```
bjorn3, line 215: Could you reword thus to avoid "for by via"?

estebank, line 215:
```suggestion
The 1.59.0 release disables incremental by default (unless explicitly asked for
by setting an environment variable: `RUSTC_FORCE_INCREMENTAL=1`). This mitigates
```

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,274 @@
...
+
+[cargo-docs]: https://doc.rust-lang.org/beta/cargo/reference/profiles.html#strip
+
+### Stabilized APIs
+
+The following methods and trait implementations are now stabilized:
+
+- [`std::thread::available_parallelism`][available_parallelism]
+- [`hash_map::Entry::insert_entry`][entry_insert_entry]
+- [`hash_map::VacantEntry::insert_entry`][vacant_insert_entry]
```
m-ou-se, line 182: These stabilizations were reverted iirc.

posts/2022-02-24-Rust-1.59.0.md
```diff
@@ -0,0 +1,281 @@
...
+binaries.
+
+See [Cargo's documentation][cargo-docs] for more details.
+
+[cargo-docs]: https://doc.rust-lang.org/beta/cargo/reference/profiles.html#strip
+
+### Incremental compilation off by default
+
+The 1.59.0 release disables incremental by default (unless explicitly asked for
+by via an environment variable: `RUSTC_FORCE_INCREMENTAL=1`). This mitigates a
```
pierwill, line 190: Somehow this typo crept back in?

```suggestion
by via an environment variable: `RUSTC_FORCE_INCREMENTAL=1`). This mitigates
```

pierwill (changes requested):

PR #933, opened 2022-02-17
Fix link to source-based code coverage guide

pwoolcoc:
When I searched for "rust code coverage", this blog post was the very first result in Google. The "Take a look at the guide" link, however, 404'd. I've found what I *think* is the best replacement link, but I'm happy to correct it if there is a better source of info.


pwoolcoc:
@ehuss thanks, fixed

rylev (approved):
Thanks!

PR #982, opened 2022-06-03
Add an embedded web server

pnkfelix:
Every time I want to write a blog post, I have to go look up what way to invoke `python http` or whatever to get a locally visible preview of my work.

Today I decided to try something different.


src/bin/serve.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[path="../main.rs"] mod main;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    main::main()?;
+
+    let footer = format!("{} {}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"));
+
+    warpy::server::run(format!("{}/site", env!("CARGO_MANIFEST_DIR")), [0, 0, 0, 0], footer, Some(8000), false).await?;
```
calebcartwright, line 11: would it be worth allowing an env override of the port? i know i often end up forgetting things running on machine, particularly docker containers, and 8000 is obviously fairly common. think it'd be nice to be able to change this instead of having to first go kill something already on that port

Mark-Simulacrum:
I think we're 90-95% done with migrating to relative urls (thanks to @m-ou-se), which should make file:/// urls work out of the box, I'd hope - I'd rather avoid the extra complexity here if we can, particularly in terms of compile times in CI.

Would file:/// URLs work for you, @pnkfelix?

m-ou-se:
See https://github.com/rust-lang/blog.rust-lang.org/pull/980 and https://github.com/rust-lang/blog.rust-lang.org/pull/981

pietroalbini:
Yeah, the worry I have with this is the compile time increase, as we're now pulling tokio, hyper and other heavyweight dependencies.

I'm not opposed to including the web server, but it must not cause compile time increases. A way to do that would be to setup the repo as a workspace and add the dependencies only to the `server` binary crate.

pnkfelix:
The file approach won’t work for me; I do my development on a distinct box from where I do the preview

I’ll look into the variant that Pietro proposed.

bjorn3:
Maybe use [tiny_http](https://crates.io/crates/tiny_http)? It is a light weight http server with only a few dependencies. Including dependencies (part of which are probably already used here) it takes 10s to compile in debug mode, 15s in release mode and 7s with cg_clif.

Mark-Simulacrum:
> I have to go look up what way to invoke python http or whatever to get a locally visible preview of my work.

FWIW, if this is the problem, I have no objections to us printing "run python3 -m http.server output/ or open file:///..." or some similar message after running the blog, that seems like an easy change to make and a good improvement regardless.

PR #1001, opened 2022-07-08
Fixing Overflow on Long Inline Elements

### LINKED ISSUE #727: Long line unnecessarily increase page width
This line in [Announcing Rust 1.47.0](https://blog.rust-lang.org/2020/10/08/Rust-1.47.html) causes the width of the page to increase and breaks the uniformity of the website

**This issue only occurs in mobile devices**

```
checked_add,checked_sub,checked_mul,checked_neg, 
```

### Suggested Improvement
One should probably break it into seperate lines to wrap it correctly, putting a line break right after `checked_mul,`

EliasPrescott:
closes #727

Here is the post mentioned in the linked issue: https://blog.rust-lang.org/2020/10/08/Rust-1.47.html

Adding ```overflow-wrap: break-word``` to the post content container ensures all child inline elements will wrap properly instead of overflowing.

I was originally going to scope the new property to ```.post a```, but putting it directly on ```.post``` could prevent future issues instead of just fixing the tagged issue.

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -270,6 +270,8 @@ h3, .post h2, header h2 {
 }
 
 .post {
+  overflow-wrap: break-word;
+
   h2 {
     font-size: 2em;
   }


senekor:
Looks like the relevant issue was closed in another way, sorry that nobody took a look until now.

PR #1007, opened 2022-07-19
Add an embedded web server in own workspace

pnkfelix:
Revisiting #982 by putting the new static site serving code into its own Cargo workspace, to avoid increasing build time for the usual blog authors.

Also took @Mark-Simulacrum 's advice from https://github.com/rust-lang/blog.rust-lang.org/pull/982#issuecomment-1146676023 and added a message at the end of the default running binary to remind the user about ways they might get their hands a preview of the generated web site.

(I did explore `tiny_http` as suggested by @bjorn3 in https://github.com/rust-lang/blog.rust-lang.org/pull/982#issuecomment-1146669769 , but it took me a lot more code to get a static site up and running, and I wasn't as confident that my prototype would behave well against "unexpected" urls it received)


README.md
```diff
@@ -26,6 +26,13 @@ Open `site/index.html` in your web browser to view the site.
 > firefox site/index.html
 ```
 
+You can also run a server, if you need to preview your changes on a different machine:
+
+```console
+> cargo run --bin serve
```
pnkfelix, line 32:
```suggestion
> cargo run -p serve
```

pnkfelix:
> Just for my own understanding, is there a reason printing the right python command isn't enough for your use case? It seems ok to do this regardless, ultimately, but I would appreciate the insight.

The `cargo run -p serve` command I'm providing here integrates both the blog generation and the file service. (In my ideal world, it would be generalized to regenerate the html pages as the markdown source is updated; this is admittedly at best a baby step towards such a goal.)

I freely admit that one can get the same composition of actions via `cargo run && python3 -m http.server --directory {ROOT}/site`. So really, this is more my trying to snapshot some work while still thinking about how best to push it further forward in the future.

Mark-Simulacrum:
Hm, sure. I think even with the watch functionality we'd plausibly be just generating the files on disk and letting the webserver (python or something fancier which might auto refresh browser) handle that part of the puzzle.

I'll give it a review code wise for now and merge if it seems good.

Mark-Simulacrum (approved):

PR #1008, opened 2022-07-20
fix footer links to show html instead of xml

pro465:


rylev:
These links are under an RSS header so I think the xml pages are appropriate. Am I missing some context here?

rylev:
I don't use RSS so I'm not 100% sure what's normal here. The link is definitely meant to be xml that is not viewed in the browser but rather added to an RSS client. So, I think this is expected behavior. I'll go ahead and close this as "won't fix", but we can reopen it if someone who uses RSS says this is not how things normally work.

PR #1011, opened 2022-07-23
Add keyword generics initiative announcement post to the internals blog

yoshuawuyts:
This adds the announcement of the keyword generics initiative to the rust internals blog. I've set the release date for Monday the 25th of July, and marked the blog post as `release: false` in case it gets merged before then.

Thanks!

cc/ @rust-lang/initiative-keyword-generics

### DIFF
--- a/posts/inside-rust/2022-07-27-keyword-generics.md
+++ b/posts/inside-rust/2022-07-27-keyword-generics.md
@@ -1,0 +1,406 @@
+---
+layout: post
+title: "Announcing the Keyword Generics Initiative"
+author: Yoshua Wuyts
+team: The Keyword Generics Initiative <https://github.com/rust-lang/keyword-generics-initiative>
+---
+
+We ([Oli], [Niko], and [Yosh]) are excited to announce the start of the [Keyword
+Generics Initiative][kwi], a new initiative [^initiative] under the purview of
+the language team. We're officially just a few weeks old now, and in this post
+we want to briefly share why we've started this initiative, and share some
+insight on what we're about.
+
+[Oli]: https://github.com/oli-obk
+[Niko]: https://github.com/nikomatsakis
+[Yosh]: https://github.com/yoshuawuyts
+[kwi]: https://github.com/rust-lang/keyword-generics-initiative
+
+[^initiative]: Rust governance terminology can sometimes get confusing. An
+"initiative" in Rust parlance is different from a "working group" or "team".
+Initiatives are intentionally limited: they exist to explore, design, and
+implement specific pieces of work - and once that work comes to a close, the
+initiative will wind back down. This is different from, say, the lang team -
+which essentially carries a `'static` lifetime - and whose work does
+not have a clearly defined end.
+
+## A missing kind of generic
+
+One of Rust's defining features is the ability to write functions which are
+_generic_ over their input types. That allows us to write a function once,
+leaving it up to the compiler to generate the right implementations for us.
+
+Rust allows you to be generic over types - it does not allow you to be generic
+over other things that are usually specified by keywords. For example, whether a
+function is async, whether a function can fail or not, whether a function is
+const or not, etc.
+
+The post ["What color is your function"][color] [^color] describes what happens
+when a language introduces async functions, but with no way to be generic over
+them:
+
+> I will take async-await over bare callbacks or futures any day of the week.
+> But we’re lying to ourselves if we think all of our troubles are gone. As soon
+> as you start trying to write higher-order functions, or reuse code, you’re
+> right back to realizing color is still there, bleeding all over your codebase.
+
+This isn't just limited to async though, it applies to all modifier keywords -
+including ones we may define in the future. So we're looking to fill that gap
+by exploring something we call "keyword generics" [^name]: the ability to be
+generic over keywords such as `const` and `async`.
+
+[color]: https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/
+[^color]: R. Nystrom, “What Color is Your Function?,” Feb. 01, 2015.
+https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/
+(accessed Apr. 06, 2022).
+
+[^name]: The longer, more specific name would be: "keyword modifier generics".
+We've tried calling it that, but it's a bit of a mouthful. So we're just
+sticking with "keyword generics" for now, even if the name for this feature may
+end up being called something more specific in the reference and documentation.
+
+To give you a quick taste of what we're working on, this is roughly how we
+imagine you may be able to write a function which is generic over "asyncness"
+in the future:
+
+> Please note that this syntax is entirely made up, just so we can use something
+> in examples. Before we can work on syntax we need to finalize the semantics,
+> and we're not there yet. This means the syntax will likely be subject to
+> change over time.
+
+```rust
+async<A> trait Read {
+    async<A> fn read(&mut self, buf: &mut [u8]) -> Result<usize>;
+    async<A> fn read_to_string(&mut self, buf: &mut String) -> Result<usize> { ... }
+}
+
+/// Read from a reader into a string.
+async<A> fn read_to_string(reader: &mut impl Read * A) -> std::io::Result<String> {
+    let mut string = String::new();
+    reader.read_to_string(&mut string).await?;
+    string
+}
+```
+
+This function introduces a "keyword generic" parameter into the function of `A`.
+You can think of this as a flag which indicates whether the function is being
+compiled in an async context or not. The parameter `A` is forwarded to the `impl
+Read`, making that conditional on "asyncness" as well.
+
+In the function body you can see a `.await` call. Because [the `.await` keyword
+marks cancellation sites][cancel] we unfortunately can't just infer them
+[^cancellation]. Instead we require them to be written for when the code is
+compiled in async mode, but are essentially reduced to a no-op in non-async
+mode.
+
+[cancel]: https://blog.yoshuawuyts.com/async-cancellation-1/
+[^cancellation]: No really, we can't just infer them - and it may not be as
+simple as omitting all `.await` calls either. The Async WG is working through
+the full spectrum of cancellation sites, async drop, and more. But for now we're
+working under the assumption that `.await` will remain relevant going forward.
+And even in the off chance that it isn't, fallibility has similar requirements
+at the call site as async does.
+
+We still have lots of details left to figure out, but we hope this at least
+shows the general *feel* of what we're going for.
+
+## A peek at the past: horrors before const
+
+Rust didn't always have `const fn` as part of the language. A long long long long
+long time ago (2018) we had to write a regular function for runtime computations
+and associated const of generic type logic for compile-time computations. As an
+example, to add the number `1` to a constant provided to you, you had to write
+([playground]):
+
+[playground]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=50e818b79b8af322ed4384d3c33e9773
+
+```rust
+trait Const<T> {
+    const VAL: T;
+}
+
+/// `42` as a "const" (type) generic:
+struct FourtyTwo;
+impl Const<i32> for FourtyTwo {
+    const VAL: i32 = 42;
+}
+
+/// `C` -> `C + 1` operation:
+struct AddOne<C: Const<i32>>(C);
+impl<C: Const<i32>> Const<i32> for AddOne<C> {
+    const VAL: i32 = C::VAL + 1;
+}
+
+AddOne::<FourtyTwo>::VAL
+```
+
+Today this is as easy as writing a `const fn`:
+
+```rust
+const fn add_one(i: i32) -> i32 {
+    i + 1
+}
+
+add_one(42)
+```
+
+The interesting part here is that you can also just call this function in
+runtime code, which means the implementation is shared between both `const`
+(CTFE[^ctfe]) and non-`const` (runtime) contexts.
+
+[^ctfe]: CTFE stands for "Compile Time Function Execution": `const` functions
+can be evaluated during compilation, which is implemented using a Rust
+interpreter (miri).
+
+## Memories of the present: async today
+
+People write duplicate code for async/non-async with the only difference being
+the `async` keyword. A good example of that code today is [`async-std`], which
+duplicates and translates a large part of the stdlib's API surface to be async
+[^async-std]. And because the Async WG has made it an explicit goal to [bring
+async Rust up to par with non-async Rust][async-vision], the issue of code
+duplication is particularly relevant for the Async WG as well. Nobody on the
+Async WG seems particularly keen on proposing we add a second instance of just
+about every API currently in the stdlib.
+
+[`async-std`]: https://docs.rs/async-std/latest/async_std/
+[async-vision]: https://rust-lang.github.io/wg-async/vision/how_it_feels.html
+[^async-std]: Some limitations in `async-std` apply: async Rust is missing async
+`Drop`, async traits, and async closures. So not all APIs could be duplicated.
+Also `async-std` explicitly didn't reimplement any of the collection APIs to be
+async-aware, which means users are subject to the "sandwich problem". The
+purpose of `async-std` was to be a proving ground to test whether creating
+an async mirror of the stdlib would be possible: and it's proven that it is, as
+far as was possible with missing language features.
+
+We're in a similar situation with `async` today as `const` was prior to 2018.
+Duplicating entire interfaces and wrapping them in `block_on` calls is the
+approach taken by e.g. the `mongodb`
+[[async](https://docs.rs/mongodb/latest/mongodb/index.html),
+[non-async](https://docs.rs/mongodb/latest/mongodb/sync/index.html)], `postgres`
+[[async](https://docs.rs/tokio-postgres/latest/tokio_postgres/index.html),
+[non-async](https://docs.rs/postgres/latest/postgres/)], and `reqwest`
+[[async](https://docs.rs/reqwest/latest/reqwest/),
+[non-async](https://docs.rs/reqwest/latest/reqwest/blocking/index.html)] crates:
+```rust
+// Async functionality like this would typically be exposed from a crate "foo":
+async fn bar() -> Bar { 
+    // async implementation goes here
+}
+```
+
+```rust
+// And a sync counterpart would typically be exposed from a crate
+// named "blocking_foo" or a submodule on the original crate as
+// "foo::blocking". This wraps the async code in a `block_on` call:
+fn bar() -> Bar {
+    futures::executor::block_on(foo::bar())
+}
+```
+
+This situation is not ideal. Instead of using the host's synchronous syscalls,
+we're now going through an async runtime to get the same results - something
+which is often not zero-cost. But more importantly, it's rather hard to
+keep both a sync and async API version of the same crate in, err, sync with each
+other. Without automation it's really easy for the two APIs to get out of sync,
+leading to mismatched functionality.
+
+The ecosystem has come up with some solutions to this, perhaps most notably the
+proc-macro based [`maybe-async` crate][maybe-async].  Instead of writing two
+separate copies of `foo`, it generates a sync and async variant for you:
+
+```rust
+#[maybe_async]
+async fn foo() -> Bar { ... }
+```
+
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
The release keyword is a marker for Rust releases, used to produce https://blog.rust-lang.org/releases.json (I believe currently feeding the version number on https://www.rust-lang.org/, but I'm not sure).

So it won't help with accidental merging :)

Mark-Simulacrum:
It also looks like this post is in posts/, not posts/inside-rust/, which is presumably an oversight? It sounds like you intended to put this on Inside Rust.

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+not have a clearly defined end.
+
+## A missing kind of generic
+
+One of Rust's defining features is the ability to write functions which are
+_generic_ over their input types. That allows us to write a function once,
+leaving it up to the compiler to generate the right implementations for us.
+
+But while we're able to be generic over _types_ in a function, we're not able to
+be generic over the qualifier keywords of functions and traits. The post: ["What
```
Mark-Simulacrum, line 35:
```suggestion
be generic over the qualifier keywords of functions and traits. The post ["What
```

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+Today this is as easy as writing a `const fn`:
+
+```rust
+const fn add_one(i: i32) -> i32 {
+    i + 1
+}
+
+add_one(42)
```
Mark-Simulacrum, line 130: I feel like this example is missing something, since to my knowledge `42 + 1` worked for a long time already in const contexts (e.g., `const F: u32 = 42 + 1;`). The only time you might need the complex invocation is if you're using this as a const generic, but there a `const fn` isn't really going to help you either.

Maybe I'm the one missing something though :)

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+async Rust up to par with non-async Rust][async-vision], the issue of code
+duplication is particularly relevant for the Async WG as well. Nobody on the
+Async WG seems particularly keen on proposing we add a second instance of just
+about every API currently in the stdlib.
+
+[`async-std`]: https://docs.rs/async-std/latest/async_std/
+[async-vision]: https://rust-lang.github.io/wg-async/vision/how_it_feels.html
+[^async-std]: Some limitations in `async-std` apply: async Rust is missing async
+`Drop`, async traits, and async closures. So not all APIs could be duplicated.
+Also we explicitly didn't reimplement any of the collection APIs to be
```
Mark-Simulacrum, line 156: I would avoid "we" here, since this is coming as a post from/on behalf of the keywords generics initiative. AFAIK, the initiative isn't intending to actively maintain or anything like that async-std, just make use of the valuable lessons learned in its design and usage -- I think this formulation mixes that up a bit too much. (Particularly later in the paragraph, around being a proving ground and successful proof).

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+// take the `async fn foo` and block the thread until
+// it finishes executing.
+fn foo() -> Bar {
+    futures::executor::block_on(crate_name::foo())
+}
+```
+
+This requires effort on the user's side to find and use the right crates for
+their code. And it requires effort by the crate authors to keep the sync and
+async APIs in sync with each other.
```
Mark-Simulacrum, line 185: The problem I recall is that you get panics if you use a reqwest::blocking API from within an existing tokio runtime (called block_on in async context...) or something like that. Maybe with `futures::...` that's not a problem. Maybe keyword generics help here, though.

Mainly commenting since I think finding and using the right crates seems like.. well, a problem we're not going to solve with keyword generics, so it feels a little odd to comment on it. Crate authors may be able to more easily expose sync and async APIs, though; that seems mostly true. The problem I note with blocking interfaces being wrong to use in an async context (even if not in an async fn), though, doesn't seem like one you can tackle unless you have runtime dispatch (dyn keywords?). And the blocking is presumably mostly a problem only when you're blocking in an async context; if you're intending to block the thread my guess is the overhead is not that high?

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+That's a lot of API surface for just a single method, and __that problem
+multiplies across the entire API surface in the stdlib__. We expect that once we
+start applying "keyword generics" to traits, we will be able to solve the
+sandwich problem. The type `f` would be marked generic over a set of effects,
+and the compiler would choose the right variant during compilation.
+
+## Affecting all effects
+
+Both `const` and `async` share a very similar issue, and we expect that other
+"effects" will face the same issue. "fallibility" particularly on our mind here,
```
Mark-Simulacrum, line 255:
```suggestion
"effects" will face the same issue. "fallibility" is particularly on our mind here,
```

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+made. This is complexity which cannot be avoided, and which needs to exist
+somewhere. Currently in Rust that complexity is thrust entirely on users of
+Rust, making them responsible for choosing whether their code should support
+async Rust or not. But other languages have made diferent choices. For example
+Go doesn't distinguish between "sync" and "async" code, and has a runtime which
+is able to remove that distinction.
+
+The work we're doing would make it so that complexity of choosing between sync
+and async can be moved from user code into the type system, and from there on
+out can be handled by the compiler instead.
```
Mark-Simulacrum, line 280: I feel like this isn't quite right -- presumably, the user still needs to choose what kind of I/O model they're using, since that has major performance (at least) implications, and plausibly correctness ones around timeouts etc.

I think the right claim earlier would be that the "end/final/binary" user must choose, but perhaps the individual developer of a library (e.g., of serde) may not need to choose. That resonates more with me.

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+working on is intentionally limited to built-in keywords only, and wouldn't
+allow you to implement anything like that at all.
+
+What we do share with effect systems is that we're integrating modifier keywords
+more directly into the type system. Modifier keywords like `async` are often
+referred to as "effects", so being able to be conditional over them in
+composable ways effectively gives us an "effect algebra". But that's very
+different from "generalized effect systems" in other languages.
+
+### Are you looking at other keywords beyond `async` and `const`?
```
Mark-Simulacrum, line 296:
```suggestion
### Q: Are you looking at other keywords beyond `async` and `const`?
```

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+[Oli]: https://github.com/oli-obk
+[Niko]: https://github.com/nikomatsakis
+[Yosh]: https://github.com/yoshuawuyts
+[kwi]: https://github.com/rust-lang/keyword-generics-initiative
+
+[^initiative]: Rust governance terminology can sometimes get confusing. An
+"initiative" in Rust parlance is different from a "working group" or "team".
+Initiatives are intentionally limited: they exist to explore, design, and
+implement specific pieces of work - and once that work comes to a close, the
+initiative will wind back down. This is different from, say, the lang team -
```
Mark-Simulacrum, line 24: I think it's probably worth explicitly noting somewhere in this post that initiatives are really intended as explore/design space - there's not actually any guarantees or even strong confidence at this early stage in the process that any of the particulars will be adopted by Rust. Many of our initiatives come to fruition in somewhat different forms or not at all, particularly those relatively early in the process.

Might make for a good FAQ question (something like "Q: Where is the RFC?", perhaps?).

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+language, but not in others. We hope that with this feature we can start to
+systematically close those gaps, leading to a more consistent Rust experience
+for _all_ Rust users.
+
+### Q: What will the backwards compatibility story be like?
+
+Rust has pretty strict backwards-compatibility guarantees, and any feature we
+implement needs to adhere to this. Luckily we have some wiggle room because of
+the edition mechanism, but our goal is to shoot for maximal backwards compat. We
+have some ideas of how we're goingt to make this work though, and we're
```
fee1-dead, line 322:
```suggestion
have some ideas of how we're going to make this work though, and we're
```

Saw this on the hackmd, then saw this PR :)

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,344 @@
...
+
+In this post we've introduced the new keyword generics initiatve, explained why
+it exists, and shown a brief example of what it might look like in the future.
+
+The initiative is active on the Rust Zulip under
+[`t-lang/keyword-generics`][zulip] - if this seems interesting to you, please
+pop by!
+
+_Thanks to everyone who's helped review this post, but in particular:
+[fee1-dead][fd], [Daniel Henry-Mantilla], and [Ryan Levick][rylev]_
```
fee1-dead, line 339:
```suggestion
[fee1-dead][fd], [Daniel Henry-Mantilla][dhm], and [Ryan Levick][rylev]_
```

yoshuawuyts: Ohh, good catch. Thank you!

posts/2022-07-25-keyword-generics.md
yoshuawuyts: > I think the right claim earlier would be that the "end/final/binary" user must choose, but perhaps the individual developer of a library (e.g., of serde) may not need to choose. That resonates more with me.

Yeah, that's exactly what we're going for as well. I agree that the current phrasing doesn't quite capture that. Let me find a way to update that.

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,343 @@
...
+which essentially carries a `'static` lifetime - and whose work (ominously) does
+not have a clearly defined end.
+
+## A missing kind of generic
+
+One of Rust's defining features is the ability to write functions which are
+_generic_ over their input types. That allows us to write a function once,
+leaving it up to the compiler to generate the right implementations for us.
+
+But while we're able to be generic over _types_ in a function, we're not able to
```
rylev, line 33: This paragraph seems to jump from "being able to be generic over types" to "being generic over asyncness" which is a bit confusing. It seems like you're describing what this feature *isn't* then an *example* of what this feature i. Yet you never attempt to define what the feature itself is. 

It would be nice to say something to the effect of "Rust allows you to be generic over types - it does not allow you to be generic over other things that are usually specified by keywords in Rust. For example, whether a function is async, whether a function can fail or not, whether a function is const or not, etc."

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,343 @@
...
+[^color]: R. Nystrom, “What Color is Your Function?,” Feb. 01, 2015.
+https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/
+(accessed Apr. 06, 2022).
+
+[^name]: The longer, more specific name would be: "keyword modifier generics".
+We've tried calling it that, but it's a bit of a mouthful. So we're just
+sticking with "keyword generics" for now, even if the name for this feature may
+end up being called something more specific in the reference and documentation.
+
+To give you a quick taste of what we're working on, this is how we imagine you
```
rylev, line 51: I would explicitly note that this syntax is just for demonstration is not really the point. The point is what it allows one to express.

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,343 @@
...
+[^cancellation]. Instead we require them to be written for when the code is
+compiled in async mode, but are essentially reduced to a no-op in non-async
+mode.
+
+[cancel]: https://blog.yoshuawuyts.com/async-cancellation-1/
+[^cancellation]: No really, we can't just infer them - and it may not be as
+simple as omitting all `.await` calls either. The Async WG is working through
+the full spectrum of cancellation sites, async drop, and more. But for now we're
+working under the assumption that `.await` will remain relevant going forward.
+And even in the off chance that it isn't, fallibility has similar requirements
```
rylev, line 85: You mention fallibility but you did not mention before that this is something that this feature can make generic.

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,343 @@
...
+at the call site as async does.
+
+We still have lots of details left to figure out, but we hope this at least
+shows the general *feel* of what we're going for. We want to make it
+significantly easier to write code which works in both sync and non-async Rust.
```
rylev, line 90: This line makes this effort feel very specific to asyncness but it's my understanding this is meant to be much more general.

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,343 @@
...
+Duplicating entire interfaces and wrapping them in inefficient `block_on` calls
+is the approach taken by e.g. the `mongodb`
+[[async](https://docs.rs/mongodb/latest/mongodb/index.html),
+[non-async](https://docs.rs/mongodb/latest/mongodb/sync/index.html)], `postgres`
+[[async](https://docs.rs/tokio-postgres/latest/tokio_postgres/index.html),
+[non-async](https://docs.rs/postgres/latest/postgres/)], and `reqwest`
+[[async](https://docs.rs/reqwest/latest/reqwest/),
+[non-async](https://docs.rs/reqwest/latest/reqwest/blocking/index.html)] crates:
+```rust
+// "crate_name"
```
rylev, line 171: These comments are confusing. I suppose they're meant to indicate the modules these would live in, but that is very much not clear.

posts/2022-07-25-keyword-generics.md
```diff
@@ -0,0 +1,343 @@
+---
+layout: post
+title: "Announcing the Keyword Generics Initiative"
+author: Yoshua Wuyts, on behalf of the Keyword Generics Initiative
+release: false
```
rylev, line 5:
```suggestion
```

false is the default

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Good call; done!

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Yep, that makes sense. Changed!

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Agreed; added more clarifying comments and split it into two code blocks.

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Changed the focus of this section. I agree it's probably not the right thing to highlight.

posts/2022-07-25-keyword-generics.md
oli-obk: you can always inline a `const fn` at its call site. After a certain complexity that is not useful though. Showing a complex example here is not very helpful to the reader either. So I chose to write a simple `add_one` function, but honestly it could be anything that doesn't require iteration or branching (as that wasn't supported before `const fn`)

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Good call; thank you!

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Good call!

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Incorporated your earlier feedback; now it does ^^

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Yep, true. Removed!

yoshuawuyts:
Updated with all feedback! - Also changed to date to tomorrow (Tuesday) to publish as it's getting quite late today.

posts/inside-rust/2022-07-26-keyword-generics.md
```diff
@@ -0,0 +1,405 @@
...
+which essentially carries a `'static` lifetime - and whose work (ominously) does
+not have a clearly defined end.
+
+## A missing kind of generic
+
+One of Rust's defining features is the ability to write functions which are
+_generic_ over their input types. That allows us to write a function once,
+leaving it up to the compiler to generate the right implementations for us.
+
+Rust allows you to be generic over types - it does not allow you to be generic over other things that are usually specified by keywords in Rust. For example, whether a function is async, whether a function can fail or not, whether a function is const or not, etc.
```
rylev, line 33:
```suggestion
Rust allows you to be generic over types - it does not allow you to be generic over other things that are usually specified by keywords. For example, whether a function is async, whether a function can fail or not, whether a function is const or not, etc.
```

posts/inside-rust/2022-07-26-keyword-generics.md
```diff
@@ -0,0 +1,405 @@
...
+[^name]: The longer, more specific name would be: "keyword modifier generics".
+We've tried calling it that, but it's a bit of a mouthful. So we're just
+sticking with "keyword generics" for now, even if the name for this feature may
+end up being called something more specific in the reference and documentation.
+
+To give you a quick taste of what we're working on, this is roughly how we
+imagine you may be able to write a function which is generic over "asyncness"
+in the future:
+
+_Please note that this syntax is entirely made up, just so we can use something
```
rylev, line 63: Would a footnote or `>` block look better?

posts/inside-rust/2022-07-26-keyword-generics.md
```diff
@@ -0,0 +1,405 @@
...
+    Some(T),
+    None,
+}
+
+impl<T> Option<T> {
+    fn map<J>(self, f: impl FnOnce(T) -> J) -> Option<J> { ... }
+}
+```
+
+```rust
```
rylev, line 248: Why two blocks?

rylev (approved):

posts/inside-rust/2022-07-26-keyword-generics.md
```diff
@@ -0,0 +1,405 @@
...
+but it isn't the only effect. In order for the language to feel consistent we
+need consistent solutions.
+
+## FAQ
+
+### Q: Is there an RFC available to read?
+
+Rust initiatives are intended for _exploration_. The announcement of the Keyword
+Generics Initiative marks the _start_ of the exploration process. Part of
+exploring is not knowing what the oucomes will be. Right now we're in the
```
Mark-Simulacrum, line 284:
```suggestion
exploring is not knowing what the outcomes will be. Right now we're in the
```

posts/inside-rust/2022-07-26-keyword-generics.md
yoshuawuyts: Changed it to use `>`!

oli-obk (approved):

posts/2022-07-25-keyword-generics.md
yoshuawuyts: Can we mark this as resolved then?

oli-obk:
You need to mark it as `release: false`, that flag is about whether the post is a rust version release, not whether to publish the post.

I can press the merge button, I don't know if I'm allowed to XD who makes the final decision on blog posts?

yoshuawuyts:
> You need to mark it as release: false, that flag is about whether the post is a rust version release, not whether to publish the post.

I don't believe such a flag is needed for the inside Rust blog? Here's the front matter from a previous post, which matches most other posts too:

```yaml
---
layout: post
title: Rust Compiler June 2022 Steering Cycle
author: Felix Klock
description: "The compiler team's June 2022 steering cycle"
team: The Compiler Team <https://www.rust-lang.org/governance/teams/compiler>
---
```

> I can press the merge button, I don't know if I'm allowed to XD who makes the final decision on blog posts?

My understanding from talking about this with the lang team is that any Rust team can publish to the inside rust blog, as long as it's relevant to the Rust project and we follow proper review etiquette - like we've done. I believe this is the reason why when you're added to a Rust initiative, you get commit access on the blog as well. Ideally this would all be documented in a `CONTRIBUTING.md` file on the repo, but there is none currently.

PR #1023, opened 2022-08-12
Fix the note link for rust 1.63

Dengjianping:
The blog for rust 1.63 is pointing to the wrong note link(1.62).


rylev (approved):
Thanks!

PR #1025, opened 2022-08-17
inside-rust: fix links in diag effort post

davidtwco:
Links that are clickable are better than links that aren't. QED.


rylev (approved):
🎉

PR #1041, opened 2022-10-21
Add missing generic in "perfect derive" paragraph

kupiakos:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9662ba9557d61e0507477e386a81e260

```rust
use std::rc::Rc;

struct T(u32);

// This does compile:
// #[derive(Clone)]
// struct MyStruct(Rc<T>);

// This does not compile, needing perfect derives:
#[derive(Clone)]
struct MyStruct<T>(Rc<T>);

fn main() {
    let x = Rc::new(T(10));
    MyStruct(x).clone();
}
```


rylev (approved):
Thanks!

PR #1043, opened 2022-10-28
Announcing Rust 1.65.0

cuviper:


posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,177 @@
...
+channel (`rustup default nightly`). Please
+[report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you
+might come across!
+
+## What's in 1.65.0 stable
+
+### Generic associated types (GATs)
+
+- [Stabilize generic associated types (GATs)](https://github.com/rust-lang/rust/pull/96709/)
+- Previous blog: <https://blog.rust-lang.org/2021/08/03/GATs-stabilization-push.html>
```
joshtriplett, line 47:
```suggestion
- [More details in this recent blog post on Inside Rust](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html)
- Previous blog: <https://blog.rust-lang.org/2021/08/03/GATs-stabilization-push.html>
```

posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,177 @@
...
+
+(TODO: Jack Huey offered to write here...)
+
+### `let`-`else` statements
+
+This introduces a new type of `let` statement with a refutable pattern and a
+diverging `else` block that executes when that pattern doesn't match.
+
+    let PATTERN: TYPE = EXPRESSION else DIVERGING_BLOCK;
```
joshtriplett, line 56:
```suggestion
    let PATTERN: TYPE = EXPRESSION else {
        DIVERGING_CODE;
    };
```

posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,177 @@
...
+        panic!("Can't parse integer: '{count_str}'");
+    };
+    (count, item)
+}
+assert_eq!(get_count_item("3 chairs"), (3, "chairs"));
+```
+
+The scope of name bindings is the main thing that makes this different from
+`match` or `if let`-`else` expressions. You could previously approximate these
+patterns with an unfortunate bit of repeation to an outer `let`:
```
joshtriplett, line 82:
```suggestion
patterns with an unfortunate bit of repetition and an outer `let`:
```

posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,177 @@
...
+- As the final step of the [RLS
+  deprecation](https://blog.rust-lang.org/2022/07/01/RLS-deprecation.html),
+  this release has replaced RLS with a small LSP server showing a deprecation
+  warning, advising users to migrate to `rust-analyzer`.
+
+### Other changes
+
+There are other changes in the Rust 1.65 release, including:
+
+- TODO: anything small to call out?
```
joshtriplett, line 166: I think enabling MIR optimizations should get called out.

posts/2022-11-03-Rust-1.65.0.md
cuviper: Oh, I only put these links for my own reference until we write something here for real. (The TODO just below.)

joshtriplett: Feel free to replace this as well, just wanted to make sure the link was handy. I do think the most recent blog post would be worth linking for more details.

jackh726: Here's a draft of this section (sorry it took so long!):

Lifetime, type, and const generics can now be defined on associated types, like so:

```rust
trait Foo {
    type Bar<'x>;
}
```

It's hard to put into few words just how useful these can be, so here are a
few example traits, to get a sense of their power:

```rust
/// An `Iterator`-like trait that can borrow from `Self`
trait LendingIterator {
    type Item<'a> where Self: 'a;

    fn next<'a>(&'a mut self) -> Option<Self::Item<'a>>;
}

/// Can be implemented over smart pointers, like `Rc` or `Arc`,
/// in order to allow being generic over the pointer type
trait PointerFamily {
    type Pointer<T>: Deref<Target = T>;

    fn new<T>(value: T) -> Self::Pointer<T>;
}

/// Allows borrowing an array of items. Useful for
/// `NdArray`-like types that don't necesarily store
/// data contiguously.
trait BorrowArray<T> {
    type Array<'x, const N: usize> where Self: 'x;

    fn borrow_array<'a, const N: usize>(&'a self) -> Self::Array<'a, N>;
}
```

As you can see, GATs are quite versatile and enable a number
of patterns that are not currently able to be written. For more
information, check out the post announcing the
[push for stabilization](https://blog.rust-lang.org/2021/08/03/GATs-stabilization-push.html) 
published last year or the
[stabilization announcement post](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html) 
published last week. The former goes into a bit more depth
of a couple of the examples above, while the latter talks a
bit about some of the known limitations of this stabilization.

More in depth reading can be found in the associated types
section of the [nightly reference](https://doc.rust-lang.org/nightly/reference/items/associated-items.html#associated-types)
or the [original RFC](https://rust-lang.github.io/rfcs/1598-generic_associated_types.html) (which was initially opened over 6.5 years ago!).

posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+### Splitting Linux debuginfo
+
+Back in Rust 1.51, the compiler team added support for [split debug
+information](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#splitting-debug-information)
+on macOS with `-Csplit-debuginfo=unpacked`, and now this option is stable for
+use on Linux as well, producing separate DWARF `.dwo` object files. This
+separation may improve build times, especially for the linker as it will simply
+have less data to process.
```
CryZe, line 130: This sentence is pretty confusing to me, and possibly to other readers as well. Why does it have less data to process now, isn't it just split and therefore the same, but in different files? I'm not sure how that makes anything be processed faster.

cuviper: It's possible that I'm misunderstanding or misrepresenting it too -- @davidtwco can you help clarify expectations here?

davidtwco: Something like this, maybe?

> Back in Rust 1.51, the compiler team stabilized support for [split debug information](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#splitting-debug-information) on macOS, and now this option is stable for use on Linux as well. 
>
> `-Csplit-debuginfo=unpacked` will split debuginfo out into multiple `.dwo` DWARF object files and `-Csplit-debuginfo=packed` will produce a single `.dwp` DWARF package alongside your output binary with all the debuginfo packaged together.
>
> Split DWARF lets the linker avoid processing the debuginfo (because it isn't in the object files being linked anymore), which can speed up link times!
```

Comment (cuviper) [reply posts/2022-11-03-Rust-1.65.0.md]: Thanks!

File: posts/2022-11-03-Rust-1.65.0.md (hunk 1 above)
Comment (cuviper) [reply posts/2022-11-03-Rust-1.65.0.md]: Added, thanks!

File: posts/2022-11-03-Rust-1.65.0.md [hunk 6]
```diff
@@ -0,0 +1,234 @@
...
+/// Can be implemented over smart pointers, like `Rc` or `Arc`,
+/// in order to allow being generic over the pointer type
+trait PointerFamily {
+    type Pointer<T>: Deref<Target = T>;
+
+    fn new<T>(value: T) -> Self::Pointer<T>;
+}
+
+/// Allows borrowing an array of items. Useful for
+/// `NdArray`-like types that don't necesarily store
```
golddranks, line 74:
```suggestion
/// `NdArray`-like types that don't necessarily store
```

posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,234 @@
...
+  this release has replaced RLS with a small LSP server showing a deprecation
+  warning, advising users to migrate to `rust-analyzer`.
+
+### Other changes
+
+There are other changes in the Rust 1.65 release, including:
+
+- MIR inlining is now enabled for optimized compilations. This provides a 3-10%
+  improvement in compiletimes for real world crates.
+- Improved the performance of Cargo build scheduling by sorting the queue of pending jobs.
```
inquisitivecrystal, line 223:
```suggestion
- When scheduling builds, Cargo now sorts the queue of pending jobs to improve performance.
```
This flows a bit better, though it's not a huge deal.

posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,234 @@
...
+    };
+```
+
+### `break` from labeled blocks
+
+Plain block expressions can now be labeled as a `break` target, terminating
+that block early. This may sound a little like a `goto` statement, but it's not
+an arbitrary jump, only from within a block to its end. This was already
+possible with `loop` blocks, and you may have seen people write loops that
+always execute just once, only to get a labeled `break`.
```
inquisitivecrystal, line 150:
```suggestion
always execute only once, just to get a labeled `break`.
```

posts/2022-11-03-Rust-1.65.0.md
```diff
@@ -0,0 +1,234 @@
...
+
+Now there is a direct language feature for that! Labeled `break` may also
```
inquisitivecrystal, line 152:
```suggestion
Now there's a specific language feature for that! Labeled `break` may also
```

PR #1053, opened 2022-12-03
Add blog post announcing Rust 1.66.0

tmandry:


posts/2022-12-15-Rust-1.66.0.md
```diff
@@ -0,0 +1,171 @@
...
+assert_eq!(42, Bar::C as u8);
+```
+
+You could even add fields to enums with `#[repr(Int)]`, and they would be laid out in a predictable way. Previously, however, you could not use these features together. That meant that making `Foo::C`'s discriminant equal to 42 as above would be harder to achieve. You would need to add 41 hidden variants in between as a workaround with implicit discriminants!
+
+Starting in Rust 1.66.0, the above example compiles, allowing you to use explicit discriminants on any enum with a `#[repr(Int)]` attribute.
+
+### `core::hint::black_box`
+
+When benchmarking or examining the machine code produced by a compiler, it's often useful to prevent optimizations from ocurring in certain places. In the following example, the function `push_cap` executes `Vec::push` 4 times in a loop:
```
ehuss, line 71:
```suggestion
When benchmarking or examining the machine code produced by a compiler, it's often useful to prevent optimizations from occurring in certain places. In the following example, the function `push_cap` executes `Vec::push` 4 times in a loop:
```

posts/2022-12-15-Rust-1.66.0.md
```diff
@@ -0,0 +1,171 @@
...
+```rust
+#[repr(u8)]
+enum Bar {
+    A,
+    B,
+    C = 42,
+}
+```
+
+Here the `Foo` enum is guaranteed to have the same layout as `u8`. Each variant will use either the specified discriminant value or default to starting with 0.
```
Veykril, line 57:
```suggestion
Here the `Bar` enum is guaranteed to have the same layout as `u8`. Each variant will use either the specified discriminant value or default to starting with 0.
```
This is referring to `Bar`, not `Foo` given the examples right?

posts/2022-12-15-Rust-1.66.0.md
```diff
@@ -0,0 +1,171 @@
...
+Notice that the generated code is not at all concerned with the possibility of allocations introduced by the `push` call. This is because the compiler is still using the fact that we called `Vec::with_capacity(4)` in the `bench_push` function. You can play around with the placement of `black_box`, or try using it in multiple places, to see its effects on compiler optimizations.
+
+### cargo remove
+
+In Rust 1.62.0 we introduced `cargo add`, a command line utility to add dependencies to your project. Now you can use `cargo remove` to remove dependencies.
+
+### Stabilized APIs
+
+- [`proc_macro::Span::source_text`](https://doc.rust-lang.org/stable/proc_macro/struct.Span.html#method.source_text)
+- [`uX::{checked_add_signed, overflowing_add_signed, saturating_add_signed, wrapping_add_signed}`](https://doc.rust-lang.org/stable/std/primitive.u8.html#method.checked_add_signed)
```
therealprof, line 143: I do find the naming a bit unfortunate because there is a crate called [`uX`](https://crates.io/crates/ux) providing non-power-of-2 integers and someone might think that Rust has gotten that long thought after ability now... it certainly got me excited for a minute.

Techie-Pi, line 143: What about using ``*`` instead of ``x`` as seen on the [Rust 1.64.0 announcement blog post](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#stabilized-apis)?

Techie-Pi (changes requested):

posts/2022-12-15-Rust-1.66.0.md
```diff
@@ -0,0 +1,171 @@
...
+- [`core::hint::black_box`](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html)
+- [`Duration::try_from_secs_{f32,f64}`](https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.try_from_secs_f32)
+- [`Option::unzip`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.unzip)
+- [`std::os::fd`](https://doc.rust-lang.org/stable/std/os/fd/index.html)
+
+### Other changes
+
+There are other changes in the Rust 1.66 release, including:
+
+- You can now use `..X` and `..=X` ranges in patterns.
```
pietroalbini, line 159:
```suggestion
- You can now use `..=X` ranges in patterns.
```

Not actually stabilized https://github.com/rust-lang/rust/pull/105658.

pietroalbini (approved):

PR #1055, opened 2022-12-14
Announce an impending update the Android NDK target

chriswailes:
The language is a little rough.  Please let me know if there is anything I can do to improve the announcement.

### DIFF
--- a/posts/2023-01-09-android-ndk-update-r25.md
+++ b/posts/2023-01-09-android-ndk-update-r25.md
@@ -1,6 +1,6 @@
 ---
 layout: post
-title: "Updating the Android NDK in Rust <version>"
+title: "Updating the Android NDK in Rust 1.68"
 author: Android Platform Team
 description: "Modernizing Android support in Rust"
 ---
--- a/posts/2023-01-09-android-ndk-update-r25.md
+++ b/posts/2023-01-09-android-ndk-update-r25.md
@@ -1,0 +1,28 @@
+---
+layout: post
+title: "Updating the Android NDK in Rust <version>"
+author: Android Platform Team
+description: "Modernizing Android support in Rust"
+---
+
+We are pleased to announce that Android platform support in Rust will be
+modernized in Rust 1.66 as we update the target NDK from r17 to r25.  As a
+consequence the minimum supported API level will increase from 15 (Ice Cream
+Sandwich) to 19 (KitKat).
+
+In NDK r23 Android switched to using LLVM's `libunwind` for all architectures.
+This meant that
+1. If a project were to target NDK r23 or newer with previous versions of Rust
+   [a workaround](https://github.com/rust-lang/rust/pull/85806#issuecomment-1096266946)
+   would be required to redirect attempts to link against `libgcc` to instead
+   link against `libunwind`.  Following this update this workaround will no
+   longer be necessary.
+2. If a project uses NDK r22 or older it will need to be updated to use [r23 or
+   newer](https://developer.android.com/ndk/downloads).  Information about the
+   layout of the NDK's toolchain can be found
+   [here](https://developer.android.com/ndk/guides/other_build_systems).
+
+Going forward the Android platform will target the most recent LTS NDK, allowing
+Rust developers to access platform features sooner.  These updates should occur
+yearly and will be announced in release notes.
+
--- a/posts/2023-01-09-android-ndk-update-r25.md
+++ b/posts/2023-01-09-android-ndk-update-r25.md
@@ -6,7 +6,7 @@ description: "Modernizing Android support in Rust"
 ---
 
 We are pleased to announce that Android platform support in Rust will be
-modernized in Rust 1.66 as we update the target NDK from r17 to r25.  As a
+modernized in Rust 1.68 as we update the target NDK from r17 to r25.  As a
 consequence the minimum supported API level will increase from 15 (Ice Cream
 Sandwich) to 19 (KitKat).
 


posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,22 @@
...
+
+In NDK r23 Android switched to using LLVMs `libunwind` for all architectures.
+This meant that if a project were to target NDK r23 or newer with previous
+versions of Rust a workaround would be required to redirect attempts to link
+against `libgcc` to instead link against `libunwind`.  Following this update
+this workaround will no longer be necessary.  Projects using NDKs between r17
+and r22 should not be affected.
+
+Going forward the Android platform will target the most recent LTS NDK, allowing
+Rust developers to access platform featuers sooner.
```
jyn514, line 21:
```suggestion
Rust developers to access platform features sooner.
```

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,22 @@
...
+We are pleased to announce that Android platform support in Rust will be
+modernized in Rust <version> as we update the target NDK from r17 to r25.  As a
+consequence the minimum supported API level will increase from 15 (Jelly Bean)
+to 19 (KitKat).
+
+In NDK r23 Android switched to using LLVMs `libunwind` for all architectures.
+This meant that if a project were to target NDK r23 or newer with previous
+versions of Rust a workaround would be required to redirect attempts to link
+against `libgcc` to instead link against `libunwind`.  Following this update
+this workaround will no longer be necessary.  Projects using NDKs between r17
```
jyn514, line 17: this should probably mention that the workaround will cause issues; I think people on https://github.com/rust-lang/rust/issues/103673 had more detailed suggestions.

posts/2022-12-14-android-ndk-update.md
chriswailes: Thanks!  Fixed in later version.

posts/2022-12-14-android-ndk-update.md
chriswailes: I've added language explaining the need to update for projects targeting an NDK between r17 and r23.

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,24 @@
...
+
+We are pleased to announce that Android platform support in Rust will be
+modernized in Rust <version> as we update the target NDK from r17 to r25.  As a
+consequence the minimum supported API level will increase from 15 (Jelly Bean)
+to 19 (KitKat).
+
+In NDK r23 Android switched to using LLVMs `libunwind` for all architectures.
+This meant that
+1. If a project were to target NDK r23 or newer with previous versions of Rust
+   a workaround would be required to redirect attempts to link against `libgcc`
```
alex-pinkus, line 16: It might be useful to link to some documentation explaining the workaround in more detail, e.g. https://github.com/rust-lang/rust/pull/85806#issuecomment-1096266946. This does double-duty of 1. making sure it's clear what workaround the blog post is referring to and 2. making it easier for anyone who reads this and _wants_ to apply the workaround to find it.

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,24 @@
...
+   to instead link against `libunwind`.  Following this update this workaround
+   will no longer be necessary.
+2. If a project targets an NDK between r17 and r23 it will need to be updated to
```
alex-pinkus, line 19: Technically, I believe a project can target a version as low as `r9b` today. Maybe earlier, but that's the oldest one I recall testing. Might be simplest to say "If a project targets NDK version r22 or lower" instead of specifying a lower bound.

jyn514 (approved):
LGTM, it needs the version number before being merged though

posts/2022-12-14-android-ndk-update.md
chriswailes: Done

posts/2022-12-14-android-ndk-update.md
chriswailes: Done

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: post
+title: "Updating the Android NDK in Rust <version>"
+author: Android Platform Team
+description: "Modernizing Android support in Rust"
+---
+
+We are pleased to announce that Android platform support in Rust will be
+modernized in Rust <version> as we update the target NDK from r17 to r25.  As a
+consequence the minimum supported API level will increase from 15 (Jelly Bean)
```
alex, line 10: I believe 15 is Ice Cream Sandwich.

stephenhines: It it ICS. https://en.wikipedia.org/wiki/Android_version_history is actually the easiest resource for this.

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,25 @@
...
+to 19 (KitKat).
+
+In NDK r23 Android switched to using LLVMs `libunwind` for all architectures.
```
stephenhines, line 13: This should have an apostrophe like "LLVM's" or you could drop the "s" entirely.

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,25 @@
...
+This meant that
+1. If a project were to target NDK r23 or newer with previous versions of Rust
+   [a workaround](https://github.com/rust-lang/rust/pull/85806#issuecomment-1096266946)
+   would be required to redirect attempts to link against `libgcc` to instead
+   link against `libunwind`.  Following this update this workaround will no
+   longer be necessary.
+2. If a project targets NDK r22 or older it will need to be updated to target
```
stephenhines, line 20: Minor nitpick when I read this all the way through: Would "uses" and "use" be better than "targets"/"target" here? I generally feel like target is used more for API level (like the 19/KitKat mentioned above) or architecture (like aarch64).

posts/2022-12-14-android-ndk-update.md
chriswailes: Done

posts/2022-12-14-android-ndk-update.md
chriswailes: Done

posts/2022-12-14-android-ndk-update.md
chriswailes: Done

rylev:
@chriswailes when would you like this to go live? Could you make sure to update the post's file name to have the right date. Also, don't forget to update the Rust version where the change is supposed to land.

chriswailes:
Is the announcement date up to me?  If so, I'll change the date to Monday.

What version would the NDK update land in if the PR is accepted next week?  Would that be 1.68?

jyn514:
1.66 stable just released, so you have a 6 week window to make it into 1.68.

rylev:
I think we should wait for the PR to land first, no? It would be a shame for the announcement to happen only for the PR to be delayed.

Mark-Simulacrum:
I'm largely happy with merging this before the PR; the worst case is that folks have a little more time. But in practice we've already landed this once I believe, so I don't anticipate it being difficult to do again.

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,25 @@
...
+Sandwich) to 19 (KitKat).
+
+In NDK r23 Android switched to using LLVM's `libunwind` for all architectures.
+This meant that
+1. If a project were to target NDK r23 or newer with previous versions of Rust
+   [a workaround](https://github.com/rust-lang/rust/pull/85806#issuecomment-1096266946)
+   would be required to redirect attempts to link against `libgcc` to instead
+   link against `libunwind`.  Following this update this workaround will no
+   longer be necessary.
+2. If a project uses NDK r22 or older it will need to be updated to use r23 or
```
Mark-Simulacrum, line 20: Is there anything we can link to that's an explainer for how to do this? Do we expect anyone targeting Android to know what they're currently using, or would it be possible to give some guidance (e.g., "run this command to check what you're targeting today").

posts/2022-12-14-android-ndk-update.md
```diff
@@ -0,0 +1,25 @@
...
+   newer.
+
+Going forward the Android platform will target the most recent LTS NDK, allowing
```
Mark-Simulacrum, line 23: Have we discussed how quickly this bump will happen? I think at minimum we probably bump when it releases on nightly, which gives folks 6-12 weeks to update if needed. It seems reasonable to guarantee 10-12 weeks, but no strong feelings there.

I do think we should add a sentence here like "this will be mentioned in release notes", and if possible provide guidance on how often it happens (yearly?).

posts/2022-12-14-android-ndk-update.md
chriswailes: I believe users should know how to do this.  They would have had to have installed the NDK (either through Studio or by downloading it manually) and then passed that path into their Rust build system to use it at all, so presumably they've already accomplished these steps at least once.

And if they ended up needing to apply this workaround they should be acutely aware of what version of the NDK they are using.

Mark-Simulacrum: At least once a few years back (e.g., for an app that is only partially maintained) seems pretty likely to at the very least be out of cache.

Is it hard to link to *something* here? Even the page for NDK downloads or Studio is probably of some help to users.

posts/2022-12-14-android-ndk-update.md
chriswailes: I'd like to set up some kind of automation somewhere that tests nightly against NDK betas.  I'm not sure if that would give us enough confidence to bump the NDK version the Rust release immediately after the NDK release, but it would at least allow us to file potential issues during the beta process.

If that strategy wouldn't work then I think adding the new NDK to nightly and let it go through the normal process is reasonable.

I believe NDK LTS releases will occur at most once a year going forward.

Mark-Simulacrum: > confidence to bump the NDK version the Rust release immediately after the NDK release

I don't think we would ever want to be backporting the NDK bump, so unless you mean that we would bump to a not yet(?) released version, I don't see this happening regardless of how much testing it received.

We do bump to not-yet-released versions for LLVM, so that might be feasible, but I imagine that the community pain with LLVM (which the vast majority of Rust users aren't building, and if they are, we're not forcing them to immediately jump to a different branch on our switch) is significantly lower than with NDKs, at least from what I understand.

chriswailes: My idea was to have a special CI configuration that builds and tests nightly with the Android NDK beta.  The goal being to catch any issues before the actual nightly builds are moved to newly released versions of the NDK.

I'm not suggesting that Rust releases be built against beta NDKs.

chriswailes:
Are there any more edits I should make to the body of the test or should I update the date to next Monday?

Mark-Simulacrum:
https://github.com/rust-lang/blog.rust-lang.org/pull/1055#discussion_r1052707192 I think still needs to be resolved.

posts/2022-12-14-android-ndk-update.md
chriswailes: I've added a link to a download page and the NDK's page on integrating the toolchain with non-CMake build systems.  I can't think of any other helpful links off the top of my head but I'd be happy to add any more that we come up with.

posts/2022-12-14-android-ndk-update.md
Mark-Simulacrum: That doesn't seem unreasonable, though to some extent depends on how often that builder would break and whether it would take a significant chunk of time.

I think part of the confusion might be that to be honest, bumping the version to the new NDK version directly after that releases on nightly Rust isn't something I think we need *confidence* for -- we either have policy that says we do that or not, but it's not really based on knowing whether such a bump will land. It's obviously gated on it landing (and bors approving that), but not more than that.

chriswailes:
> [#1055 (comment)](https://github.com/rust-lang/blog.rust-lang.org/pull/1055#discussion_r1052707192) I think still needs to be resolved.

Yeah, sorry.  I'm still not used to GitHub's threading and saw that after I made the last post.

posts/2022-12-14-android-ndk-update.md
chriswailes: With that setup do you think it would be possible to skip the nightly branch and move the NDK updates directly into the beta branch?

Mark-Simulacrum: No, we don't backport or directly land changes in the beta branch for anything except to fix regressions - I don't think this would qualify.

posts/2023-01-09-android-ndk-update-r25.md
```diff
@@ -0,0 +1,28 @@
+---
+layout: post
+title: "Updating the Android NDK in Rust <version>"
+author: Android Platform Team
+description: "Modernizing Android support in Rust"
+---
+
+We are pleased to announce that Android platform support in Rust will be
+modernized in Rust 1.66 as we update the target NDK from r17 to r25.  As a
```
Mark-Simulacrum, line 9:
```suggestion
modernized in Rust 1.68 as we update the target NDK from r17 to r25.  As a
```

Presuming we expect to land this prior to January 20th, which is the cut-off for 1.68; otherwise 1.69 is the right thing here. Not sure if we will hit the 20th; I think an alternative is writing 1.68 or 1.69 here.

posts/2022-12-14-android-ndk-update.md
Mark-Simulacrum: Skimming that page it looks great to me.

Mark-Simulacrum (approved):

posts/2023-01-09-android-ndk-update-r25.md
```diff
@@ -0,0 +1,28 @@
+---
+layout: post
+title: "Updating the Android NDK in Rust <version>"
```
Mark-Simulacrum, line 3:
```suggestion
title: "Updating the Android NDK in Rust 1.68"
```

PR #1071, opened 2023-01-31
Create postmortem for DNS outage

jdno:
When we deployed changes to the Content Delivery Network for static.crates.io last week, we accidentally caused a brief DNS outage for its record. This post summarizes the incident and the lessons that we have learned from it.

cc @rust-lang/infra


rylev:
@jdno when do you want to merge this?

jdno:
I've bumped the date of the post to tomorrow, which would be two weeks after the incident. That feels like a nice day to release this.

rylev (approved):

PR #1085, opened 2023-02-23
Add missing ?const

tyilo:


yoshuawuyts:
Thanks for filing this PR! - I get that it's confusing, because what you wrote is indeed what we think the syntax should be! But by this point in the post, we haven't yet suggested changing the syntax of `const fn` to `?const fn`, despite having mentioned that `const fn` already does mean "maybe-const". Confusing!

We only suggest changing `const fn` to `?const fn` for consistency later on in the post, in the section titled: ["consistent syntax"](https://blog.rust-lang.org/inside-rust/2023/02/23/keyword-generics-progress-report-feb-2023.html#consistent-syntax). So I think you're absolutely right that this looks really weird and inconsistent, especially compared to the `?const trait` notation - but that's something we intentionally call out later in the post and is something we would actually like to address!

tyilo:
> Thanks for filing this PR! - I get that it's confusing, because what you wrote is indeed what we think the syntax should be! But by this point in the post, we haven't yet suggested changing the syntax of `const fn` to `?const fn`, despite having mentioned that `const fn` already does mean "maybe-const". Confusing!
> 
> We only suggest changing `const fn` to `?const fn` for consistency later on in the post, in the section titled: ["consistent syntax"](https://blog.rust-lang.org/inside-rust/2023/02/23/keyword-generics-progress-report-feb-2023.html#consistent-syntax). So I think you're absolutely right that this looks really weird and inconsistent, especially compared to the `?const trait` notation - but that's something we intentionally call out later in the post and is something we would actually like to address!

Hmm, but then shouldn't it also be `const ?async fn read_string` in the "Combining const and async" section and not `?const ?async`?

yoshuawuyts:
Ah, yes! That’s a good catch!

PR #1088, opened 2023-03-08
Announcing Rust 1.68.0

cuviper:


posts/2023-03-09-Rust-1.68.0.md
```diff
@@ -0,0 +1,132 @@
...
+subject to the normal borrow checker to make sure it can't outlive that local.
+
+### Default `alloc` error handler
+
+When allocation fails in Rust, APIs like `Box::new` and `Vec::push` have no way
+to indicate that failure, so some divergent execution path needs to be taken.
+When using the `std` crate, the program will print to `stderr` and abort. But
+with `#![no_std]` and the `alloc` crate, it was required to provide a
+`#[alloc_error_handler]` function, similar to `#[panic_handler]`, but that
+attribute is still unstable.
```
CryZe, line 101: It may make sense to reword that, as the attribute is likely getting removed entirely. It's currently in FCP for removal: https://github.com/rust-lang/rust/issues/51540#issuecomment-1448404145

CryZe:
```suggestion
`#[alloc_error_handler]` function, similar to `#[panic_handler]`. Moving forward the plan is to treat running out of memory as normal panics.
```

Maybe something like that?

posts/2023-03-09-Rust-1.68.0.md
```diff
@@ -0,0 +1,132 @@
...
+
+Now in Rust 1.68.0, a default handler is provided which turns that `alloc` error
+into a `panic!`, which may be processed in a `#[panic_handler]` like any other.
```
Mark-Simulacrum, line 104:
```suggestion
When allocation fails in Rust, APIs like `Box::new` and `Vec::push` have no way
to indicate that failure, so some divergent execution path needs to be taken.
When using the `std` crate, the program will print to `stderr` and abort.
As of Rust 1.68.0, binaries which do not include `std` will continue to have
this behavior. Binaries which only include the `alloc` crate will now `panic!`
on allocation failure, which may be further adjusted via a `#[panic_handler]` if desired.

In the future, it's likely that the behavior for `std` will also be changed to match that of `alloc`-only binaries.
```

posts/2023-03-09-Rust-1.68.0.md
Mark-Simulacrum: Posted a more complete rewrite in https://github.com/rust-lang/blog.rust-lang.org/pull/1088/files#r1130197777 to incorporate this feedback, would appreciate feedback.

posts/2023-03-09-Rust-1.68.0.md
cuviper: > binaries which do not include `std` will continue to have this behavior.

I think you meant *do* include `std`?

Mark-Simulacrum: Yes, updated suggestion. Thanks!

cuviper: > Binaries which do not include `std`, only including `alloc`, will now `panic!`

My only hesitation is that we're not saying what it used to do -- although I guess that's empty/nonexistent on stable. Should we add that it wasn't allowed without `std` at all before?

cuviper: OK, fair enough.

Mark-Simulacrum (approved):

PR #1102, opened 2023-04-28
Post: Stabilizing async fn in traits in 2023

tmandry:
This post lays out our plan for how to stabilize `async fn` in traits in 2023. Co-authored with @nikomatsakis.

cc @rust-lang/wg-async 

Targeting Wednesday May 3 for publishing.


yoshuawuyts:
I'd love to make time to review this, but won't be able to during the weekend. Can the target publication date be pushed back to Tuesday instead of Monday so we can have at least one workday to read this over?

tmandry:
Pushed to Tuesday.

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,197 @@
...
+
+In November, we [blogged about nightly support for async fn in trait][pp] and identified some key next steps, most importantly [support for send bounds to allow spawning in generic functions](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#limitation-spawning-from-generics). Since then we've done a significant amount of design exploration and collected a set of case studies evaluating how well the current code works in practice.
+
+As of now, all of the functionality described in this blog post is implemented and available in some form on the nightly compiler. This was done to prove out the viability, but not all of it has been formally RFC'd. We'll link to playground examples and RFCs where appropriate so you can try it for yourself or read about the details.
+
+[pp]: https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html
+
+## MVP Part 1: Core support for "async functions in traits"
+
+The easiest way to explain what we are going to stabilize is to use a code example. To start, we will permit the use of `async fn` in traits definitions...
```
calebcartwright, line 18:
```suggestion
The easiest way to explain what we are going to stabilize is to use a code example. To start, we will permit the use of `async fn` in trait definitions...
```

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,197 @@
...
+{
+    tokio::task::spawn(async move {
+        if !hc.check().await {
+            log_health_check_failure().await;
+        }
+    });
+}
+```
+
+Of course, it's kind of unfortunate that we had to rewrite from taking an `impl HealthCheck` to an explicit `HC` type parameter in order to use this notation. RFC #2289, "associated type bounds", introduced a compact notation to address this problem. That RFC is not part of this MVP, but if it were stabilized, then one could simply write:
```
calebcartwright, line 87: Not sure whether the link was intentionally omitted, but perhaps worth including if not?

```suggestion
Of course, it's kind of unfortunate that we had to rewrite from taking an `impl HealthCheck` to an explicit `HC` type parameter in order to use this notation. [RFC 2289](https://github.com/rust-lang/rfcs/pull/2289), "associated type bounds", introduced a compact notation to address this problem. That RFC is not part of this MVP, but if it were stabilized, then one could simply write:
```

calebcartwright (commented):
exciting stuff! one likely typo and potential missing link noted below

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,197 @@
...
+    //   -------------------------
+    //   Does not capture `self` as it does
+    //   not include a `+ '_`.
+}
+```
+
+Even though the need for "impl trait in traits" comes up a lot in async, they are a general feature that will be useful in many contexts having nothing to do with async (for example, returning iterators from trait methods).
+
+**Status:** Return-position impl trait in traits have an experimental implementation and are described in the recently opened [RFC 3425].
+
```
nrc, line 127: I think it is worth mentioning the ability to impl an async fn in a trait with an `impl Future` method. This feels like a significant part of 3425 to me

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,197 @@
...
+## What the MVP will not support or won't support well
+
+The case studies revealed two situations that the MVP doesn't support very well, but both of them have workarounds available. These workarounds are mechanical and once the MVP is available on stable it will be possible to automate them via a custom derive or other crates on crates.io.
+
+### Modeling dynamic dispatch
+
+In the MVP, traits that use async functions are not "dyn safe", meaning that they don't support dynamic dispatch. So e.g. given the `HealthCheck` trait we saw earlier, one could not write `Box<dyn HealthCheck>`. At first, this seems like a crucial limitation, since many of the use cases require dynamic dispatch! But it turns out that there is a workaround. One can define an "erased" trait internally to your crate that enables dynamic dispatch. The process was pioneered by crates like [erased serde][] and is explained in detail in the [builder-provider case study][].
+
+In the future, async fn should work with `dyn Trait` directly.
```
nrc, line 152: Is it worth linking to one of the blog posts on the possible designs here?

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,197 @@
...
+    tokio::task::spawn(async move {
+        if !hc.check().await {
+            log_health_check_failure().await;
+        }
+    });
+}
+```
+
+In our [previous post][pp], we [hypothesized](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#hypothesis-this-is-uncommon) that this problem might not occur often in practice. However, our case studies found that it comes up quite frequently, and so we decided that a solution is needed. We explored a number of solutions and concluded that associated return types are the most practical.
+
```
nrc, line 101: I think we should probably have async methods be Send by default and use RTN to opt out of the Send-ness (given that Send futures are more common). We should discuss this at some point before getting too far along with stabilisation. Not sure it's worth mentioning the possibility in the blog post?

nrc (approved):

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
yoshuawuyts: We have discussed this on a number of occasions already, and the following two objections are usually raised:
1. Rust uses opt-in `Send` bounds everywhere; making `async fn` in traits imply `+ Send` by default would be inconsistent with the rest of the language. For example: `async fn` free functions do not imply `+ Send`, making the "send"-ness of `async fn` contextually dependent on whether it's used in a trait. That seems tricky to remember, and easy to mess up.
2. While `+ Send` types and bounds are the norm today for `async-std`/`tokio`-style executors today, these aren't the only domain and executor styles we need to care about. My understanding is that this would cause issues for single-threaded domains such as embedded. @dirbaio should be able to elaborate this point.

Dirbaio: Yes, in embedded the default is single-threaded executors. Multi-core chips are rare, and most of them are "asymmetric" (different cores have separate dedicated RAM/flash, separate peripherals, different speeds or even different architecture). Multi-core is usually used by manually assigning different functions to different cores, to get better isolation/security/predictability thanks to not having latency for one function impacted by other functions running on the same core. You never need to "load-balance work" between multiple cores to scale performance, these chips are usually slow (<200Mhz) so it's simpler to get faster cores if you need more perf. It's only when you reach multi-GHz territory that you need multiple cores to scale perf.

So, embedded uses one single-threaded executor for each core, instead of a big multithreaded one. Having AFIT be Sync by default would be very, very annoying, and inconsistent with the rest of the language. Forcing Send is especially annoying in embedded, because you don't have `std::sync::Mutex`, you often have no heap and some archs dont't have atomic CAS (like `thumbv6`, so no `Arc` either.

yoshuawuyts (changes requested):

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
nrc: > We have discussed this on a number of occasions already, and the following two objections are usually raised:

We have, but I feel like every time the discussion has just kind of drifted off without us making a plan or coming to a decision (in part, because decisions about defaults are hard - its just about ergonomics rather than anything which can be argued logically). I don't really want to discuss here, my comment was just about whether we should say something about it in the blog post.

To address your points though, Rust doesn't use opt-in Send bounds, it infers the Send bounds. It's just that with trait methods you can't infer them so we have to pick a default. This will be confusing either way, so we just have to pick the least confusing. Which segues into point 2 - nobody is saying we shouldn't support non-Send methods at all, just whether they should be opt-in or opt-out. Single-threaded runtimes are totally valid and should be supported, but I don't see them being more popular than work stealing. To clarify, these are defaults for the callsite, not for the declaration, so I don't think there is any downside other than more typing.

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,197 @@
...
+    //                                            -------------
+    tokio::task::spawn(async move {
+        if !hc.check().await {
+            log_health_check_failure().await;
+        }
+    });
+}
+```
+
+In our [previous post][pp], we [hypothesized](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#hypothesis-this-is-uncommon) that this problem might not occur often in practice. However, our case studies found that it comes up quite frequently, and so we decided that a solution is needed. We explored a number of solutions and concluded that associated return types are the most practical.
```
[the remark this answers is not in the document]
tmandry: I don't agree with the conclusion of being "deeply unpleasant to both read and write". The case studies do not support that. The only critical text I could find was in the Microsoft study:

>  Once we did this, we needed to add `save(..): Send` and `restore(..): Send` bounds to the places where the Component trait was used as a DynComponent. There were only two methods we needed to bound, but this was still mildly annoying.

> Return type notation looks like it would work, but even in our limited experience with it we have run into ergonomics issues.

The syntax has become more ergonomic since the studies by removing the need for `..`.

We did discuss adding a proc macro for the cases where you need multiple `Send` bounds for a given trait, to make that more convenient to write, and it might be good to mention that in this post. (Quite possibly for doing the `DynTrait` workaround for you as well.)

But I expect that people will frequently use RTN directly, so I would not frame it as purely a low-level feature. I also don't want to include trait transformers in this post because we are not including it in the MVP. We can mention that we are thinking of ways to make common cases more ergonomic, and that this feature could be seen as a more granular building block for that.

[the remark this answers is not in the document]
tmandry: I added a mention of an upcoming proc macro alongside the possibility of trait transformers.

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
tmandry: Added a link to the "user's guide" on the async-fundamentals site.

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,197 @@
...
+{}
+
+impl<T> SendHealthCheck for T
+where
+    T: HealthCheck + Send,
+    T::check(): Send,
+{}
+```
+
+Using a pattern like this means you can write `T: SendHealthCheck`. In the future, something like [trait transformers][] may provide a more concise syntax.
```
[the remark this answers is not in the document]
tmandry: I think I'd like to stick with a reference to a proc macro for now.

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
tmandry: We should move this discussion to Zulip.

tmandry:
Feedback should be addressed now.

yoshuawuyts:
I've un-resolved an issue because in my opinion this post does not yet fully capture the collective decisions made by the Async WG. Unless these decisions were walked back in a later meeting I was not present at [^on-readers], I would like to see them represented in this post.

[^on-readers]: For on-readers: The Async WG has moved to a meeting schedule which alternates between EU-friendly and Asia/Oceania-friendly times every week. This means it's a legitimate possibility that this decision was walked back in a meeting in one of the EU off-weeks.

tmandry:
@yoshuawuyts I added a preview of the proc macro syntax along with some wording alluding to this in the MVP Part 2 section. I think the post heading is quite clear in saying what I think the post should say:

> ### Send bounds are verbose, especially for traits with lots of methods

I don't personally see a need to put this more front and center, because I don't think most readers will be directly affected or think about this, and I still think we need to gain more experience with the feature to see how _often_ this becomes annoying in practice. (I hope this doesn't come across as dismissive of your or @eholk's experience – I think it's valid and that's why this section exists in the post in the first place.)

> The phrase: "We explored a number of solutions and concluded that associated return types are the most practical." could be read to suggest we are happy with the ergonomics of associated return types, while in the meeting on case studies it was clearly established that using it felt ["overly verbose"](https://hackmd.io/YdliaHunQ_2StmO3kIoVRA#Yosh-told-me-you-disliked-RTN-eholk) and unpleasant. The "MVP part 2" section should make it clear that we're not yet satisfied with the ergonomics.

Implying that isn't really my intent, but I think the statement still rings true. I added some wording to the post that explains why: People have a need for the fine-grained control provided by ART, so we should stabilize that more general mechanism first. I feel good about the sentence you quoted, but happy to take suggestions for rewording.

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,229 @@
...
+team: The Rust Async Working Group <https://www.rust-lang.org/governance/wgs/wg-async>
+---
+
+The async working group's headline goal for 2023 is to stabilize a "minimum viable product" (MVP) version of async functions in traits. We are currently targeting Rust 1.74 for stabilization. This post lays out the features we plan to ship and the status of each one.
+
+In November, we [blogged about nightly support for async fn in trait][pp] and identified some key next steps, most importantly [support for send bounds to allow spawning in generic functions](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#limitation-spawning-from-generics). Since then we've done a significant amount of design exploration and collected a set of case studies evaluating how well the current code works in practice.
+
+As of now, all of the functionality described in this blog post is implemented and available in some form on the nightly compiler. This was done to prove out the viability, but not all of it has been formally RFC'd. We'll link to playground examples and RFCs where appropriate so you can try it for yourself or read about the details.
```
nikomatsakis, line 12:
```suggestion
As of now, all of the functionality described in this blog post is available on the nightly compiler. Some of this work was implemented using [experimental feature gates](https://lang-team.rust-lang.org/how_to/experiment.html) so that we could conduct case studies and prove out viability; we are in the process of authoring RFCs for those features now (full details below).
```

nikomatsakis: Probably a good idea to be clear about the process here, it mightbe confusing

nikomatsakis (commented):
Minor nits.

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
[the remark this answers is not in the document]
nikomatsakis: Hmm, I don't disagree with what you're saying Yosh, but I do feel like you're saying it in stronger terms than I would. I think a proc macro is a good idea to address both dyn dispatch + multiple methods, and all things being equal I think it'd be good for the async WG to ship it (though it'd probably be ok to have it come from community too). I think ART (nee RTN) seem like a flexible solution that is fine for a single method but scales poorly if you have many methods, in which case the proc macro will be very desirable. This is not the same as "deeply unpleasant" though and I wouldn't want those words in the post. I do think it's fine though to mention the proc macro as a part of the strategy. 

(Also, I probably shouldn't have volunteered to write code, that was not realistic. We should coordinate with someone else.)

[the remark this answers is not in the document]
nikomatsakis:
```suggestion
In our [previous post][pp], we [hypothesized](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#hypothesis-this-is-uncommon) that this problem might not occur often in practice. However, our case studies found that it comes up quite frequently, and so we decided that a solution is needed. We explored a number of solutions and concluded that associated return types (ARTs) are a flexible and reasonably ergonomic building block, which makes them a good fit for an MVP. However, we also found that in traits with many methods, ARTs become verbose. As discussed later in the post, we expect to ship a proc-macro that can help workaround this in the short term while we gain experience.
```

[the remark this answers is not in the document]
nikomatsakis: Maybe something like this?

yoshuawuyts:
I had already logged off for the day, but I also don't want to further delay this post. @tmandry: from skimming the updated text it seems like it now more closely represents what we had agreed on. I haven't yet read Niko's proposed change, but from the context I trust it enough that between you two you can agree on the right phrasing to use.

Removing my blocking concern; thanks so much to both of you!

posts/inside-rust/2023-05-02-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,229 @@
...
+        if !hc.check().await {
+            log_health_check_failure().await;
+        }
+    });
+}
+```
+
+In our [previous post][pp], we [hypothesized](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#hypothesis-this-is-uncommon) that this problem might not occur often in practice. However, our case studies found that it comes up quite frequently, and so we decided that a solution is needed. We explored a number of solutions and concluded that associated return types (ARTs) are a flexible and reasonably ergonomic building block, which makes them a good fit for an MVP. However, we also found that in traits with many methods, ARTs become verbose. As discussed later in the post, we expect to ship a proc-macro that can help workaround this in the short term while we gain experience.
+ 
+**Status:** Associated return types have an experimental implementation and we are currently drafting an RFC. There are several open bugs that will need to be fixed. We are considering more concise syntax for the future (see below).
```
nikomatsakis, line 104:
```suggestion
In our [previous post][pp], we [hypothesized](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#hypothesis-this-is-uncommon) that this problem might not occur often in practice. However, our case studies found that it comes up quite frequently, and so we decided that a solution is needed. We explored a number of solutions and concluded that associated return types (ARTs) are a flexible and reasonably ergonomic building block, which makes them a good fit for an MVP.
 
**Status:** Associated return types have an experimental implementation and we are currently drafting an RFC. There are several open bugs that will need to be fixed. We also found that in traits with many methods, ARTs become verbose, and in the future may consider more concise syntax (see below).
```

nikomatsakis: I didn't notice the text in the status before; I like that. Maybe like this? Anyway, however we do it, I think something in the status line is good.

nikomatsakis (approved):

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,229 @@
...
+}
+```
+
+In our [previous post][pp], we [hypothesized](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#hypothesis-this-is-uncommon) that this problem might not occur often in practice. However, our case studies found that it comes up quite frequently, and so we decided that a solution is needed. We explored a number of solutions and concluded that associated return types (ARTs) are a flexible and reasonably ergonomic building block, which makes them a great fit for an MVP.
+ 
+**Status:** Associated return types have an experimental implementation and we are currently drafting an RFC. There are several open bugs that will need to be fixed. We also found that in traits with many methods, ARTs become verbose, and in the future may consider more concise syntax (see below).
+
+[Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=2066934a05cb9eafc0b47af7bdf8c57f)
+
+## MVP Part 3: "impl trait in traits" (return position)
```
eholk, line 108: This feels separable enough from AFIT that we could stabilize them separately, even though they obviously share a lot of the underlying implementation. In my mind they're separate features rather than different facets of the same one. That said, I'm in favor of stabilizing both and if it's easier to bundle them together then that's fine.

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,229 @@
...
+        &mut self, 
+        request: Request,
+    ) -> impl Future<Output = u32>;
+    //   -------------------------
+    //   Does not capture `self` as it does
+    //   not include a `+ '_`.
+}
+```
+
+Since `async fn` is sugar for a regular function returning `impl Future`, these two syntactic forms will work interchangeably.
```
eholk, line 128: If I remember right, the original AFIT RFC said specifically we wouldn't allow mixing and matching at first, right? Do we want to call out the design change?

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
```diff
@@ -0,0 +1,229 @@
...
+
+The case studies revealed two situations that the MVP doesn't support very well, but both of them have workarounds available. These workarounds are mechanical and once the MVP is available on stable it will be possible to automate them via a custom derive or other crates on crates.io.
+
+### Modeling dynamic dispatch
+
+In the MVP, traits that use async functions are not "dyn safe", meaning that they don't support dynamic dispatch. So e.g. given the `HealthCheck` trait we saw earlier, one could not write `Box<dyn HealthCheck>`.
+
+At first, this seems like a crucial limitation, since many of the use cases require dynamic dispatch! But it turns out that there is a workaround. One can define an "erased" trait internally to your crate that enables dynamic dispatch. The process was pioneered by crates like [erased serde] and is explained in detail in the [builder-provider case study].
+
+To make this workaround easier in the near term, we hope to provide a proc macro to automate it. In the future, async fn [should work][ugf-dyn] with `dyn Trait` directly.
```
eholk, line 172:
```suggestion
To make this workaround easier in the near term, we plan to provide a proc macro to automate it. In the future, async fn [should work][ugf-dyn] with `dyn Trait` directly.
```

I'd like to make this sound like are committing to delivering this proc macro quickly, rather than just hoping we or someone else writes the macro.

eholk (approved):
Looks good to me. I made a couple suggestions, but I think the most important thing is to get this post out there so feel free to ignore my comments.

eholk:
> > ### Send bounds are verbose, especially for traits with lots of methods
> 
> I don't personally see a need to put this more front and center, because I don't think most readers will be directly affected or think about this, and I still think we need to gain more experience with the feature to see how _often_ this becomes annoying in practice. (I hope this doesn't come across as dismissive of your or @eholk's experience – I think it's valid and that's why this section exists in the post in the first place.)

I'm okay with how this is written. It highlights that we are aware of some ergonomic issues, but I think it's true we need more experience to find out big of an issue this is in practice. Publishing this blog post also serves as a call for people to try it out and see if it works for them.

The only potential open question we have is whether we are 100% convinced people will need or want the extra flexiblity afforded by ART. It seems likely to me that we will, but I also don't feel like we have a lot of data yet that says this. That said, I think it's fair to say that the sense of the group is that we will need this flexibility so we can stabilize it now and iterate on a less verbose solution.

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
tmandry: I think the AFIT story feels incomplete without RPITIT, but RPITIT can stand alone without AFIT. We could reorder the sections to reflect this, but since the post is focused on AFIT, maybe the current ordering makes the most sense.

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
tmandry: It's called out in the RPITIT RFC that's up right now. I don't think we need to highlight it in this post so much, but would be fine with adding a footnote or something like that.

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
eholk: I agree that this seems like the best ordering!

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
eholk: Sounds good. Yeah, the RFC is the right place to mention that. It's probably too small of a detail to call out in this blog post.

posts/inside-rust/2023-05-03-stabilizing-async-fn-in-trait.md
tmandry: I added some wording to the status line to reinforce that RPITIT can stand alone.

nikomatsakis:
FYI: I plan to post today around 10am US Eastern

PR #1121, opened 2023-07-13
make link clickable

tshepang:


ehuss (approved):
Thanks!  Angle brackets around the URL would also work.

PR #1125, opened 2023-07-21
Update all dependencies to their latest version

Urgau:
cc @jdno


Turbo87:
@Urgau it looks like most of these changes have been addressed by renovatebot PRs now, expect for the SASS replacement. I guess it's probably easiest to open a dedicated PR for that if you want :)

PR #1129, opened 2023-08-23
Announcing Rust 1.72.0

cuviper:


posts/2023-08-24-Rust-1.72.0.md
```diff
@@ -0,0 +1,60 @@
...
+  - The second checks for calls to `std::str::from_utf8` and `std::str::from_utf8_mut` with an invalid UTF-8 literal, which will always return an error. This lint is a warning by default.
+
+* [`clippy::cmp_nan`](https://rust-lang.github.io/rust-clippy/rust-1.71.0/index.html#cmp_nan) to [`invalid_nan_comparisons`](https://doc.rust-lang.org/1.72.0/rustc/lints/listing/warn-by-default.html#invalid-nan-comparisons) (warn)
+  - This checks for comparisons with `f32::NAN` or `f64::NAN` as one of the operands. NaN does not compare meaningfully to anything – not even itself – so those comparisons are always false. This lint is a warning by default, and will suggest calling the `is_nan()` method instead.
+
+* [`clippy::cast_ref_to_mut`](https://rust-lang.github.io/rust-clippy/rust-1.71.0/index.html#cast_ref_to_mut) to [`invalid_reference_casting`](https://doc.rust-lang.org/1.72.0/rustc/lints/listing/allowed-by-default.html#invalid-reference-casting) (allow)
```
Nilstrieb, line 36: Maybe it's worth pointing out that it's currently `allow` because there are some issues with it, and it will be raised when they are fixed. People may be surprised why a "this is UB" lint is allow-by-default.

posts/2023-08-24-Rust-1.72.0.md
```diff
@@ -0,0 +1,60 @@
...
+- [`CStr::from_bytes_with_nul`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_bytes_with_nul)
+- [`CStr::to_bytes`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.to_bytes)
+- [`CStr::to_bytes_with_nul`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.to_bytes_with_nul)
+- [`CStr::to_str`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.to_str)
+
+### Other changes
+
+Check out everything that changed in [Rust](https://github.com/rust-lang/rust/releases/tag/1.72.0), [Cargo](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-172-2023-08-24), and [Clippy](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-172).
+
+In a future release we're planning to increase the minimum supported Windows version to 10. The accepted proposal in compiler [MCP 651](https://github.com/rust-lang/compiler-team/issues/651) is that Rust 1.75 will be the last to officially support Windows 7, 8, and 8.1. When Rust 1.76 is released in February 2024, only Windows 10 and later will be supported as tier-1 targets.
```
djc, line 56: This feels a bit tucked away. Maybe extract it from "Other changes" and give it its own header?

cuviper: We could -- I was just repeating the way we pre-announced a Linux bump in 1.61:
https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html#other-changes

How about a heading like, "Future Windows compatibility"?

posts/2023-08-24-Rust-1.72.0.md
cuviper: I was trying to hint at that in the following line, although I don't really know what exactly is lacking in the lint as-is. False positives?

posts/2023-08-24-Rust-1.72.0.md
thomcc: Not opposed to this but we should probably pin down what the requirements are for updating a target. E.g. if dropping old windows needs only an MCP and mention in a release post, it's a bit odd that dropping old macOS needs an MCP, multi-team FCP, and dedicated blog post (especially when old macOS is far more obscure than old windows).

thomcc: Yeah, I'm just surprised to see this mentioned without doing that for Windows.

cuviper: It was requested [here](https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/1.2E72.2E0/near/386580207) -- if we feel this is premature, we can wait for a later release. Again looking at the Linux case, the initial mention in release notes was 2 months after FCP was _proposed_, but that was blocked on having a dedicated blog post, which was posted 3 months later.

cuviper: ^cc @ChrisDenton

ChrisDenton: For Windows XP we didn't have all of that. But then nobody knew what the process was at that time so I just jumped on that when someone suggested it. I dunno. There's no official process so it feels like it's always different every time.

ChrisDenton: I really do want to give people as much notice as possible. Maybe it could be hedged? I mean, "planned" doesn't necessarily mean it's definite.

Mark-Simulacrum: I think we should keep this notice. I don't think it should be taken as implying this is *enough* or not enough -- this is not a post stating what the requirements are for future changes (or even this change). If we push out the actual switch I doubt people will be unhappy, all this really commits us to (and even that lightly) is a *minimum* delay until 1.75.

posts/2023-08-24-Rust-1.72.0.md
cuviper: @Urgau, can you clarify the reason for "allow" in brief layman's terms?
(between rust-lang/rust#113422 and rust-lang/rust#112431)

Urgau: Does it work for you?

This lint was temporarily switched to "allow" because the clippy implementation that was uplifted was linting on expressions that are UB only with Stack Borrow and Tree Borrow and since none of these are official, it was decided to made to temporarily "allow" the lint while fixing it for the next release.

cuviper: > Does it work for you?

I haven't tried it. I'm only trying to describe what we're releasing.

Is it sufficient to just say it was allowed "due to potential false positives"? I think that covers "things that may be UB but aren't decided yet," i.e. SB / TB.

Urgau: Certainly, that's fine by me.

cuviper: Oh, I think I misunderstood that you were suggesting text to add -- but I think brevity is fine here. Thanks!

posts/2023-08-24-Rust-1.72.0.md
```diff
@@ -0,0 +1,81 @@
...
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.72.0](https://github.com/rust-lang/rust/releases/tag/1.72.0) on GitHub.
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.72.0 stable
+
```
joshtriplett, line 21:
``````suggestion

### Rust reports potentially useful `cfg`-disabled items in errors

You can conditionally enable Rust code using `cfg`, such as to provide certain
functions only with certain crate features, or only on particular platforms.
Previously, items disabled in this way would be effectively invisible to the
compiler. Now, though, the compiler will remember the name and `cfg` conditions
of those items, so it can report (for example) if a function you tried to call
is unavailable because you need to enable a crate feature.

```
   Compiling my-project v0.1.0 (/tmp/my-project)
error[E0432]: unresolved import `rustix::io_uring`
   --> src/main.rs:1:5
    |
1   | use rustix::io_uring;
    |     ^^^^^^^^^^^^^^^^ no `io_uring` in the root
    |
note: found an item that was configured out
   --> /home/username/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustix-0.38.8/src/lib.rs:213:9
    |
213 | pub mod io_uring;
    |         ^^^^^^^^
    = note: the item is gated behind the `io_uring` feature

For more information about this error, try `rustc --explain E0432`.
error: could not compile `my-project` (bin "my-project") due to previous error
```

``````

Comment (joshtriplett) [reply posts/2023-08-24-Rust-1.72.0.md:21]: This seems like another headline user-visible feature of this release, making `cfg` easier to work with.

File: posts/2023-08-24-Rust-1.72.0.md [hunk 4]
```diff
@@ -0,0 +1,81 @@
...
+- [`CStr::to_bytes_with_nul`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.to_bytes_with_nul)
+- [`CStr::to_str`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.to_str)
+
+### Other changes
+
+Check out everything that changed in [Rust](https://github.com/rust-lang/rust/releases/tag/1.72.0), [Cargo](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-172-2023-08-24), and [Clippy](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-172).
+
+### Future Windows compatibility
+
+In a future release we're planning to increase the minimum supported Windows version to 10. The accepted proposal in compiler [MCP 651](https://github.com/rust-lang/compiler-team/issues/651) is that Rust 1.75 will be the last to officially support Windows 7, 8, and 8.1. When Rust 1.76 is released in February 2024, only Windows 10 and later will be supported as tier-1 targets.
```
alex, line 77: FWIW, it's not clear to me if section is describing the requirements for running `rustc` on Windows, or if this is describing a mininum requirement for all rust projects built for windows.

ChrisDenton: With tier 1, this is currently the same thing. The MCP leaves open the possibility of a new target that could be a lower tier, if there are people willing and able to maintain it.

alex: Understood. I think it'd be helpful to readers if it was made unambiguous in the blog post.

cuviper: Do we want to _raise_ that new-target possibility? Or just say something like "This change will apply both as a host compiler and as a compilation target."

posts/2023-08-24-Rust-1.72.0.md
```diff
@@ -0,0 +1,109 @@
...
+To prevent user-provided const evaluation from getting into a compile-time
+infinite loop or otherwise taking unbounded time at compile time, Rust
+previously limited the maximum number of *statements* run as part of any given
+constant evaluation. However, especially creative Rust code could hit these
+limits and produce a compiler error. Worse, whether code hit the limit could
+vary wildly based on libraries invoked by the user; if a library you invoked
+split a statement into two within one of its functions, your code could then
+fail to compile.
+
+Now, you can do an unlimited amount of const evaluation at compile time, To
```
djc, line 61: "at compile time, To"?

joshtriplett, line 61:
```suggestion
Now, you can do an unlimited amount of const evaluation at compile time. To
```

Mark-Simulacrum (approved):

PR #1132, opened 2023-08-23
Remove a couple of possible panics when opening a post

LawnGnome:
I'm not actually so bothered about the panics per se, but this way a naïve user like myself who accidentally created a zero length file in `posts` gets a useful error message.

### DIFF
--- a/src/posts.rs
+++ b/src/posts.rs
@@ -50,10 +50,15 @@ impl Post {
         let filename = split.next().unwrap().to_string();
 
         let contents = std::fs::read_to_string(path)?;
+        if contents.len() < 5 {
+            return Err(
+                format!("{path:?} is empty, or too short to have valid front matter").into(),
+            );
+        }
 
         // yaml headers.... we know the first four bytes of each file are "---\n"
         // so we need to find the end. we need the fours to adjust for those first bytes
-        let end_of_yaml = contents[4..].find("---").unwrap() + 4;
+        let end_of_yaml = contents[4..].find("---\n").unwrap() + 4;
         let yaml = &contents[..end_of_yaml];
         let YamlHeader {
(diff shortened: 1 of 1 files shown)


src/posts.rs
```diff
@@ -50,10 +50,15 @@ impl Post {
...
+        if contents.len() < 5 {
+            return Err(
+                format!("{path:?} is empty, or too short to have valid front matter").into(),
+            );
+        }
 
         // yaml headers.... we know the first four bytes of each file are "---\n"
         // so we need to find the end. we need the fours to adjust for those first bytes
-        let end_of_yaml = contents[4..].find("---").unwrap() + 4;
+        let end_of_yaml = contents[4..].find("---\n").unwrap() + 4;
```
CleanCut, line 61: Given that the method returns a `Result`, it seems like every `.unwrap()` call in this method would be better as a `?`, including this one. Even with your extra check at the top, a missing or malformed header will still crash. Why crash when you can return the error?

```suggestion
        let end_of_yaml = contents[4..].find("---\n")? + 4;
```

LawnGnome: I mean, honestly, I was just fitting in with the house style around most of it panicking instead of erroring and making the minimal change to get a useful error, rather than trying to fix all the things.

I'm happy to spend more time on this if the responsible people (T-website?) are interested in having that happen.

CleanCut: 🤷🏻‍♂️ I'm just making a drive-by comment as an observer. Feel free to ignore me if it doesn't sound like a good idea.

Turbo87 (approved):
seems like a useful and well-scoped improvement 👍 

I agree that the rest can be cleaned up separately 😅

PR #1146, opened 2023-09-22
Specify what “threads on the RFC diff” is in opposition to.

kpreid:
I believe this is a useful clarification because without the contrast, it may not be clear what is being asked for, to people not already familiar with the details of GitHub code review features.


Turbo87 (approved):
👍 

thank you!

PR #509, opened 2020-01-30
Switch to GitHub Actions from Travis CI

XAMPPRocky:
r? @pietroalbini

PR #1152, opened 2023-10-13
Remove mention of Travis CI from README

foonathan:
As of #509, GitHub Actions is used instead.

It just bothered me when I read it :D


ehuss (approved):
Thanks!

PR #1155, opened 2023-10-20
Announce major Generator -> Coroutine rename

oli-obk:
cc https://github.com/rust-lang/rust/pull/116447
cc https://github.com/rust-lang/rust/pull/116958

### DIFF
--- a/posts/inside-rust/2023-10-23-coroutines.md
+++ b/posts/inside-rust/2023-10-23-coroutines.md
@@ -1,0 +1,25 @@
+---
+layout: post
+title: Generators are dead, long live coroutines, generators are back
+author: oli-obk
+---
+
+We have renamed the unstable `Generator` trait to `Coroutine` and adjusted all terminology accordingly.
+If you want to see all 3800 modified lines of code, you can find the PR [here](https://github.com/rust-lang/rust/pull/116958).
+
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):
Seems ok!

the8472:
Probably should mention that it's only relevant for nightly users.

oli-obk:
Updated. what do I need to do to publish this?

Mark-Simulacrum (approved):

Mark-Simulacrum:
Currently date is set for tomorrow, I can merge then (ping me). Or happy to publish today, but then date should be updated.

Mark-Simulacrum (approved):

jeffparsons:
> huh... I... should probably not be writing blog posts at 2am... not sure why I used the 24th.

No, you definitely should, because this was super entertaining reading. Can you write all the Rust blog posts from now on? 😛

One question, though: you mentioned that generators are a convenient way to write iterators. Isn't there also a performance angle? For some reason I had the impression that the compiler-generated state machine would be faster than what, e.g., `chain` can do today. Is that actually true, or did I make that up?

oli-obk:
For a very simple `chain` I would think that it comes out as equivalent. We'll have to actually look at examples and codegen to see where we can improve things. In the end, we'll try to be zero cost where plausible

PR #1156, opened 2023-10-24
Add "crates.io: Dropping support for non-canonical downloads" blog post

Turbo87:
As discussed in the crates.io team meeting last week, we would like to sunset support for "non-canonical downloads" on crates.io. This blog post officially deprecates it with a cutoff date roughly a month from now.

Bildschirmfoto 2023-10-24 um 10 55 12

@rust-lang/crates-io r?

### DIFF
--- a/posts/2023-10-24-crates-io-non-canonical-downloads.md
+++ b/posts/2023-10-24-crates-io-non-canonical-downloads.md
@@ -10,11 +10,11 @@ team: the crates.io team <https://www.rust-lang.org/governance/teams/crates-io>
 - We want to improve the reliability and performance of crate downloads.
 - "Non-canonical downloads" (that use URLs containing hyphens or underscores where the crate published uses the opposite) are blocking these plans.
 - On 2023-11-20 support for "non-canonical downloads" will be disabled.
-- Cargo users are unaffected.
+- cargo users are unaffected.
 
 ## What are "non-canonical downloads"?
 
-The "non-canonical downloads" feature allows everyone to download the `serde_derive` crate from <https://crates.io/api/v1/crates/serde_derive/1.0.189/download>, but also from <https://crates.io/api/v1/crates/serde-derive/1.0.189/download>, where the underscore was replaced with a hyphen (Crates.io normalizes underscores and hyphens to be the same for uniqueness purposes, so it isn't possible to publish a crate named `serde-derive` because `serde_derive` exists). The same also works vice versa, if the canonical crate name uses hyphens and the download URL uses underscores instead. It even works with any other combination for crates that have multiple such characters (please don't mix them…!).
+The "non-canonical downloads" feature allows everyone to download the `serde_derive` crate from <https://crates.io/api/v1/crates/serde_derive/1.0.189/download>, but also from <https://crates.io/api/v1/crates/serde-derive/1.0.189/download>, where the underscore was replaced with a hyphen (crates.io normalizes underscores and hyphens to be the same for uniqueness purposes, so it isn't possible to publish a crate named `serde-derive` because `serde_derive` exists). The same also works vice versa, if the canonical crate name uses hyphens and the download URL uses underscores instead. It even works with any other combination for crates that have multiple such characters (please don't mix them…!).
 
 ## Why remove it?
 
--- a/posts/2023-10-24-crates-io-non-canonical-downloads.md
+++ b/posts/2023-10-24-crates-io-non-canonical-downloads.md
@@ -1,0 +1,53 @@
+---
+layout: post
+title: "crates.io: Dropping support for non-canonical downloads"
+author: Tobias Bieniek
+team: the crates.io team <https://www.rust-lang.org/governance/teams/crates-io>
+---
+
+## tl;dr
+
+- we want to improve the reliability and performance of crate downloads
+- "non-canonical downloads" are blocking these plans
+- on 2023-11-20 support for "non-canonical downloads" will be disabled
+- cargo users are unaffected
+
+## What are "non-canonical downloads"?
+
+The "non-canonical downloads" feature allows everyone to download the `serde_derive` crate from <https://crates.io/api/v1/crates/serde_derive/1.0.189/download>, but also from <https://crates.io/api/v1/crates/serde-derive/1.0.189/download>, where the underscore was replaced with a hyphen. The same also works vice versa, if the canonical crate name uses hyphens and the download URL uses underscores instead. It even works with any other combination for crates that have multiple such characters (please don't mix them…!).
+
+## Why remove it?
+
+Supporting such non-canonical download requests means that the crates.io server needs to perform a database lookup for every download request to figure out the canonical crate name. The canonical crate name is then used to construct a download URL and the client is HTTP-redirected to that URL.
+
+While we have introduced a caching layer some time ago to address some of the performance concerns, having all download requests go through our backend servers has still started to become problematic and at the current rate of growth will not become any easier in the future.
+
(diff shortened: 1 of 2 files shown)


posts/2023-10-24-crates-io-non-canonical-downloads.md
```diff
@@ -0,0 +1,53 @@
...
+
+Looking at the crates.io request logs, the following user-agents are currently relying on "non-canonical downloads" support:
+
+- cargo-binstall/1.1.2
+- Faraday v0.17.6
+- Go-http-client/2.0
+- GNU Guile
+- python-requests/2.31.0
+
+Three of these are just generic HTTP client libraries. [GNU Guile](https://www.gnu.org/software/guile/) is apparently a programming language, so most likely this is also a generic user-agent from a custom user program.
```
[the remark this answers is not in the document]
Turbo87: interesting, thanks for the context. I assume that means that potentially some of these build recipes would have to be fixed to use the canonical names?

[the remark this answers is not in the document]
bjorn3: I presume so. I don't use Guix myself, so you may want to ask someone who does or one of the Guix devs: https://guix.gnu.org/en/contribute/

Mark-Simulacrum (approved):
Text looks fine (but please make sure team is happy with it :)

posts/2023-10-24-crates-io-non-canonical-downloads.md
```diff
@@ -0,0 +1,53 @@
...
+author: Tobias Bieniek
+team: the crates.io team <https://www.rust-lang.org/governance/teams/crates-io>
+---
+
+## tl;dr
+
+- we want to improve the reliability and performance of crate downloads
+- "non-canonical downloads" are blocking these plans
+- on 2023-11-20 support for "non-canonical downloads" will be disabled
+- cargo users are unaffected
```
carols10cents, line 13: Pedants will complain about the capitalization/punctuation, also I think the TL;DR deserves a quick explanation of non-canonical downloads.

```suggestion
## TL;DR

- We want to improve the reliability and performance of crate downloads.
- "Non-canonical downloads" (that use URLs containing hyphens or underscores where the crate published uses the opposite) are blocking these plans.
- On 2023-11-20 support for "non-canonical downloads" will be disabled.
- Cargo users are unaffected.
```

posts/2023-10-24-crates-io-non-canonical-downloads.md
```diff
@@ -0,0 +1,53 @@
...
+
+## What are "non-canonical downloads"?
+
+The "non-canonical downloads" feature allows everyone to download the `serde_derive` crate from <https://crates.io/api/v1/crates/serde_derive/1.0.189/download>, but also from <https://crates.io/api/v1/crates/serde-derive/1.0.189/download>, where the underscore was replaced with a hyphen. The same also works vice versa, if the canonical crate name uses hyphens and the download URL uses underscores instead. It even works with any other combination for crates that have multiple such characters (please don't mix them…!).
```
carols10cents, line 17: Added a bit more explanation so people don't think underscore/hyphen confusion is a typosquatting vector:

```suggestion
The "non-canonical downloads" feature allows everyone to download the `serde_derive` crate from <https://crates.io/api/v1/crates/serde_derive/1.0.189/download>, but also from <https://crates.io/api/v1/crates/serde-derive/1.0.189/download>, where the underscore was replaced with a hyphen (Crates.io normalizes underscores and hyphens to be the same for uniqueness purposes, so it isn't possible to publish a crate named `serde-derive` because `serde_derive` exists). The same also works vice versa, if the canonical crate name uses hyphens and the download URL uses underscores instead. It even works with any other combination for crates that have multiple such characters (please don't mix them…!).
```

posts/2023-10-24-crates-io-non-canonical-downloads.md
```diff
@@ -0,0 +1,53 @@
...
+- Faraday v0.17.6
+- Go-http-client/2.0
+- GNU Guile
+- python-requests/2.31.0
+
+Three of these are just generic HTTP client libraries. [GNU Guile](https://www.gnu.org/software/guile/) is apparently a programming language, so most likely this is also a generic user-agent from a custom user program.
+
+`cargo-binstall` refers to https://github.com/cargo-bins/cargo-binstall. The maintainer is already aware of the upcoming change and confirmed that more recent versions of `cargo-binstall` should not be affected by this change.
+
+We recommend that any scripts relying on non-canonical downloads be adjusted to use the canonical names from the package index, the database dump, or the crates.io API instead. If you don't know which data source is best suited for you, we welcome you to take a look at <https://crates.io/data-access>.
```
carols10cents, line 43: Thinking about accessibility, I think it's better to have a screenreader read a description of the page we're linking to rather than the URL

```suggestion
We recommend that any scripts relying on non-canonical downloads be adjusted to use the canonical names from the package index, the database dump, or the crates.io API instead. If you don't know which data source is best suited for you, we welcome you to take a look at [the crates.io data access page](https://crates.io/data-access).
```

hi-rustin (approved):
Looks good to me. Thanks!

posts/2023-10-24-crates-io-non-canonical-downloads.md
```diff
@@ -0,0 +1,53 @@
...
+
+- cargo-binstall/1.1.2
+- Faraday v0.17.6
+- Go-http-client/2.0
+- GNU Guile
+- python-requests/2.31.0
+
+Three of these are just generic HTTP client libraries. [GNU Guile](https://www.gnu.org/software/guile/) is apparently a programming language, so most likely this is also a generic user-agent from a custom user program.
+
+`cargo-binstall` refers to https://github.com/cargo-bins/cargo-binstall. The maintainer is already aware of the upcoming change and confirmed that more recent versions of `cargo-binstall` should not be affected by this change.
```
carols10cents, line 41: Thinking about accessibility, I don't think it's great to have a screenreader read this URL; I think it'd be nicer to have a link and a quick explanation.

```suggestion
[`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall) is a tool enabling installation of binary artifacts of crates. The maintainer is already aware of the upcoming change and confirmed that more recent versions of `cargo-binstall` should not be affected by this change.
```

posts/2023-10-24-crates-io-non-canonical-downloads.md
```diff
@@ -0,0 +1,53 @@
...
+
+We recommend that any scripts relying on non-canonical downloads be adjusted to use the canonical names from the package index, the database dump, or the crates.io API instead. If you don't know which data source is best suited for you, we welcome you to take a look at <https://crates.io/data-access>.
+
+## What is the plan?
+
+1. **Today:** Announce the removal of support for non-canonical downloads on the main Rust blog.
+2. **2023-11-20:** Disable support for non-canonical downloads and return a migration error message instead.
```
carols10cents, line 48: I would explain the purposes and downsides of this stage a bit more, something like:

```suggestion
2. **2023-11-20:** Disable support for non-canonical downloads and return a migration error message instead, to alert remaining users of this feature of the need to migrate. This still needs to put load on the application to detect a request is using a non-canonical download URL.
```

carols10cents (commented):
Some suggestions to maybe prevent some questions/worries people might have

jtgeibel (approved):
In addition to `-` vs `_`, crate names are case sensitive. Do we want to mention that as well? (Marking as approved because I'm fine merging as-is if we haven't actually seen casing differences show up in the logs.)

Turbo87:
@jtgeibel very good point! I admit I had forgotten about that part. I've added another sentence to the blog post mentioning that.

PR #1145, opened 2023-09-21
Add post announcing project director nominees

eholk:
Getting this ready now. The plan is to go live on Friday, September 22.

PR #1158, opened 2023-10-24
Fix frontmatter assumption

Turbo87:
Using `---\n` assumes that all files use `\n` line endings, but one of our latest blog posts is using `\r\n` instead. This commit reverts the assumption about the frontmatter again to fix the rendering of that blog post.

This fixes a hidden merge conflict between #1145 and #1132.


jdno (approved):
Discussed with @Turbo87 offline. Fine to revert the assumption for now, but we should fix this properly to avoid similar issues in the future.

PR #1163, opened 2023-10-30
Elevating Performance and Page Optimization 🏎

sanjaiyan-dev:
Hi, this PR introduces some minor performance improvements:

1) Moving the script from the end of the body tag to the head tag and adding the `defer` attribute for faster page loading and rendering. (Benefit: Improved page loading speed)
Reference link: https://web.dev/articles/efficiently-load-third-party-javascript#defer

2) Adding `decoding="async"` to images for optimized rendering, enabling the browser to parallelize text and image loading, enhancing the overall user experience.

### DIFF
--- a/static/images/site.webmanifest
+++ b/static/images/site.webmanifest
@@ -3,12 +3,12 @@
     "short_name": "",
     "icons": [
         {
-            "src": "/android-chrome-192x192.png",
+            "src": "/images/android-chrome-192x192.png",
             "sizes": "192x192",
             "type": "image/png"
         },
         {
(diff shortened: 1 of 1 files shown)


bjorn3:
> Adding decoding="async" to images for optimized rendering, enabling the browser to parallelize text and image loading, enhancing the overall user experience.

Wouldn't that result in a relayout once the size of the image is known? In any case the logo is tiny enough that I don't think there is any benefit to this. Also the browser already does async decoding if it thinks it is beneficial. `decoding="auto"` is the default.

PR #1173, opened 2023-12-07
Add 1.74.1 post

Mark-Simulacrum:


posts/2023-12-07-Rust-1.74.1.md
```diff
@@ -0,0 +1,35 @@
...
+If you don't have it already, you can [get `rustup`][rustup] from the
+appropriate page on our website.
+
+[rustup]: https://www.rust-lang.org/install.html
+
+## What's in 1.74.1
+
+1.74.1 resolves a few regressions introduced in 1.74.0:
+
+- [Resolved spurious STATUS_ACCESS_VIOLATIONs in LLVM](https://github.com/rust-lang/rust/pull/118464)
```
Nilstrieb, line 28: I think "Fix crashes with LLVM FFI" or something like that would be more accurate, this makes it sound like the cause came from LLVM and not from rustc freeing LLVM things too early, causing use after frees. Not that it matters a lot.

PR #1175, opened 2023-12-10
Add 2024 Edition Call For Proposals post

bstrie:
cc @ehuss @m-ou-se

### DIFF
--- a/posts/2023-12-10-2024-Edition-CFP.md
+++ b/posts/2023-12-10-2024-Edition-CFP.md
@@ -1,0 +1,112 @@
+---
+layout: post
+title: "A Call for Proposals for the Rust 2024 Edition"
+author: "Ben Striegel on behalf of the Edition 2024 Project Group"
+release: false
+---
+
+The year 2024 is soon to be upon us, and as long-time Rust aficionados know,
+that means that a new Edition of Rust is on the horizon!
+
+## What is an Edition?
+
+You may be aware that a new *version* of Rust is released every six weeks.
+New versions of the language can both add things as well as change things,
+but only in backwards-compatible ways, according to Rust's
+[1.0 stability guarantee][stability].
+
+[stability]: https://blog.rust-lang.org/2014/10/30/Stability.html
+
+But does that mean that Rust can *never* make backwards-incompatible changes?
+Not quite! This is what an Edition is:
+Rust's mechanism for introducing backwards-incompatible changes in a backwards-compatible way.
+If that sounds like a contradiction,
+there are three key properties of Editions that preserve the stability guarantee:
+
+1. Editions are *opt-in*;
+crates only receive breaking changes if its authors explicitly ask for them.
+
+2. Crates that use older editions *never get left behind*;
+a crate written for the original Rust 2015 Edition is still supported by every Rust release,
+and can still make use of all the new goodies that accompany each new version,
+e.g. new library APIs, compiler optimizations, etc.
+
+3. An Edition *never splits the library ecosystem*;
+crates using new Editions can depend on crates using old Editions (and vice-versa!),
+so nobody ever has to worry about Edition-related incompatibility.
+
+In order to keep churn to a minimum, a new Edition of Rust is only released once every three years.
+We've had the [2015 Edition][2015], the [2018 Edition][2018], the [2021 Edition][2021],
+and soon, the 2024 Edition. And we could use your help!
+
+[2015]: https://doc.rust-lang.org/edition-guide/rust-2015/index.html
+
+[2018]: https://doc.rust-lang.org/edition-guide/rust-2018/index.html
+
+[2021]: https://doc.rust-lang.org/edition-guide/rust-2021/index.html
+
+## A call for proposals for the Rust 2024 Edition
+
+We know how much you love Rust, but let's be honest, no language is perfect,
+and Rust is no exception.
+So if you've got ideas for how Rust could be better if only that pesky stability guarantee
+weren't around, now's the time to share!
+Also note that potential Edition-related changes aren't just limited to the language itself:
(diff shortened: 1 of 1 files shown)


ehuss (approved):

posts/2023-12-10-2024-Edition-CFP.md
```diff
@@ -0,0 +1,112 @@
...
+[jeff]: https://www.youtube.com/watch?v=9nazm3_OXac
+
+As an example, in the 2015 Edition a dynamically-dispatched trait object
+has a type that looks like `Box<Foo>`.
+In the 2018 Edition, in order to distinguish this from the then-new,
+statically-dispatched `impl Trait` syntax, trait objects began to require the `dyn` keyword,
+so `Box<Foo>` became `Box<dyn Foo>`.
+To make this migration automatic, `cargo fix` gained the ability to rewrite the types
+of trait objects,
+as well as to rewrite identifiers in case any code out there was using `dyn` as a variable name.
```
m-ou-se, line 89: Not sure if this is a great example, because `&dyn Trait` was already valid syntax in Rust 2015, so the edition change is a bit subtle. Maybe `async` is easier, since that simply changed from a meaningless identifier to a keyword.

ehuss, line 89: If you want to have a different example, here's a suggestion:

```suggestion
As an example, the `async` and `await` keywords were added in the 2018 Edition in preparation for adding async support.
In the 2015 Edition, it was possible to have identifiers using those keywords like `let async = 1;`.
Adding the keywords would have been a breaking change without the Edition system.
With Edition support, the new keywords only exist in the 2018 and newer Editions.
Projects still declared as using the 2015 Edition continue to work.
To automatically migrate to the 2018 edition, `cargo fix --edition` will translate the code to use the raw-identifier syntax such as `let r#async = 1;` which allows it to continue to use the same identifier without conflicting with the new keyword. 
```

bstrie: I'm not particularly trying to highlight the `dyn` keyword itself moving from a weak to a strict keyword, but rather the syntax change of `Trait` to `dyn Trait`. My reasoning here is that "reserving a keyword" is IMO a relatively uninteresting solved problem, since we have both raw identifiers and now raw keywords to cover that niche. I would rather inspire people to think of less trivial ways that the syntax could be changed, of which `dyn Trait` is an example. This is also why I'm trying to think of a more interesting example than the introduction of `async`.

posts/2023-12-10-2024-Edition-CFP.md
```diff
@@ -0,0 +1,112 @@
...
+[2021]: https://doc.rust-lang.org/edition-guide/rust-2021/index.html
+
+## A call for proposals for the Rust 2024 Edition
+
+We know how much you love Rust, but let's be honest, no language is perfect,
+and Rust is no exception.
+So if you've got ideas for how Rust could be better if only that pesky stability guarantee
+weren't around, now's the time to share!
+Also note that potential Edition-related changes aren't just limited to the language itself:
+we'll also consider proposed changes to the standard library, Cargo, and rustfmt.
```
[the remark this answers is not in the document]
bstrie: Good point.

posts/2023-12-10-2024-Edition-CFP.md
```diff
@@ -0,0 +1,112 @@
...
+As an example, in the 2015 Edition a dynamically-dispatched trait object
+has a type that looks like `Box<Foo>`.
+In the 2018 Edition, in order to distinguish this from the then-new,
+statically-dispatched `impl Trait` syntax, trait objects began to require the `dyn` keyword,
+so `Box<Foo>` became `Box<dyn Foo>`.
+To make this migration automatic, `cargo fix` gained the ability to rewrite the types
+of trait objects,
+as well as to rewrite identifiers in case any code out there was using `dyn` as a variable name.
+
+## How to contribute
```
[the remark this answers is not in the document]
bstrie: Sure, I'll pull out some of the steps from the RFCs readme to make it more evident.

posts/2023-12-10-2024-Edition-CFP.md
ehuss: I'm fine with the existing example, this is just an alternate per mara's suggestion above.

I don't think there is an existing example that really exemplifies a substantial syntax change. All of the syntax changes are relatively subtle.

`dyn Trait` was implemented as a weak keyword, which was backwards-compatible and not related to an edition. The edition only *rejects* the old syntax under the intent that we wanted to force a stylistic change.

Anonymous trait function parameters is also like `dyn Trait`. It only rejects old syntax to force a style change. Same with `start...end` patterns.

The only syntax changes I can think of that aren't rejection of existing syntax for style reasons is async and literal prefixes which will be relevant in 1.76 in February with c-string literals, which would otherwise be a breaking change without edition support.

bstrie: C-string literals sound like they could make for a good example, since that also serves as a preview of Rust 2024. :P

ehuss: Just FYI, c-string literals are in 2021 (because that's where it was reserved), not 2024.

ehuss (approved):

rylev:
@bstrie do you want to update the post with today's date, and then we can merge it?

ehuss:
Sorry, was waiting to see if Mara had any other feedback. Thanks @bstrie!

ehuss:
Oops, I renamed the file but it didn't push to the correct remote. 😦

bstrie:
@ehuss Can we correct the date? It would be nice to not have it be buried in the index under the previous post.

PR #1177, opened 2023-12-14
Improve Page Load Performance by Deferring Scripts and Lazy Loading Images ✈️

sanjaiyan-dev:
Hi, I believe this pull request implements small but valuable performance improvements and follows best web practices.
The script tag is now run with the defer attribute, which improves performance. For more information, please see https://web.dev/articles/efficiently-load-third-party-javascript#defer.
We've added the loading attribute to images below the viewport, enabling lazy loading and optimizing network usage. For more information, please refer to https://web.dev/articles/browser-level-image-lazy-loading#the_loading_attribute.

### DIFF
--- a/templates/footer.hbs
+++ b/templates/footer.hbs
@@ -21,11 +21,11 @@
       <div class="four columns mt3 mt0-l">
         <h4>Social</h4>
         <div class="flex flex-row flex-wrap">
-          <a rel="me" href="https://social.rust-lang.org/@rust" target="_blank" rel="noopener" alt="mastodon link"><img src="{{root}}images/mastodon.svg" alt="mastodon logo" title="Mastodon"/></a>
-          <a href="https://twitter.com/rustlang" target="_blank" rel="noopener" alt="twitter link"><img src="{{root}}images/twitter.svg" alt="twitter logo" title="Twitter"/></a>
-          <a href="https://www.youtube.com/channel/UCaYhcUwRBNscFNUKTjgPFiA" target="_blank" rel="noopener" alt="youtube link"><img style="padding-top: 6px; padding-bottom:6px" src="{{root}}images/youtube.svg" alt="youtube logo" title="YouTube"/></a>
-          <a href="https://discord.gg/rust-lang" target="_blank" rel="noopener" alt="discord link"><img src="{{root}}images/discord.svg" alt="discord logo" title="Discord"/></a>
-          <a href="https://github.com/rust-lang" target="_blank" rel="noopener" alt="github link"><img src="{{root}}images/github.svg" alt="github logo" title="GitHub"/></a>
+          <a rel="me" href="https://social.rust-lang.org/@rust" target="_blank" rel="noopener" alt="mastodon link"><img decoding="async" loading="lazy" src="{{root}}images/mastodon.svg" alt="mastodon logo" title="Mastodon"/></a>
+          <a href="https://twitter.com/rustlang" target="_blank" rel="noopener" alt="twitter link"><img decoding="async" loading="lazy" src="{{root}}images/twitter.svg" alt="twitter logo" title="Twitter"/></a>
+          <a href="https://www.youtube.com/channel/UCaYhcUwRBNscFNUKTjgPFiA" target="_blank" rel="noopener" alt="youtube link"><img decoding="async" loading="lazy" style="padding-top: 6px; padding-bottom:6px" src="{{root}}images/youtube.svg" alt="youtube logo" title="YouTube"/></a>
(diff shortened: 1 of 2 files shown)


templates/headers.hbs
```diff
@@ -30,4 +30,7 @@
 <meta name="theme-color" content="#ffffff">
 
  <!-- atom -->
- <link type="application/atom+xml" rel="alternate" href="https://blog.rust-lang.org/{{blog.prefix}}feed.xml" title="{{blog.title}}" />
+<link type="application/atom+xml" rel="alternate" href="https://blog.rust-lang.org/{{blog.prefix}}feed.xml" title="{{blog.title}}" />
+
+ <!-- scripts -->
+<script defer src="{{root}}scripts/highlight.js"></script>
```
apiraino, line 36: If I remember correctly, one way to improve website loading times is to put `<script>` at the very end of the HTML, so I think this change here could theoretically lead to a perf. disadvantage compared to the previous solution

sanjaiyan-dev: Hi @apiraino, you are correct. Adding a script tag at the end of the body will have its own benefits, but it will start to fetch the script after parsing the HTML. However, with the `defer` attribute, we can achieve the same but with parallel fetching during HTML parsing, and it is executed at the end of the body.

Here is a small blog which explains the advantages of it: [Improving Third-Party Web Performance at The Telegraph](https://medium.com/the-telegraph-engineering/improving-third-party-web-performance-at-the-telegraph-a0a1000be5#4123)

sanjaiyan-dev: Google's Article -: https://web.dev/articles/efficiently-load-third-party-javascript#defer

templates/footer.hbs
```diff
@@ -21,11 +21,11 @@
...
+          <a rel="me" href="https://social.rust-lang.org/@rust" target="_blank" rel="noopener" alt="mastodon link"><img decoding="async" loading="lazy" src="{{root}}images/mastodon.svg" alt="mastodon logo" title="Mastodon"/></a>
+          <a href="https://twitter.com/rustlang" target="_blank" rel="noopener" alt="twitter link"><img decoding="async" loading="lazy" src="{{root}}images/twitter.svg" alt="twitter logo" title="Twitter"/></a>
+          <a href="https://www.youtube.com/channel/UCaYhcUwRBNscFNUKTjgPFiA" target="_blank" rel="noopener" alt="youtube link"><img decoding="async" loading="lazy" style="padding-top: 6px; padding-bottom:6px" src="{{root}}images/youtube.svg" alt="youtube logo" title="YouTube"/></a>
+          <a href="https://discord.gg/rust-lang" target="_blank" rel="noopener" alt="discord link"><img decoding="async" loading="lazy" src="{{root}}images/discord.svg" alt="discord logo" title="Discord"/></a>
+          <a href="https://github.com/rust-lang" target="_blank" rel="noopener" alt="github link"><img decoding="async" loading="lazy" src="{{root}}images/github.svg" alt="github logo" title="GitHub"/></a>
```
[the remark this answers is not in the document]
sanjaiyan-dev: I thought setting the image under the user viewport not to be fetched until it's visible, which can improve network usage on the client and lead to faster website loading. Users won't download unnecessary images at the bottom unless they scroll to that section, saving network usage and ensuring the website loads faster. Here's the Google article: [Browser-level image lazy loading](https://web.dev/articles/browser-level-image-lazy-loading#the_loading_attribute).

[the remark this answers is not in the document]
senekor: > Images with loading set to lazy will never be loaded if they do not intersect a visible part of an element, even if loading them would change that

That situation doesn't apply to us. These images to intersect with the visible part of an element (they are placed within the footer) even if their size is 0x0 in the unloaded state.

senekor (approved):
@sanjaiyan-dev It's been a while, but I tested the changes and will merge if you'd like to rebase this PR.

apiraino:
Do we have some numbers about the perf. diff (+/-) after this patch?

senekor:
No and I don't think there would be any if we actually measured. The images are just a few logos with a few kilobytes each. Internet connection would have to be really bad for that to be significant. I did verify that the images are in fact only loaded when you scroll to the bottom of the page.

I was leaning towards merging unless there is a reason not to, because the change is good practice on the face of it. Do you think we shouldn't merge unless there is a concrete benefit?

apiraino:
> I was leaning towards merging unless there is a reason not to, because the change is good practice on the face of it. Do you think we shouldn't merge unless there is a concrete benefit?

I'm mostly neutral about this patch. I was just interested in supporting the claimed perf. improvement. IIRC the browser developer tools allow restraining bandwidth also for this kind of testing (don't have time to investigate rn).

PR #1180, opened 2023-12-20
Announce `async fn` and return-position `impl Trait` in traits

tmandry:
@rust-lang/wg-async

### DIFF
--- a/posts/2023-12-21-async-fn-rpit-in-traits.md
+++ b/posts/2023-12-21-async-fn-rpit-in-traits.md
@@ -1,0 +1,229 @@
+---
+layout: post
+title: "Announcing `async fn` and return-position `impl Trait` in traits"
+author: Tyler Mandry
+team: The Async Working Group <https://www.rust-lang.org/governance/wgs/wg-async>
+---
+
+The Rust Async Working Group is excited to announce major progress towards our goal of enabling the use of `async fn` in traits. The next Rust release will include stable support for both `-> impl Trait` notation and `async fn` in traits.
+
+This is a big milestone, and we know many users will be itching to try these out in their own code. However, we are still missing some important features that many users need. Read on for an recommendations on when and how to use the stabilized features.
+
+## What's stabilizing
+
+Ever since [RFC #1522], Rust has allowed users to write `impl Trait` as the return type of functions. This means that the function returns "some type that implements `Trait`". This is commonly used to return iterators, closures, or other types that are complex or even impossible to write explicitly:
+
+[RFC #1522]: https://rust-lang.github.io/rfcs/1522-conservative-impl-trait.html
+
+```rust
+/// Given a list of players, return an iterator
+/// over their names.
+fn player_names(
+    players: &[Player]
+) -> impl Iterator<Item = &String> {
+    players
+        .iter()
+        .map(|p| &p.name)
+}
+```
+
+Starting in Rust 1.75, which hits stable next week, you can use **return-position `impl Trait` in traits** (RPITIT) and trait impls. For example, you could use this to write a trait that returns an iterator:
+
+```rust
+trait Container {
+    fn items(&self) -> impl Iterator<Item = Widget>;
+}
+
+impl Container for MyContainer {
+    fn items(&self) -> impl Iterator<Item = Widget> {
+        self.items.iter().cloned()
+    }
+}
+```
+
+So what does all of this have to do with async functions? Well, async functions are "just sugar" for a function that returns `-> impl Future`. Since that is now permitted in traits, **we also permit you to write traits that use `async fn`**.
+
+```rust
+trait HttpService {
+    async fn fetch(&self, url: Url) -> HtmlBody;
+//  ^^^^^^^^ desugars to:
+//  fn fetch(&self, url: Url) -> impl Future<Output = HtmlBody>;
+}
+```
+
+## Where the gaps lie
+
+### `-> impl Trait` in public traits
+
+The use of `-> impl Trait` is still discouraged for general use in **public** traits, for the reason that users can't use additional bounds on the return type. For example, there is no way to write this function:
+
+```rust!
+fn print_in_reverse(container: impl Container) {
+    for item in container.items().rev() {
+        // ERROR:                 ^^^
+        // the trait `DoubleEndedIterator`
+        // is not implemented for
+        // `impl Iterator<Item = Widget>`
+        eprintln!("{item}");
+    }
+}
+```
+
+In the future we plan to add a solution for this. For now, this feature is best used in internal traits or when you're confident your users won't need additional bounds like this. Otherwise, you should continue using associated types.
+
+### `async fn` in public traits
+
+Since `async fn` desugars to `-> impl Future`, the same limitations apply. If you use bare `async fn` in a public trait today, you'll see a warning. Thankfully, we have a solution that allows using `async fn` in public traits today!
+
+```
+warning: use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
+ --> src/lib.rs:7:5
+  |
+7 |     async fn fetch(&self, url: Url) -> HtmlBody;
+  |     ^^^^^
+  |
+help: you can desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
+  |
+7 -     async fn fetch(&self, url: Url) -> HtmlBody;
+7 +     fn fetch(&self, url: Url) -> impl std::future::Future<Output = HtmlBody> + Send;
+  |
+```
+
+Of particular interest to users of async are `Send` bounds on the returned future. Since users cannot add bounds later, the error message is saying that you as a trait author need to make a choice: Do you want your trait to work with multithreaded, work-stealing executors?
+
+For public traits, we recommend using the `trait-variant` proc macro to let your users choose. Add it to your project with `cargo add trait-variant`, then use it like so:
+
+```rust
+#[trait_variant::make(HttpService: Send)]
+pub trait LocalHttpService {
+    async fn fetch(&self, url: Url) -> HtmlBody;
+}
+```
+
+This creates *two* versions of your trait: `LocalHttpService` for single-threaded executors and `HttpService` for multithreaded work-stealing executors. Since we expect the latter to be used more commonly, it has the shorter name in this example. It has additional Send bounds:
+
+```rust
+pub trait HttpService: Send {
+    fn fetch(
+        &self,
+        url: Url,
+    ) -> impl Future<Output = HtmlBody> + Send;
+}
+```
+
+This macro works for async because `impl Future` rarely requires additional bounds other than Send, so we can set our users up for success. See [this blog post](https://blog.rust-lang.org/inside-rust/2023/05/03/stabilizing-async-fn-in-trait.html) for a more thorough explanation of the problem.[^cut-scope]
+
+[^cut-scope]: Note that we originally said we would solve the Send bound problem before shipping `async fn` in traits, but we decided to cut that from the scope and ship the `trait-variant` crate instead.
+
+### Dynamic dispatch
+
+Traits that use `-> impl Trait` and `async fn` are not object-safe, which means they lack support for dynamic dispatch. We plan to provide utilities that enable dynamic dispatch in an upcoming version of the `trait-variant` crate.
+
+## What we hope to do in the future
+
+In the future we would like to allow users to add their own bounds to `impl Trait` return types, which would make them more generally useful. It would also enable more advanced uses of `async fn`. The syntax might look something like this:
+
+```rust!
+trait HttpService = LocalHttpService<fetch(): Send> + Send;
+```
+
+Since these aliases won't require any support on the part of the trait author, it will technically make the Send variants of async traits unnecessary. Those are still a nice convenience for users, so we expect that most crates will continue to provide them.
+
+## Frequently asked questions
+
+### Is it okay to use `-> impl Trait` in traits?
+
+For private traits you can use `-> impl Trait` freely. For public traits, it's best to avoid them for now unless you can anticipate all the bounds your users might want (in which case you can use `#[trait_variant]`, as we do for async). We expect to lift this restriction in the future.
+
+### Should I still use the `#[async_trait]` macro?
+
+There are a couple of reasons you might need to continue using async-trait:
+
+* You want to support Rust versions older than 1.75.
+* You want dynamic dispatch.
+
+As stated above, we hope to enable dynamic dispatch in a future version of trait-variant.
+
+### Is it okay to use `async fn` in traits? What are the limitations?
+
+Assuming you don't need to use `#[async_trait]` for one of the reasons stated above, it's totally fine to use regular `async fn` in traits. Just remember to use `#[trait_variant]` if you want to support multithreaded runtimes.
+
+The biggest limitation is that a type must always decide if it implements the Send or non-Send version of a trait. It cannot implement the Send version *conditionally* on one of its generics. This can come up in the [middleware](https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md) pattern, for example, `RequestLimitingService<T>` that is HttpService if `T: HttpService`.
+
+### Why do I need `#[trait_variant]` and Send bounds?
+
+In simple cases you may find that your trait appears to work fine with a multithreaded executor. There are some patterns that just won't work, however. Consider the following:
+
+```rust!
+fn spawn_task(service: impl HttpService + 'static) {
+    tokio::spawn(async move {
+        let url = Url::from("https://rust-lang.org");
+        let _body = service.fetch(url).await;
+    });
+}
+```
+
+Without Send bounds on our trait, this would fail to compile with the error: "future cannot be sent between threads safely". By creating a variant of your trait with Send bounds, you avoid sending your users into this trap.
+
+### Can I mix async fn and impl trait?
+
+Yes, you can freely move between the `async fn` and `-> impl Future` spelling in your traits and impls. This is true even when one form has a Send bound.[^leakage] This makes the traits created by `trait_variant` nicer to use.
+
+```rust!
+trait HttpService: Send {
+    fn fetch(&self, url: Url)
+    -> impl Future<Output = HtmlBody> + Send;
+}
+
+impl HttpService for MyService {
+    async fn fetch(&self, url: Url) -> HtmlBody {
+        // This works, as long as `do_fetch(): Send`!
+        self.client.do_fetch(url).await.into_body()
+    }
+}
+```
+
+[^leakage]: This works because of *auto-trait leakage*, which allows knowledge of auto traits to "leak" outside of an impl signature that does not specify them.
+
+### Why don't these signatures use `impl Future + '_`?
+
+For `-> impl Trait` in traits we adopted the [2024 Capture Rules] early. This means that the `+ '_` you often see today is unnecessary in traits, because the return type is already assumed to capture input lifetimes. In the 2024 edition this rule will apply to all function signatures. See the linked RFC for more.
+
+[2024 Capture Rules]: https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html
+
+### Why am I getting a "refine" warning when I implement a trait with `-> impl Trait`?
+
+If your impl signature includes more detailed information than the trait itself, you'll [get a warning](https://play.rust-lang.org/?version=beta&mode=debug&edition=2021&gist=6248cfe0419a693d1331ef47c729d1fe):
+
+```rust!
+pub trait Foo {
+    fn foo(self) -> impl Debug;
+}
+
+impl Foo for u32 {
+    fn foo(self) -> String {
+//                  ^^^^^^
+//  warning: impl trait in impl method signature does not match trait method signature
+        self.to_string()
+    }
+}
+```
+
+The reason is that you may be leaking more details of your implementation than you meant to. For instance, should the following code compile?
+
+```rust!
+fn main() {
+    // Did the implementer mean to allow
+    // use of `Display`, or only `Debug` as
+    // the trait says?
+    println!("{}", 32.foo());
+}
+```
+
+Thanks to [refined trait implementations][3245] it does compile, but the compiler asks you to confirm your intent to refine the trait interface with `#[allow(refining_impl_trait)]` on the impl.
+
+[3245]: https://rust-lang.github.io/rfcs/3245-refined-impls.html
+
+## Conclusion
+
+The Async Working Group is very happy to end 2023 by announcing the completion of our primary goal for the year! Thank you to everyone who participated in design, implementation, and stabilization discussions. Thanks also to the users of Async Rust who have given great feedback over the years. We're looking forward to seeing what you build, and to delivering continued improvements in the year to come.


posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+        eprintln!("{item}");
+    }
+}
+```
+
+In the future we plan to add a solution for this. For now, this feature is best used in internal traits or when you're confident your users won't need additional bounds like this. Otherwise, you should continue using associated types.
+
+### `async fn` in public traits
+
+Since `async fn` desugars to `-> impl Future`, the same limitations apply. If you use bare `async fn` in a public trait today, you'll see a warning. Thankfully, we have a solution that allows using `async fn` in public traits today!
```
Jules-Bertholet, line 76: "Thankfully, we have a solution that allows using `async fn` in public traits today!" should probably be moved down to where `trait-variant` is introduced

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+help: you can desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
+  |
+7 -     async fn fetch(&self, url: Url) -> HtmlBody;
+7 +     fn fetch(&self, url: Url) -> impl std::future::Future<Output = HtmlBody> + Send;
+  |
+```
+
+Of particular interest to users of async are `Send` bounds on the returned future. Since users cannot add bounds later, the error message is saying that you as a trait author need to make a choice: Do you want your trait to work with multithreaded, work-stealing executors?
+
+For public traits, we recommend using the `trait-variant` proc macro to let your users choose. Add it to your project with `cargo add trait-variant`, then use it like so:
```
Jules-Bertholet, line 94:
```suggestion
For public traits, we recommend using the [`trait-variant`](https://crates.io/crates/trait-variant) proc macro to let your users choose. Add it to your project with `cargo add trait-variant`, then use it like so:
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
+---
+layout: post
+title: "Announcing `async fn` and return-position `impl Trait` in traits"
+author: Tyler Mandry
+team: The Async Working Group <https://www.rust-lang.org/governance/wgs/wg-async>
+---
+
+The Rust Async Working Group is excited to announce major progress towards our goal of enabling the use of `async fn` in traits. The next Rust release will include stable support for both `-> impl Trait` notation and `async fn` in traits.
```
crlf0710, line 8: Maybe indicate the version number more explicitly? When this post is translated into other languages, there's a delay in time, and people might get confused about which release this is...

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+This is a big milestone, and we know many users will be itching to try these out in their own code. However, we are still missing some important features that many users need. Read on for an recommendations on when and how to use the stabilized features.
+
+## What's stabilizing
+
+Ever since [RFC #1522], Rust has allowed users to write `impl Trait` as the return type of functions. This means that the function returns "some type that implements `Trait`". This is commonly used to return iterators, closures, or other types that are complex or even impossible to write explicitly:
```
traviscross, line 14:
```suggestion
Ever since the stabilization of [RFC #1522] in Rust 1.26, Rust has allowed users to write `impl Trait` as the return type of functions (often called "RPIT"). This means that the function returns "some type that implements `Trait`". This is the only way to return types that cannot be named, such as futures and closures, without boxing. It's also used when an API author wants to return a type with a well defined interface, such as an iterator, without making a stability commitment to the exact type (such as, e.g., `Zip<Once<T>, Once<T>>`).
```

The changes here:

- It may be better to note the release in which RPIT stabilized, as in some sense that's when it became available.
- It's probably good to mention "RPIT" here somewhere, as many people do know and may scan for that term, and it helps connect this to "RPITIT" below.
- The framing of RPIT as being for "complex" types has proved to be misleading (this language originates in the RFC).  The choice of whether to return an opaque type (when the type *could* instead be returned concretely) hinges on whether you want that type to be *opaque*, with all that implies for the limitations of how it can be used by the caller and what it means for the stability commitment of the callee.  We don't want to suggest to people that they should switch to RPIT just because the type has become "complex" (plenty of complex type should be returned concretely).  So it may be better to frame the choice to use RPIT as being for when 1) the type cannot be named or when 2) we want to not make a SemVer commitment to the concrete type.

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+fn player_names(
+    players: &[Player]
+) -> impl Iterator<Item = &String> {
+    players
+        .iter()
+        .map(|p| &p.name)
+}
+```
+
+Starting in Rust 1.75, which hits stable next week, you can use **return-position `impl Trait` in traits** (RPITIT) and trait impls. For example, you could use this to write a trait that returns an iterator:
```
traviscross, line 30:
```suggestion
Starting in Rust 1.75, which hits stable next week, you can use **return-position `impl Trait` in trait** (RPITIT) definitions and in trait impls. For example, you could use this to write a trait that returns an iterator:
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+So what does all of this have to do with async functions? Well, async functions are "just sugar" for a function that returns `-> impl Future`. Since that is now permitted in traits, **we also permit you to write traits that use `async fn`**.
+
+```rust
+trait HttpService {
+    async fn fetch(&self, url: Url) -> HtmlBody;
+//  ^^^^^^^^ desugars to:
+//  fn fetch(&self, url: Url) -> impl Future<Output = HtmlBody>;
+}
+```
+
```
traviscross, line 53:
```suggestion

(See the FAQs below for why this desugaring is correct.)

```

It's probably worth a forward reference here so that readers don't start to wonder (and perhaps start to comment on) whether we've made a mistake.

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+## Where the gaps lie
+
+### `-> impl Trait` in public traits
+
+The use of `-> impl Trait` is still discouraged for general use in **public** traits, for the reason that users can't use additional bounds on the return type. For example, there is no way to write this function:
```
traviscross, line 58:
```suggestion
The use of `-> impl Trait` is still discouraged for general use in **public** traits for the reason that implementors of the trait cannot provide additional bounds on the opaque return types that can be relied on by users. For example, there is no way to write this function:
```

We're making a really subtle point here, and we do want to keep it concise, but it perhaps may be good to spell this out a bit further.

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+        // ERROR:                 ^^^
+        // the trait `DoubleEndedIterator`
+        // is not implemented for
+        // `impl Iterator<Item = Widget>`
+        eprintln!("{item}");
+    }
+}
+```
+
+In the future we plan to add a solution for this. For now, this feature is best used in internal traits or when you're confident your users won't need additional bounds like this. Otherwise, you should continue using associated types.
```
traviscross, line 72:
```suggestion
In the future we plan to add a solution for this. For now, this feature is best used in internal traits or when you're confident your users won't need additional bounds like this. Otherwise, you should continue using associated types where possible (note that associated types cannot yet be used when returning types that cannot be named).
```

We probably shouldn't leave people with the impression that it's always possible to use associated types instead of RPITIT.  It will be, once we stabilize ATPIT, but until then, it's not.

posts/2023-12-21-async-fn-rpit-in-traits.md
traviscross:
```suggestion
For public traits, we recommend using the [`trait_variant::make`](https://docs.rs/trait-variant/latest/trait_variant/attr.make.html) proc macro from the [`trait-variant`](https://crates.io/crates/trait-variant) crate to let your users choose. Add it to your project with `cargo add trait-variant`, then use it like so:
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+For public traits, we recommend using the `trait-variant` proc macro to let your users choose. Add it to your project with `cargo add trait-variant`, then use it like so:
+
+```rust
+#[trait_variant::make(HttpService: Send)]
+pub trait LocalHttpService {
+    async fn fetch(&self, url: Url) -> HtmlBody;
+}
+```
+
+This creates *two* versions of your trait: `LocalHttpService` for single-threaded executors and `HttpService` for multithreaded work-stealing executors. Since we expect the latter to be used more commonly, it has the shorter name in this example. It has additional Send bounds:
```
traviscross, line 103:
```suggestion
This creates *two* versions of your trait: `LocalHttpService` for single-threaded executors and `HttpService` for multithreaded work-stealing executors. Since we expect the latter to be used more commonly, it has the shorter name in this example. It has additional `Send` bounds:
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+```rust
+pub trait HttpService: Send {
+    fn fetch(
+        &self,
+        url: Url,
+    ) -> impl Future<Output = HtmlBody> + Send;
+}
+```
+
+This macro works for async because `impl Future` rarely requires additional bounds other than Send, so we can set our users up for success. See [this blog post](https://blog.rust-lang.org/inside-rust/2023/05/03/stabilizing-async-fn-in-trait.html) for a more thorough explanation of the problem.[^cut-scope]
```
traviscross, line 114:
```suggestion
This macro works for async because `impl Future` rarely requires additional bounds other than `Send`, so we can set our users up for success. See [this blog post](https://blog.rust-lang.org/inside-rust/2023/05/03/stabilizing-async-fn-in-trait.html) for a more thorough explanation of the problem.[^cut-scope]
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+[^cut-scope]: Note that we originally said we would solve the Send bound problem before shipping `async fn` in traits, but we decided to cut that from the scope and ship the `trait-variant` crate instead.
```
traviscross, line 116:
```suggestion
[^cut-scope]: Note that we originally said we would solve the `Send` bound problem before shipping `async fn` in traits, but we decided to cut that from the scope and ship the `trait-variant` crate instead.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+## What we hope to do in the future
+
+In the future we would like to allow users to add their own bounds to `impl Trait` return types, which would make them more generally useful. It would also enable more advanced uses of `async fn`. The syntax might look something like this:
+
+```rust!
+trait HttpService = LocalHttpService<fetch(): Send> + Send;
+```
+
+Since these aliases won't require any support on the part of the trait author, it will technically make the Send variants of async traits unnecessary. Those are still a nice convenience for users, so we expect that most crates will continue to provide them.
```
traviscross, line 130:
```suggestion
Since these aliases won't require any support on the part of the trait author, it will technically make the `Send` variants of async traits unnecessary. However, those variants will still be a nice convenience for users, so we expect that most crates will continue to provide them.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Frequently asked questions
+
+### Is it okay to use `-> impl Trait` in traits?
+
+For private traits you can use `-> impl Trait` freely. For public traits, it's best to avoid them for now unless you can anticipate all the bounds your users might want (in which case you can use `#[trait_variant]`, as we do for async). We expect to lift this restriction in the future.
```
traviscross, line 136:
```suggestion
For private traits you can use `-> impl Trait` freely. For public traits, it's best to avoid them for now unless you can anticipate all the bounds your users might want (in which case you can use `#[trait_variant::make]`, as we do for async). We expect to lift this restriction in the future.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Should I still use the `#[async_trait]` macro?
+
+There are a couple of reasons you might need to continue using async-trait:
+
+* You want to support Rust versions older than 1.75.
+* You want dynamic dispatch.
+
+As stated above, we hope to enable dynamic dispatch in a future version of trait-variant.
```
traviscross, line 145:
```suggestion
As stated above, we hope to enable dynamic dispatch in a future version of the `trait-variant` crate.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Is it okay to use `async fn` in traits? What are the limitations?
+
+Assuming you don't need to use `#[async_trait]` for one of the reasons stated above, it's totally fine to use regular `async fn` in traits. Just remember to use `#[trait_variant]` if you want to support multithreaded runtimes.
```
traviscross, line 149:
```suggestion
Assuming you don't need to use `#[async_trait]` for one of the reasons stated above, it's totally fine to use regular `async fn` in traits. Just remember to use `#[trait_variant::make]` if you want to support multithreaded runtimes.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+The biggest limitation is that a type must always decide if it implements the Send or non-Send version of a trait. It cannot implement the Send version *conditionally* on one of its generics. This can come up in the [middleware](https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md) pattern, for example, `RequestLimitingService<T>` that is HttpService if `T: HttpService`.
```
traviscross, line 151:
```suggestion
The biggest limitation is that a type must always decide if it implements the `Send` or non-`Send` version of a trait. It cannot implement the `Send` version *conditionally* on one of its generics. This can come up in the [middleware](https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md) pattern, for example, `RequestLimitingService<T>` that is HttpService if `T: HttpService`.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Why do I need `#[trait_variant]` and Send bounds?
```
traviscross, line 153:
```suggestion
### Why do I need `#[trait_variant::make]` and `Send` bounds?
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+```rust!
+fn spawn_task(service: impl HttpService + 'static) {
+    tokio::spawn(async move {
+        let url = Url::from("https://rust-lang.org");
+        let _body = service.fetch(url).await;
+    });
+}
+```
+
+Without Send bounds on our trait, this would fail to compile with the error: "future cannot be sent between threads safely". By creating a variant of your trait with Send bounds, you avoid sending your users into this trap.
```
traviscross, line 166:
```suggestion
Without `Send` bounds on our trait, this would fail to compile with the error: "future cannot be sent between threads safely". By creating a variant of your trait with `Send` bounds, you avoid sending your users into this trap.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Can I mix async fn and impl trait?
+
+Yes, you can freely move between the `async fn` and `-> impl Future` spelling in your traits and impls. This is true even when one form has a Send bound.[^leakage] This makes the traits created by `trait_variant` nicer to use.
```
traviscross, line 170:
```suggestion
Yes, you can freely move between the `async fn` and `-> impl Future` spelling in your traits and impls. This is true even when one form has a `Send` bound.[^leakage] This makes the traits created by `trait_variant` nicer to use.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+impl HttpService for MyService {
+    async fn fetch(&self, url: Url) -> HtmlBody {
+        // This works, as long as `do_fetch(): Send`!
+        self.client.do_fetch(url).await.into_body()
+    }
+}
+```
+
+[^leakage]: This works because of *auto-trait leakage*, which allows knowledge of auto traits to "leak" outside of an impl signature that does not specify them.
```
traviscross, line 186:
```suggestion
[^leakage]: This works because of *auto-trait leakage*, which allows knowledge of auto traits to "leak" from an item whose signature that does not specify them.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+}
+```
+
+Thanks to [refined trait implementations][3245] it does compile, but the compiler asks you to confirm your intent to refine the trait interface with `#[allow(refining_impl_trait)]` on the impl.
+
+[3245]: https://rust-lang.github.io/rfcs/3245-refined-impls.html
+
+## Conclusion
+
+The Async Working Group is very happy to end 2023 by announcing the completion of our primary goal for the year! Thank you to everyone who participated in design, implementation, and stabilization discussions. Thanks also to the users of Async Rust who have given great feedback over the years. We're looking forward to seeing what you build, and to delivering continued improvements in the year to come.
```
traviscross, line 229:
```suggestion
The Async Working Group is excited to end 2023 by announcing the completion of our primary goal for the year! Thank you to everyone who helpfully participated in design, implementation, and stabilization discussions. Thanks also to the users of async Rust who have given great feedback over the years. We're looking forward to seeing what you build, and to delivering continued improvements in the years to come.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+Assuming you don't need to use `#[async_trait]` for one of the reasons stated above, it's totally fine to use regular `async fn` in traits. Just remember to use `#[trait_variant]` if you want to support multithreaded runtimes.
+
+The biggest limitation is that a type must always decide if it implements the Send or non-Send version of a trait. It cannot implement the Send version *conditionally* on one of its generics. This can come up in the [middleware](https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md) pattern, for example, `RequestLimitingService<T>` that is HttpService if `T: HttpService`.
+
+### Why do I need `#[trait_variant]` and Send bounds?
+
+In simple cases you may find that your trait appears to work fine with a multithreaded executor. There are some patterns that just won't work, however. Consider the following:
+
+```rust!
```
yoshuawuyts, line 157:
```suggestion
```rust
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+}
+```
+
+Without Send bounds on our trait, this would fail to compile with the error: "future cannot be sent between threads safely". By creating a variant of your trait with Send bounds, you avoid sending your users into this trap.
+
+### Can I mix async fn and impl trait?
+
+Yes, you can freely move between the `async fn` and `-> impl Future` spelling in your traits and impls. This is true even when one form has a Send bound.[^leakage] This makes the traits created by `trait_variant` nicer to use.
+
+```rust!
```
yoshuawuyts, line 172:
```suggestion
```rust
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+For `-> impl Trait` in traits we adopted the [2024 Capture Rules] early. This means that the `+ '_` you often see today is unnecessary in traits, because the return type is already assumed to capture input lifetimes. In the 2024 edition this rule will apply to all function signatures. See the linked RFC for more.
+
+[2024 Capture Rules]: https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html
+
+### Why am I getting a "refine" warning when I implement a trait with `-> impl Trait`?
+
+If your impl signature includes more detailed information than the trait itself, you'll [get a warning](https://play.rust-lang.org/?version=beta&mode=debug&edition=2021&gist=6248cfe0419a693d1331ef47c729d1fe):
+
+```rust!
```
yoshuawuyts, line 198:
```suggestion
```rust
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+//                  ^^^^^^
+//  warning: impl trait in impl method signature does not match trait method signature
+        self.to_string()
+    }
+}
+```
+
+The reason is that you may be leaking more details of your implementation than you meant to. For instance, should the following code compile?
+
+```rust!
```
yoshuawuyts, line 214:
```suggestion
```rust
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+}
+```
+
+## Where the gaps lie
+
+### `-> impl Trait` in public traits
+
+The use of `-> impl Trait` is still discouraged for general use in **public** traits, for the reason that users can't use additional bounds on the return type. For example, there is no way to write this function:
+
+```rust!
```
yoshuawuyts, line 60:
```suggestion
```rust
```

This doesn't render correctly in the GitHub preview. I'm not sure whether the actual blog post can parse this though, but just for good measure it's probably worth changing

posts/2023-12-21-async-fn-rpit-in-traits.md
yoshuawuyts, line 103: Isn't this the inverse of how the language handles this? In order to guarantee `+ Send` it will need to be annotated, while `?Send` is the default? Or am I confusing things here?

If so, we should probably elaborate here why that choice is inverted to set expectations. Even if it's just a footnote.

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+This macro works for async because `impl Future` rarely requires additional bounds other than Send, so we can set our users up for success. See [this blog post](https://blog.rust-lang.org/inside-rust/2023/05/03/stabilizing-async-fn-in-trait.html) for a more thorough explanation of the problem.[^cut-scope]
+
+[^cut-scope]: Note that we originally said we would solve the Send bound problem before shipping `async fn` in traits, but we decided to cut that from the scope and ship the `trait-variant` crate instead.
+
+### Dynamic dispatch
+
+Traits that use `-> impl Trait` and `async fn` are not object-safe, which means they lack support for dynamic dispatch. We plan to provide utilities that enable dynamic dispatch in an upcoming version of the `trait-variant` crate.
+
+## What we hope to do in the future
```
yoshuawuyts, line 122: AFIT and async Drop almost certainly have interactions too. We may want to mention something like:

> ## New core language features
>
> Stabilizing AFIT is also a key component to develop other core async language features like async closures, and async destructors. With AFIT now implemented and stable, we'll be looking to see how we can use it to enable other language features.

This doesn't have to be verbatim, but something along those lines. I think it's valuable to explain that AFIT is a key feature for async which enables other things people are looking forward to as well.

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,229 @@
...
+
+In the future we would like to allow users to add their own bounds to `impl Trait` return types, which would make them more generally useful. It would also enable more advanced uses of `async fn`. The syntax might look something like this:
+
+```rust!
```
yoshuawuyts, line 126:
```suggestion
```rust
```

posts/2023-12-21-async-fn-rpit-in-traits.md
nikomatsakis:
```suggestion
For public traits, we recommend using the [`trait_variant::make`](https://docs.rs/trait-variant/latest/trait_variant/attr.make.html) proc macro to let your users choose. This proc macro is part of the [`trait-variant`](https://crates.io/crates/trait-variant), published by the rust-lang org. Add it to your project with `cargo add trait-variant`, then use it like so:
```

I think we should make clear this is not a "third party" crate somehow.

nikomatsakis (approved):
This turned out really nice @tmandry ! Great work.

traviscross (approved):
As @nikomatsakis also mentioned, overall this looks really great, and it will be exciting to get this out with RPITIT and AFIT.

posts/2023-12-21-async-fn-rpit-in-traits.md
tmandry: I would say using RPIT for complex types is a common and acceptable practice – just not in public APIs.

posts/2023-12-21-async-fn-rpit-in-traits.md
tmandry: I am not sure I agree that this clarifies it. Implementers can provide additional bounds, either through refinement or auto trait leakage, and users can rely on them _if_ they know the type being used that implements the trait. The issue is that generic code which does not know the type cannot state additional requirements which it can then rely on.

posts/2023-12-21-async-fn-rpit-in-traits.md
tmandry: The paragraph does say:

> Since we expect the latter to be used more commonly, it has the shorter name in this example.

How would you suggest clarifying further? Or are you referring to something other than the name?

tmandry: I intentionally left some of these off to reduce visual noise. The Rust blog puts a border and different background color around each one, so IMO they're best used for emphasis or longer code snippets.

posts/2023-12-21-async-fn-rpit-in-traits.md
tmandry: Thanks, I went with a footnote.

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,233 @@
...
+        // ERROR:                 ^^^
+        // the trait `DoubleEndedIterator`
+        // is not implemented for
+        // `impl Iterator<Item = Widget>`
+        eprintln!("{item}");
+    }
+}
+```
+
+In the future we plan to add a solution for this. For now, this feature is best used in internal traits or when you're confident your users won't need additional bounds like this. Otherwise, you should continue using associated types where possible (note that associated types cannot yet be used when returning types that cannot be named).
```
slanterns, line 74: Maybe it's better to leave a link to the tracking issue of ATPIT / TAIT?

tmandry:
cc @Mark-Simulacrum, since this will be linked from the release blog post

posts/2023-12-21-async-fn-rpit-in-traits.md
tmandry: Made some clarifications.

traviscross: Fair enough.  Earlier drafts of my suggestion did get into this but were too long.  I agree it's OK as it is.

It's a subtle point and I hope that people get what we mean.

traviscross: We were apparently typing at the same time, and GH didn't show me the clarifications prior to my message above.  The clarifications look good.

posts/2023-12-21-async-fn-rpit-in-traits.md
traviscross: Got it, sounds good.

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,235 @@
+---
+layout: post
+title: "Announcing `async fn` and return-position `impl Trait` in traits"
+author: Tyler Mandry
+team: The Async Working Group <https://www.rust-lang.org/governance/wgs/wg-async>
+---
+
+The Rust Async Working Group is excited to announce major progress towards our goal of enabling the use of `async fn` in traits. Rust 1.75, which hits stable next week, will include support for both `-> impl Trait` notation and `async fn` in traits.
+
+This is a big milestone, and we know many users will be itching to try these out in their own code. However, we are still missing some important features that many users need. Read on for an recommendations on when and how to use the stabilized features.
```
traviscross, line 10:
```suggestion
This is a big milestone, and we know many users will be itching to try these out in their own code. However, we are still missing some important features that many users need. Read on for recommendations on when and how to use the stabilized features.
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,235 @@
...
+fn player_names(
+    players: &[Player]
+) -> impl Iterator<Item = &String> {
+    players
+        .iter()
+        .map(|p| &p.name)
+}
+```
+
+Starting in Rust 1.75, you can use **return-position `impl Trait` in trait** (RPITIT) definitions and in trait impls. For example, you could use this to write a trait that returns an iterator:
```
traviscross, line 30:
```suggestion
Starting in Rust 1.75, you can use **return-position `impl Trait` in trait** (RPITIT) definitions and in trait impls. For example, you could use this to write a trait method that returns an iterator:
```

posts/2023-12-21-async-fn-rpit-in-traits.md
```diff
@@ -0,0 +1,237 @@
...
+}
+
+impl Container for MyContainer {
+    fn items(&self) -> impl Iterator<Item = Widget> {
+        self.items.iter().cloned()
+    }
+}
+```
+
+So what does all of this have to do with async functions? Well, async functions are "just sugar" for a function that returns `-> impl Future`. Since that is now permitted in traits, **we also permit you to write traits that use `async fn`**.[^desugar-correct]
```
traviscross, line 44:
```suggestion
So what does all of this have to do with async functions? Well, async functions are "just sugar" for functions that returns `-> impl Future`. Since these are now permitted in traits, **we also permit you to write traits that use `async fn`**.[^desugar-correct]
```

posts/2023-12-21-async-fn-rpit-in-traits.md
tmandry: I added a paragraph about future goals that mentions this.

traviscross (approved):

posts/2023-12-21-async-fn-rpit-in-traits.md
tmandry: Resolving for now to unblock merging.

tmandry:
Thanks for the feedback, everyone! I think this is ready to merge.

PR #1210, opened 2024-01-11
Fix clippy warnings

Turbo87:
Together with #1208 and #1209, this fixes the remaining clippy warnings that prevent us from introducing a `cargo clippy` run to the CI workflow.

### DIFF
--- a/src/blogs.rs
+++ b/src/blogs.rs
@@ -154,7 +154,7 @@ fn load_recursive(
     Ok(())
 }
 
-fn add_postfix_slash<S>(path: &PathBuf, serializer: S) -> Result<S::Ok, S::Error>
+fn add_postfix_slash<S>(path: &Path, serializer: S) -> Result<S::Ok, S::Error>
 where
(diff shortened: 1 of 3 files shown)


rylev (approved):

Mark-Simulacrum:
I don't mind the warnings showing up, but we're not planning to gate CI on clippy, right? That will potentially harm our ability to deploy release blog posts if there's new warnings in the new release, which isn't great. (Or typo fixes post merge depending on when exactly we enforce).

Manishearth (approved):

Turbo87:
> I don't mind the warnings showing up, but we're not planning to gate CI on clippy, right?

I was planning on doing that, but with a fixed Rust version that renovatebot updates whenever there is a new stable release. I'm not a big fan of having moving targets on CI, precisely because I want to avoid a broken CI for reasons unrelated to a PR.

Mark-Simulacrum:
Sounds good!

PR #1226, opened 2024-01-23
Fix some typos in This Development-cycle in Cargo: 1.76

### LINKED ISSUE #1185: Several typos in This Development-cycle in Cargo: 1.76
https://github.com/rust-lang/blog.rust-lang.org/blob/ebc2dbb305b8b897f664e090e6434b1fb6e8f7d7/posts/inside-rust/2024-01-03-this-development-cycle-in-cargo-1-76.md?plain=1#L28C1-L28C88

I have no idea what that line is supposed to say.

scimas:
Fixes #1185.

The "year" is just an assumption.

There are some minor consistency issues too, like the use of colons (`:`) at the beginning of lists, ending list items with periods, that could be addressed for future posts by the post writers themselves (through something like a style guide perhaps?).


ehuss (approved):
Thanks!

PR #1229, opened 2024-01-23
CI: Add `lint` job

Turbo87:
as discussed in https://github.com/rust-lang/blog.rust-lang.org/pull/1210 :)

r? @rust-lang/website


rylev (approved):
Looks good to me, but I'd want to check that @Mark-Simulacrum is ok with this solution.

PR #1236, opened 2024-01-31
Add "crates.io: API status code changes" blog post

Turbo87:
Bildschirmfoto 2024-01-31 um 10 51 23

r? @rust-lang/crates-io


posts/2024-02-05-crates-io-status-codes.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: post
+title: "crates.io: API status code changes"
+author: Tobias Bieniek
+team: the crates.io team <https://www.rust-lang.org/governance/teams/crates-io>
+---
+
+When cargo and crates.io were first implemented, [HTTP response status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) were ignored. cargo assumed that crates.io would always respond with a "[200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)" and any error messages are detected from the response JSON body. When crates.io responded with a different status code then cargo was showing the raw JSON body instead of a regular error message:
```
carols10cents, line 8: I think we should captialize Cargo:

```suggestion
When Cargo and crates.io were first implemented, [HTTP response status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) were ignored. Cargo assumed that crates.io would always respond with a "[200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)" and any error messages are detected from the response JSON body. When crates.io responded with a different status code then Cargo was showing the raw JSON body instead of a regular error message:
```

carols10cents, line 8: > When cargo and crates.io were first implemented, [HTTP response status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) were ignored.

This is a bit uncharitable and overly blame-ey IMO-- there were absolutely places that the API *did* use semantically correct HTTP status codes, just not everywhere. I think it'd be nicer to say something like: 

> In the rush leading up to Rust 1.0's release, with the desire to also have a tool to manage dependencies and a registry where people could share code, there were places in the API between Cargo and crates.io that didn't return the correct [HTTP response status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status). Then Rust's stability guarantees around backwards compatibility made this nontrivial to fix, as we wanted older versions of Cargo to continue working with the current crates.io API.

posts/2024-02-05-crates-io-status-codes.md
```diff
@@ -0,0 +1,35 @@
...
+headers:
+    HTTP/1.1 400 Bad Request
+    Content-Type: application/json; charset=utf-8
+    Content-Length: 171
+
+body:
+{"errors":[{"detail":"missing or empty metadata fields: description, license. Please see https://doc.rust-lang.org/cargo/reference/manifest.html for how to upload metadata"}]}
+```
+
+This was improved in [cargo#6771](https://github.com/rust-lang/cargo/pull/6771), which was released in cargo 1.34 (mid-2019). Since then, cargo has supported [4xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) and [5xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses) status codes too and extracts the error message from the JSON response, if available.
```
carols10cents, line 21:
```suggestion
This was improved in [Cargo pull request #6771](https://github.com/rust-lang/cargo/pull/6771), which was released in Cargo 1.34 (mid-2019). Since then, Cargo has supported receiving [4xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) and [5xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses) status codes too and extracts the error message from the JSON response, if available.
```

posts/2024-02-05-crates-io-status-codes.md
```diff
@@ -0,0 +1,35 @@
...
+
+The crates.io team has decided to phase out these "200 OK" response status codes for error responses to make our API a little more intuitive. Since this can be seen as a breaking change to our API we decided to announce this change publicly, including a cut-off date. On **2024-03-04** we will switch the API from the previous "200 OK" behavior to the new 4xx/5xx behavior, for the endpoints that are used by `cargo`:
```
carols10cents, line 23: Are there any plans to do any "scream tests" before the cutoff? That is, make the change temporarily and see if anyone screams at us? Do we think we need that?

posts/2024-02-05-crates-io-status-codes.md
```diff
@@ -0,0 +1,35 @@
...
+- `PUT /api/v1/crates/new`
+- `PUT /api/v1/crates/:crate/:version/yank`
+- `DELETE /api/v1/crates/:crate/:version/unyank`
+- `GET /api/v1/crates/:crate/owners`
+- `PUT /api/v1/crates/:crate/owners`
+- `DELETE /api/v1/crates/:crate/owners`
+
+All other endpoints have already been using regular HTTP status codes for some time.
+
+**cargo 1.33 and below will keep working after this change**, but will show the raw JSON response body instead of a regular error message. However, we recommend upgrading to a newer version of cargo to get the improved error messages and all the other nice things that the cargo team has built since then.
```
carols10cents, line 35: I feel like this is important justification for why we feel safe making one cutoff date: old Cargo versions won't be _broken_, they'll just display their errors in a nicer way. I would put this up in the paragraph before the list of API endpoints, so something like:

>  On **2024-03-04** we will switch the API from the previous "200 OK" behavior to the new 4xx/5xx behavior. **Cargo 1.33 and below will keep working after this change**, but will show the raw JSON body instead of a clearer error message. We feel confident that this degraded error message display will not affect very many users.
> 
> The full list of API endpoints affected are:
>
> * [list]
>
> All other endpoints have already been using regular HTTP status codes for some time.
>
> If you are still using Cargo 1.33 or older, we recommend upgrading to a newer version to get the improved error messages and all the other nice things that the Cargo team has built since then.

carols10cents (commented):
Some suggestions :)

posts/2024-02-05-crates-io-status-codes.md
Turbo87: @rust-lang/cargo any thoughts? I tend to use the casing of the CLI tool (aka. lowercase) but if the official spelling is uppercase let me know :)

Turbo87: > This is a bit uncharitable and overly blame-ey

I was struggling to find a way to express this in a blame-free way, but looks like I did not succeed 😅 

happy to rephrase it :)

posts/2024-02-05-crates-io-status-codes.md
Turbo87: > Are there any plans to do any "scream tests" before the cutoff?

I haven't planned anything like that (yet).

> Do we think we need that?

IMHO it seems unlikely to actually cause any issues. do you think we need it?

posts/2024-02-05-crates-io-status-codes.md
LawnGnome: I don't have strong feelings on the exact wording, but the tense wanders a bit in the original paragraph, and the suggested version doesn't flow into the block after.

Reusing most of Carol's suggested wording but also still using this as an intro to the following output sample, how about:

> Cargo and crates.io were developed in the rush leading up to the Rust 1.0 release to fill the dual needs for a tool to manage dependencies and a registry that people could use to share code. This rapid work resulted in these tools being connected with an API that initially didn't return the correct [HTTP response status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status). After the Rust 1.0 release, Rust's stability guarantees around backward compatibility made this non-trivial to fix, as we wanted older versions of Cargo to continue working with the current crates.io API.
>
> When an old version of Cargo receives a non-"[200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)" response, it displays the raw JSON body like this:

posts/2024-02-05-crates-io-status-codes.md
LawnGnome: I don't think we need the "too" in there, so I'd slightly tweak Carol's suggestion to:

```suggestion
This was improved in [Cargo pull request #6771](https://github.com/rust-lang/cargo/pull/6771), which was released in Cargo 1.34 (mid-2019). Since then, Cargo has supported receiving [4xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) and [5xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses) status codes and extracts the error message from the JSON response, if available.
```

LawnGnome (commented):
I'm basically in agreement with @carols10cents here, but I can't resist a good bikeshed. Sorry.

LawnGnome (approved):

posts/2024-02-05-crates-io-status-codes.md
```diff
@@ -0,0 +1,39 @@
...
+    Content-Type: application/json; charset=utf-8
+    Content-Length: 171
+
+body:
+{"errors":[{"detail":"missing or empty metadata fields: description, license. Please see https://doc.rust-lang.org/cargo/reference/manifest.html for how to upload metadata"}]}
+```
+
+This was improved in pull request [#6771](https://github.com/rust-lang/cargo/pull/6771), which was released in Cargo 1.34 (mid-2019). Since then, Cargo has supported receiving [4xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) and [5xx](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses) status codes too and extracts the error message from the JSON response, if available.
+
+On **2024-03-04** we will switch the API from returning "200 OK" status codes for errors to the new 4xx/5xx behavior. **Cargo 1.33 and below will keep working after this change**, but will show the raw JSON body instead of a nicely formatted error message. We feel confident that this degraded error message display will not affect very many users.
```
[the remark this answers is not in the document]
Turbo87: good idea. I've added a sentence with a link to the dashboard :)

PR #1243, opened 2024-02-07
Rust 1.76.0 announcement

cuviper:
cc @rust-lang/release 
r? @Mark-Simulacrum

### DIFF
--- a/posts/2024-02-08-Rust-1.76.0.md
+++ b/posts/2024-02-08-Rust-1.76.0.md
@@ -1,0 +1,72 @@
+---
+layout: post
+title: "Announcing Rust 1.76.0"
+author: The Rust Release Team
+release: true
+---
+
+The Rust team is happy to announce a new version of Rust, 1.76.0. Rust is a programming language empowering everyone to build reliable and efficient software.
+
+If you have a previous version of Rust installed via rustup, you can get 1.76.0 with:
+
+```console
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.76.0](https://doc.rust-lang.org/nightly/releases.html#version-1760-2024-02-08).
+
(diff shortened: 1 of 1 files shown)


posts/2024-02-08-Rust-1.76.0.md
```diff
@@ -0,0 +1,72 @@
...
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.76.0 stable
+
+This release is relatively minor, but as always, even incremental improvements lead to a greater whole. A few of those changes are highlighted in this post, and others may yet fill more niche needs.
+
+### ABI compatibility updates
+
+A new [ABI Compatibility](https://doc.rust-lang.org/stable/std/primitive.fn.html#abi-compatibility) section in the function pointer documentation describes what it means for function signatures to be ABI-compatible. A large part of that is the compatibility of argument types and return types, with a list of those that are currently considered compatible in Rust. For the most part, this documentation is not adding any new guarantees, only describing the existing state of compatibility.
```
Mark-Simulacrum, line 26:
```suggestion
A new [ABI Compatibility](https://doc.rust-lang.org/stable/std/primitive.fn.html#abi-compatibility) section in the function pointer documentation describes what it means for Rust ABI function signatures to be ABI-compatible with each other. A large part of that is the compatibility of argument types and return types, with a list of those that are currently considered compatible in Rust. For the most part, this documentation is not adding any new guarantees, only describing the existing state of compatibility.
```

Mark-Simulacrum (approved):

posts/2024-02-08-Rust-1.76.0.md
cuviper: It mentions other ABIs as well, not just the (implicit) "Rust" ABI string.

Mark-Simulacrum: Ah, for some reason I had remembered that we had scoped the initial docs to just "Rust" but, yes, you are right. Fine as-is then.

PR #1260, opened 2024-02-21
Remove duplicate commandline prompt in `console` block displays

noritada:
This PR removes duplicate commandline prompt in `console` blocks.

Thank you for taking the time.


Turbo87 (approved):
looks like this weird behavior was introduced by https://github.com/rust-lang/blog.rust-lang.org/commit/1849fd748d0c8a6515f8df19afd05b15598c3b59, but I don't see any justification for it in the git history.

IMHO it would be better / less surprising to remove the `::before` part, but this is still a good short-term fix. thanks!

PR #1261, opened 2024-02-22
Remove automatic `$` prefix for `bash` and `console` code blocks

Turbo87:
This prefix is very surprising to blog authors because e.g. GitHub does not render it that way. It also does not work well if a code block contains multiple commands where all but the first one need a manual prefix. The [original commit that introduced this](https://github.com/rust-lang/blog.rust-lang.org/commit/1849fd748d0c8a6515f8df19afd05b15598c3b59) unfortunately does not give any reasons for why it was introduced.

Related:

- https://github.com/rust-lang/blog.rust-lang.org/pull/1260

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ Pages via GitHub Actions.
 To build the site locally:
 
 ```console
-> git clone https://github.com/rust-lang/blog.rust-lang.org
-> cd blog.rust-lang.org
-> cargo run
+$ git clone https://github.com/rust-lang/blog.rust-lang.org
+$ cd blog.rust-lang.org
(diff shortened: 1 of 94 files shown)


Manishearth (approved):

kpreid:
> The [original commit that introduced this](https://github.com/rust-lang/blog.rust-lang.org/commit/1849fd748d0c8a6515f8df19afd05b15598c3b59) unfortunately does not give any reasons for why it was introduced.

One motivation might have been to cause the `$` text to not be copiable, so that someone intending to copy and run the command doesn't get the `$` that is not part of the command.

PR #1266, opened 2024-03-04
Clippy post: remove `release: true`

ev3nvy:
Recently published clippy post had `release` set to `true`, which seems incorrect, as the changelog link on the [main website](https://www.rust-lang.org/) now links to it, and not to https://blog.rust-lang.org/2024/02/08/Rust-1.76.0.html.

If just removing the line after-the-fact is not enough to unpublish it from `https://blog.rust-lang.org/releases.json`, feel free to close this PR and open a new one with the proper fix.


ehuss (approved):
Thanks!

PR #1288, opened 2024-03-26
Fix dead links an deadname in "Shape of errors to come"

FSMaxB:
Replace Sophia's old name and replace broken links to her old blog that now point to some kind of spam website.

Note that I didn't go through other posts that might have the same issues!


ehuss (approved):
Thanks!

PR #1336, opened 2024-05-19
Update check-cfg blog post with recent Cargo change

Urgau:
With recent Cargo change, namely https://github.com/rust-lang/cargo/pull/13913, some part of the blog are no longer true/accurate, this PR therefor update some sections to reflect the recent changes.

cc @epage
r? @ehuss

### DIFF
--- a/posts/2024-05-06-check-cfg.md
+++ b/posts/2024-05-06-check-cfg.md
@@ -52,13 +52,15 @@ fn win() {}
 
 ![cargo-check](../../../../images/2024-05-06-check-cfg/cargo-check.svg)
 
-## Expecting static cfgs
+## Expecting custom cfgs
 
 *UPDATE: This section was added with the release of nightly-2024-05-19.*
 
-Some crates might use cfgs, like `loom`, `fuzzing` which are always statically known at compile time. For those cases, Cargo provides via the `[lints]` table a way to statically declare those cfgs as expected.
+> In Cargo point-of-view: a custom cfg is one that is neither defined by `rustc` nor by a Cargo feature. Think of `tokio_unstable`, `has_foo`, ... but not `feature = "lasers"`, `unix` or `debug_assertions`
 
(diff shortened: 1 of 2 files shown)


posts/2024-05-06-check-cfg.md
```diff
@@ -52,6 +52,20 @@ fn win() {}
...
+
+It is done through the special `check-cfg` config under `[lints.rust.unexpected_cfgs]`:
+
+*`Cargo.toml`*
+```toml
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(loom)', 'cfg(fuzzing)'] }
+```
+
 ## Custom cfgs and build scripts
```
epage, line 69: Should we update this section to not mention cfg's from `RUSTFLAGS`, leaving it to only those set via `build.rs`?

Urgau: If your referring to the `tokio_unstable` cfg, I can see the argument for not wanting to mix them with `build.rs`, but in that case wouldn't it be better to move the `In Cargo point-of-view` note above in the newly created section "Expecting static cfgs"?

epage: I was referring to

> Some crates use custom cfgs that they **either expected from the environment (`RUSTFLAGS`or other means)** or is enabled by some logic 
in the crate `build.rs`.

But yes, the general structure is off.  "Custom cfgs and build scripts" was written to be top-level but now we have a "dynamic" and a "static" section without updating other parts, like "In Cargo's point-of-view".

Urgau: I reworked the flow of those 2 sections. Let me know what you think.

epage (approved):

ehuss (commented):
Thanks!

ehuss (approved):

PR #1337, opened 2024-05-24
Update check-cfg blog post to include new diagnostics

Urgau:
This PR update the check-cfg blog post with the new diagnostics from the compiler/cargo check, which include the new `check-cfg` lint config option; option that we mention in the section following the SVG.

Follow-up to https://github.com/rust-lang/blog.rust-lang.org/pull/1336 and https://github.com/rust-lang/rust/pull/125219.

r? @ehuss


ehuss (approved):
Thanks!

PR #1353, opened 2024-07-03
welcome TC to the Rust lang team

nikomatsakis:
cc @tmandry


therealprof:
Maybe it's just me, but I do find it a bit weird having to work through some indirections in order to find some more information about TC. Would it be possible to include at least a link to the GH profile in the post?

Kobzol:
Let me know if you want to add the link or not, otherwise I can approve and we can ship this (this has happened a month ago, after all).

tmandry (approved):
I think we can post this. You can always follow links to the RFC threads.

PR #1361, opened 2024-07-22
Add "crates.io: development update" blog post

Turbo87:
r? @rust-lang/crates-io


posts/2024-07-26-crates-io-development-update.md
```diff
@@ -0,0 +1,69 @@
...
+title: "crates.io: development update"
+author: Tobias Bieniek
+team: the crates.io team <https://www.rust-lang.org/governance/teams/crates-io>
+---
+
+Since crates.io does not have releases in the classical sense, there are no release notes either. However, the crates.io team still wants to keep you all updated about the ongoing development of crates.io. This blog post is a summary of the most significant changes that have been made to crates.io in the past months.
+
+## `cargo install`
+
+When looking at crates like [ripgrep](https://crates.io/crates/ripgrep) you will notice that the installation instructions now say `cargo install ripgrep` instead of `cargo add ripgrep`. This change was made to make it easier for users to install crates that have binary targets. `cargo add` is still the correct command to use when adding a crate as a dependency to your project, but for binary-only crates like ripgrep, `cargo install` is the way to go.
```
MarcoIeni, line 12:
```suggestion
When looking at crates like [ripgrep](https://crates.io/crates/ripgrep) you will notice that the installation instructions now say `cargo install ripgrep` instead of `cargo add ripgrep`. The team made this change to make it easier for users to install crates that have binary targets. `cargo add` is still the correct command to use when adding a crate as a dependency to your project, but for binary-only crates like ripgrep, `cargo install` is the way to go.
```
I prefer active voice vs passive voice. Explanation [here](https://developers.google.com/tech-writing/one/active-voice).
You can even just us "we" instead of "the team".

You don't need to address this comment. I just wanted to express my preference. Maybe you can consider this style for the next blog posts :)

MarcoIeni: I also noticed that sometimes you use "our" later in the document. So you use both active and passive voice, which is inconsistent.

posts/2024-07-26-crates-io-development-update.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+## API token expiry notifications
+
+Our crates.io team member [@hi-rustin](https://github.com/hi-rustin) has been very active in improving our API tokens user experience. If you create an API token with an expiry date, you will now receive a notification email three days before the token expires. This will help you to remember to renew your token before it expires and your scripts stop working.
+
+Following this change, he also implemented a way to create API tokens based on existing tokens, which will make it easier to renew tokens without having to reconfigure all the permissions. The user interface on the "API tokens" settings page now shows a "Regenerate" button, which will allow you to copy/clone existing tokens with the same permissions. Similarly, the token expiry notifications will now also contain a link that directly fills in the permissions of the expiring token, so you can easily create a new token with the same permissions.
```
MarcoIeni, line 50: It would be cool to see screenshots!

MarcoIeni:
Consider adding a closing paragraph where you invite people to write in [zulip](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io) to leave their feedback.
In the document you mentioned multiple times that you want the user feedback, but maybe they don't know where to send it 👍

posts/2024-07-26-crates-io-development-update.md
bjorn3, line 50: > copy/clone existing tokens

This seems to suggest that it would show the exact same token string as when you originally got when creating the token, but "Regenerate" suggests that it creates a new token with the same permissions and revokes the old token. Which of the two is it?

hi-rustin: The latter one. It will create a new token.

LawnGnome (approved):
I'm not going to copy-edit this, but the content looks good, and I definitely think we should do these every couple of months. :+1:

posts/2024-07-26-crates-io-development-update.md
```diff
@@ -0,0 +1,69 @@
...
+- [#8737](https://github.com/rust-lang/crates.io/pull/8737): Improve the performance of reverse dependencies using the `default_versions` table
+
+In some of these cases we have seen a 10-20x performance improvement. We are very happy to have him on the team and look forward to more improvements in the future!
+
+
+## New database server
+
+We have recently migrated our database servers to a new provider with more memory and faster storage. This has improved the performance of the website and allowed us to run more complex queries without running into performance issues.
+
+It was previously taking multiple seconds to load e.g. https://crates.io/crates/syn/reverse_dependencies, but now the server usually responds in significantly less than a second. We are very happy with the new servers and hope that you will notice the improved performance as well.
```
eth3lbert, line 69: This is definitely good progress we've made recently and should be shared with others. However, I'm not sure about including this section here as the new server contributes to only a portion of the improvement. The main benefit seems to come from query performance improvement, which should also have responded in less than a second on the previous server, or perhaps I'm mistaken. I suggest comparing the apis that leverage `recent_crate_downloads` and `version_downloads`, as `version_downloads` should experience a significant size reduction after migrating to the new server.

Overall LGTM 👍

eth3lbert (approved):

Turbo87:
@jtgeibel looks like I need an approval from a team lead again 😉

MarcoIeni:
Minor complaint as a new-joiner: I would have loved to review the changes, but the force-push makes it hard 😵

Turbo87:
fair, sorry about that 🫣

Turbo87:
I've `git reflog`ed it and restored the intermediate commits :)

MarcoIeni:
oh, wow I didn't know you could do that!
Thanks all the changes look amazing ⭐

posts/2024-07-26-crates-io-development-update.md
```diff
@@ -0,0 +1,73 @@
...
+- [#7941](https://github.com/rust-lang/crates.io/pull/7941): Improve crates endpoint performance
+- [#8734](https://github.com/rust-lang/crates.io/pull/8734): Add partial index on versions table
+- [#8737](https://github.com/rust-lang/crates.io/pull/8737): Improve the performance of reverse dependencies using the `default_versions` table
+
+In addition to that, we have recently migrated our database servers to a new provider with more memory and faster storage. This has also improved the performance of the website and allowed us to run more complex queries without running into performance issues. It was previously taking multiple seconds to load e.g. https://crates.io/crates/syn/reverse_dependencies, but now the server usually responds in much less than a second.
+
+Another piece of the puzzle was archiving old data that is no longer needed for the website. We have moved the download counts older than 90 days into JSON files that are stored on S3 and will soon be publicly available for download via our CDNs. This has reduced the size of the database significantly and improved the performance of some of our background jobs.
```
MarcoIeni, line 68:
```suggestion
Another piece of the puzzle was archiving old data that is no longer needed for the website. We have moved the download counts older than 90 days into CSV files that are stored on S3 and will soon be publicly available for download via our CDNs. This has reduced the size of the database significantly and improved the performance of some of our background jobs.
```
right?

MarcoIeni (approved):

Turbo87:
@rust-lang/leads I already have the team member approvals. Could one of you approve to so that we can ship this? 🙏

Kobzol (approved):
The date is a bit stale (`26 July`), let me know if you want to change it and I can re-approve.

GuillaumeGomez (approved):

Turbo87:
thanks everyone! it's live now: https://blog.rust-lang.org/2024/07/29/crates-io-development-update.html :)

PR #1392, opened 2024-09-04
impl trait capture rules blog post

nikomatsakis:
Explain the changes coming for `-> impl Trait` in Rust 2024. Also gives a heads-up about the new `use` syntax.


nikomatsakis:
cc @jackh726 @traviscross @compiler-errors @tmandry

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+#### impl trait in traits
+
+As we pursued the design for impl trait in traits ([RFC 3425](https://rust-lang.github.io/rfcs/3425-return-position-impl-trait-in-traits.html)), we encountered a number of challenges related to the capturing of lifetimes. [In order to get the symmetries that we wanted to work](https://hackmd.io/zgairrYRSACgTeZHP1x0Zg) (e.g., that one can write `-> impl Future` in a trait and impl with the expected effect), we had to change the capture rules to capture all lifetime parameters uniformly.
+
+## Rust 2024 design
+
+The above problems motivated us to take a new approach in Rust 2024. The approach is a combination of two things:
+
+* a new default, applicable only in Rust 2024;
+* a syntax for explicit capture, usably in any edition, written like
```
VorpalBlade, line 140: This sentence isn't finished?

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+layout: post
+title: "Changes to impl Trait in Rust 2024"
+author: Niko Matsakis
+team: Leadership Council <https://www.rust-lang.org/governance/teams/leadership-council>
+---
+This blog post describes some small but significant changes with (return position) `impl Trait` capture rules that are coming in Rust 2024. The goal of these changes is to simplify how `impl Trait` works to better match what people want while also giving a flexible syntax that allows users to have full control when needed.
+
+## TL;DR
+
+* Return-position impl Trait in Rust 2024 can now reference all fn arguments by default.
```
nikomatsakis, line 11: we should link to the RFCs and issues where decisions were made here I think so people can click for more details

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
+---
+layout: post
+title: "Changes to impl Trait in Rust 2024"
+author: Niko Matsakis
+team: Leadership Council <https://www.rust-lang.org/governance/teams/leadership-council>
```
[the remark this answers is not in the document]
nikomatsakis: Fixed

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+
+```rust
+fn indices<T>(
+    slice: &[T],
+) -> impl Iterator<Item = usize> + use<> {
+    0 .. slice.len()
+}
+```
+
+**Implementation limitation.** Unfortunately, if you actually try the above example on nightly today, you'll see that it doesn't compile ([try it for yourself](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=1d6d23ef3813da05ac78a4b97b418f21)). That's because `use<>` bounds have only partially been implemented: currently, they must always include at least the type parameters. This corresponds to the limitations of impl trait in earlier editions, which always *must* capture type parameters. In this case, that means we can write the following, which also avoids the compilation error, but is still more conservative than necessary ([try it yourself](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=7965043f4686d5a89b47aa5bfc4f996f)):
```
CleanCut, line 195: Maybe address if/when this limitation might be eliminated?

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+This may actually be what you want: it gives you room to modify `indices` later and have it actually make use of `slice` in the returned iterator, for example. Put another way, the new default continues the `impl Trait` tradition of retaining flexibility for the function to change its implementation without breaking callers.
+
+But what if it's *not* what you want? What if you want to guarantee that `indices` will not retain a reference to `data` in its return value? You can do that now with the new `use` bounds. The idea is that your `impl Trait` can include a bound written `use<...>` which lists out the generic parameters that the return value may use. If you include `+ use<>`, then, this means "do not capture anything". So changing `indices` to be written as follows will allow `main` to compile:
+
+```rust
+fn indices<T>(
+    slice: &[T],
+) -> impl Iterator<Item = usize> + use<> {
```
CleanCut, line 190: I'm sure this is far too late for input, but I'm a bit let down that despite using the word `capture` approximately 30 times, including in the PR title, URL, headers, and all over the text...the keyword chosen was `use`.  I'm not looking forward to having to teach my students to remember this special way to use `use` to specify what to _capture_. 🤦🏻‍♂️ 😜

CleanCut (commented):
I was just having fun reading and thought I'd indulge my inner reviewer for a bit. 😆 

Inconsistent word usage:
- "return position" vs "return-position"
- "impl trait" vs "`impl trait`" (with backticks)

posts/2024-09-05-impl-trait-capture-rules.md
nikomatsakis: Appreciate the feedback! Actually, maybe I should try to avoid saying capture, because the intuition really isn't as good as "using". That is, the word "use" was chosen because it corresponds to what values are *used* by the hidden type (that is, appear in). Capture seems to imply that things are 'taken' from the place they appear.

CleanCut: Using more consistent language between the code and the description would definitely help with comprehension.

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+layout: post
+title: "Changes to impl Trait in Rust 2024"
+author: Niko Matsakis
+team: the language team <https://www.rust-lang.org/governance/teams/lang>
+---
+This blog post describes some small but significant changes with (return position) `impl Trait` capture rules that are coming in Rust 2024. The goal of these changes is to simplify how `impl Trait` works to better match what people want while also giving a flexible syntax that allows users to have full control when needed.
+
+## TL;DR
+
+* Return-position impl Trait in Rust 2024 can now reference all fn arguments by default.
```
tmandry, line 11:
```suggestion
* Return-position impl Trait in Rust 2024 can now reference all function arguments by default.
```

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
+---
+layout: post
+title: "Changes to impl Trait in Rust 2024"
+author: Niko Matsakis
+team: the language team <https://www.rust-lang.org/governance/teams/lang>
+---
+This blog post describes some small but significant changes with (return position) `impl Trait` capture rules that are coming in Rust 2024. The goal of these changes is to simplify how `impl Trait` works to better match what people want while also giving a flexible syntax that allows users to have full control when needed.
```
tmandry, line 7:
```suggestion
The default way `impl Trait` works in return position is changing in Rust 2024. These changes are meant to simplify `impl Trait` to better match what people want most of the time. We're also adding a flexible syntax that gives you full control when you need it.
```

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+
+```rust
+fn process_datums(
+    datums: &[Datum]
+) -> impl Iterator<Item = ProcessedDatum> {
+    datums
+        .iter()
+        .map(|datum| datum.process())
+}
+```
```
tmandry, line 26: The plural of "datum" is "data".

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+
+The use of `-> impl Iterator` in return position here means that the functions "some kind of iterator". The actual type will be determined by the compiler based on the function body. It is called the "hidden type" because callers do not get to know exactly what it is, they have to code against the `Iterator` trait. However, at code generation time, the compiler will generate code based on the actual precise type, which ensures that callers are fully optimized.
```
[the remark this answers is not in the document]
tmandry:
```suggestion
The use of `-> impl Iterator` in return position here means that the functions "some kind of iterator". The actual type will be determined by the compiler based on the function body. It is called the "hidden type" because callers do not get to know exactly what it is; they have to code against the `Iterator` trait. However, at code generation time, the compiler will generate code based on the actual precise type, which ensures that callers are fully optimized.
```

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+    datums: Vec<T>,
+) -> impl Iterator<Item = ()> + '_ {
+    datums
+        .into_iter()
+        .map(|datum| context.process(datum))
+}
+```
+
+Here the `process` function applies `context.process` to each of the elements in `datums` (of type `T`). Because it captures `context`, it is declared as `+ '_`. This tells the compiler that the hidden type must outlive `'_`. Because this hidden type includes the `Vec<T>`, this in turn means that `T` must outlive `'_` -- but this is not true. So you get a compilation error ([try it on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b742fbf9b083a6e837db0b170489f34a)). 
+
```
tmandry, line 100:
```suggestion

The underlying problem with this suggestion is that it uses a convenient-but-sometimes-incorrect tool, an outlives bound, to work around the existing capture rules, which say a lifetime is captured only if it appears in the bounds of the `impl Trait`. Even when this technique fixes compilation errors, it creates a semantic confusion in the minds of users. A `+ 'a` bound  means "valid for at least as long as `'a`". But here, it is being used to mean that the hidden type _captures_ the lifetime, or in other words, it is "only valid for as long as `'a`". The fact that these have nearly opposite meanings makes it extremely difficult to build a mental model of what these bounds are actually _for_.

```

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,211 @@
...
+        ..
+    }
+}
+```
+
+In practice, because of the problems with lifetime capture, this is not the actual desugaring. The actual desugaring is to a special kind of `impl Trait` that is allowed to capture all lifetimes. But that form of `impl Trait` was not exposed to end-users.
+
+#### impl trait in traits
+
+As we pursued the design for impl trait in traits ([RFC 3425](https://rust-lang.github.io/rfcs/3425-return-position-impl-trait-in-traits.html)), we encountered a number of challenges related to the capturing of lifetimes. [In order to get the symmetries that we wanted to work](https://hackmd.io/zgairrYRSACgTeZHP1x0Zg) (e.g., that one can write `-> impl Future` in a trait and impl with the expected effect), we had to change the capture rules to capture all lifetime parameters uniformly.
```
tmandry, line 133: It might be worth reiterating this later on in the post: I doubt most users are aware that RPITIT capture rules work differently.

posts/2024-09-05-impl-trait-capture-rules.md
nikomatsakis: I made the edits,  @CleanCut, take a look at the new revision. I felt it overall reads much better.

posts/2024-09-05-impl-trait-capture-rules.md
nikomatsakis: (this text has changed)

posts/2024-09-05-impl-trait-capture-rules.md
nikomatsakis: Nice.

posts/2024-09-05-impl-trait-capture-rules.md
nikomatsakis: I changed this text pretty significantly in the new revision. 

> Here the `process` function applies `context.process` to each of the elements in `datums` (of type `T`). Because the return value uses `context`, it is declared as `+ 'c`. Our real goal here is to allow the return type to use `'c`; writing `+ 'c` achieves that goal because `'c` not appears in the bound listing. However, writing `+ 'c` *also* means that the hidden type must outlive `'c` -- and in this case, that is not strictly necessary! The hidden type is going to be a wrapper around `std::vec::IntoIter<T>`, and for this type to outlive `'c`, we would need a where clause `T: 'c`. But that where-clause should only be required if we use a value of type `&'c T`, and we don't. The result is that this example doe snot compile ([try it on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b742fbf9b083a6e837db0b170489f34a)), even though it should.

I'm going to see if I can merge some elements of what you wrote into that.

posts/2024-09-05-impl-trait-capture-rules.md
nikomatsakis: I'm aware of that :) I just like `s`. But I guess I'll change it...

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -0,0 +1,248 @@
...
+) -> impl Iterator<Item = ProcessedDatum> {
+    data
+        .iter()
+        .map(|datum| datum.process())
+}
+```
+
+The use of `-> impl Iterator` in return position here means that the function returns "some kind of iterator". The actual type will be determined by the compiler based on the function body. It is called the "hidden type" because callers do not get to know exactly what it is; they have to code against the `Iterator` trait. However, at code generation time, the compiler will generate code based on the actual precise type, which ensures that callers are fully optimized.
+
+Although callers don't know the exact type, they do need to know that it will continue to borrow the `datums` argument so that they can ensure that the `datums` reference remains valid while iteration occurs. Further, callers must be able to figure this out based solely on the type signature, without looking at the function body.
```
lqd, line 36: Pluralization may have been partial, there are a handful of `datums` left.

```suggestion
Although callers don't know the exact type, they do need to know that it will continue to borrow the `data` argument so that they can ensure that the `data` reference remains valid while iteration occurs. Further, callers must be able to figure this out based solely on the type signature, without looking at the function body.
```

posts/2024-09-05-impl-trait-capture-rules.md
```diff
@@ -202,10 +216,33 @@ fn indices<T>(
...
 
+This implementation limitation is only temporary and will hopefully be lifted soon! You can follow the current status at [tracking issue #130031](https://github.com/rust-lang/rust/issues/130031).
+
+**Alternative: `'static` bounds.** For the special case of capturing **no** references at all, it is also possible to use a `'static` bound, like so ([try it yourself](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=3054bbf64652cb4890d56ac03b47a35c)):
+
+```rust
+fn indices<'s, T>(
+    slice: &'s [T],
+) -> impl Iterator<Item = usize> + 'static {
+    //                             -----
```
tmandry, line 227:
```suggestion
    //                             -------
```

tmandry (approved):

PR #1397, opened 2024-09-05
unpublish wasip2 post

yoshuawuyts:
We got unlucky on the release - the target builds, but is not included in rustup so the actual instructions of the post don't work. This will work in Rust 1.82.

(resolving this in real-time with @pietroalbini on a call)


pietroalbini (approved):

alexcrichton:
For future archivists coming across this (or anyone just now coming across this) I commented more about this in [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/1.2E81.2E0/near/467923833) and I'll copy that here too:

---

Some more context on https://github.com/rust-lang/blog.rust-lang.org/pull/1397 and the wasm32-wasip2 target:

Turns out we got unlucky on releasing this and 1.81.0 does NOT have the wasm32-wasip2 target available in rustup (e.g. `rustup target add wasm32-wasip2` fails). That raises the question of how did this get into the release notes and we end up in this situation? It turns out that there were actually two PRs that were required to get `wasm32-wasip2` into rustup. The [first PR](https://github.com/rust-lang/rust/pull/126967/) was what was linked from the release notes and  is why we thought that this was released in 1.81. That PR is indeed released in 1.81 and is on the `stable` branch of the main repository.

Unfortunately that PR had a bug where the build-manifest tool wasn't updated. That meant that despite `wasm32-wasip2` being built in CI and having artifacts uploaded rustup wasn't able to see it. This was found and fixed in a [second PR](https://github.com/rust-lang/rust/pull/127867). At this point you can probably guess where this is going, but turns out we got real unlucky and while the first PR made it into Rust 1.81 the second PR didn't make the cut and is a part of 1.82 instead. Turns out we didn't realize this until just now, creating this confusion!

After talking with some folks what we're doing now is:

* [Taking down the wasm32-wasip2 blog post and the specific note in the release notes about it](https://github.com/rust-lang/blog.rust-lang.org/pull/1397)
* I'm going to make a PR to remove this from the release notes on `main`
* I'm going to open an issue to coordinate and get this re-added to the release notes for 1.82

In the meantime this does not affect the `wasm32-wasip1` target at all and/or the rename with `wasm32-wasi`. That part of the release notes is still accurate. This is only affecting the new tier 2 status of `wasm32-wasip2`, a separate target. The target works fine on beta and the plan is to republish these communications for the 1.82 release.

Sorry for the confusion here and thanks Pietro for helping us quickly take down the blog so quickly!

PR #1400, opened 2024-09-05
Fix grammar in 1.81.0 release

schuelermine:
This sentence is on the whole a bit weird as it’s not clear what the subject of "warn if it doesn't" is, but I think this conjugation makes more sense overall

### DIFF
--- a/posts/2024-09-05-Rust-1.81.0.md
+++ b/posts/2024-09-05-Rust-1.81.0.md
@@ -45,7 +45,7 @@ requirements documented in [PartialOrd] and [Ord].
 ### `#[expect(lint)]`
 
 1.81 stabilizes a new lint level, `expect`, which allows explicitly noting that
-a particular lint *should* occur, and warning if it doesn't.  The intended use
+a particular lint *should* occur, and warn if it doesn't.  The intended use
 case for this is temporarily silencing a lint, whether due to lint
 implementation bugs or ongoing refactoring, while wanting to know when the lint
 is no longer required.


schuelermine:
I’ve just been informed that the sentence might’ve been intended to read as "allows explicitly noting, and allows explicitly warning". I feel that way of phrasing it is still worse than this one.

Kobzol (approved):

posts/2024-09-05-Rust-1.81.0.md
```diff
@@ -45,7 +45,7 @@ requirements documented in [PartialOrd] and [Ord].
 ### `#[expect(lint)]`
 
 1.81 stabilizes a new lint level, `expect`, which allows explicitly noting that
-a particular lint *should* occur, and warning if it doesn't.  The intended use
+a particular lint *should* occur, and warn if it doesn't.  The intended use
```
compiler-errors, line 48: If we're going to not use the gerund here then it needs to agree with the singular "lint level".

```suggestion
a particular lint *should* occur, and warns if it doesn't.  The intended use
```

i.e. "lint level, `expect`, which [...] warns if it doesn't"

Though I don't agree that this needs changing at all. Or at least this should be split into two separate sentences if it's ambiguous.

compiler-errors (changes requested):

posts/2024-09-05-Rust-1.81.0.md
schuelermine: The "warn" is not marked because there's an auxiliary, "should".

schuelermine:
@compiler-errors Do you have any comment on what I wrote in response to your comment?

senekor:
How about:

1.81 stabilizes a new lint level, `expect`, which allows explicitly noting that a particular lint *should* occur. 
A warning is triggered if it doesn't.

senekor:
the post is old now and nobody seems to care anymore

PR #1415, opened 2024-10-15
Announcing Rust 1.82.0

cuviper:
cc @rust-lang/release 
r? @Mark-Simulacrum

[Rendered](https://github.com/cuviper/blog.rust-lang.org/blob/rust-1.82.0/posts/2024-10-17-Rust-1.82.0.md)


posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+<b><span style="color:#0A0">homepage:</span></b> https://github.com/rust-lang/cc-rs
+<b><span style="color:#0A0">repository:</span></b> https://github.com/rust-lang/cc-rs
+<b><span style="color:#0A0">crates.io:</span></b> https://crates.io/crates/cc/1.1.23
+<b><span style="color:#0A0">features:</span></b>
+  jobserver = []
+  parallel  = [dep:libc, dep:jobserver]
+<b><span style="color:#0AA">note</span></b><b>:</b> to see how you depend on cc, run `<b><span style="color:#0AA">cargo tree --invert --package cc@1.1.23</span></b>`
+</pre>
+
+By default `cargo info` describes the package version in the local `Cargo.lock`, if any, but as you can see it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that. The feature list also indicates a `+` next to any that are en
```
Mark-Simulacrum:
```suggestion
By default `cargo info` describes the package version in the local `Cargo.lock`, if any, but as you can see it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that. The feature list also indicates a `+` next to any that are enabled.
```

Mark-Simulacrum (approved):

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+
+#### Mac Catalyst targets are now Tier 2
+
+[Mac Catalyst](https://developer.apple.com/mac-catalyst/) is a technology by Apple that allows running iOS applications natively on the Mac. This is especially useful when testing iOS-specific code, as `cargo test --target=aarch64-apple-ios-macabi --target=x86_64-apple-ios-macabi` mostly just works (in contrast to the usual iOS targets, which need to be bundled using external tooling before they can be run on a native device or in the simulator).
+
+[The targets](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-ios-macabi.html) are now tier 2, and can be downloaded with `rustup target add aarch64-apple-ios-macabi x86_64-apple-ios-macabi`, so now is an excellent time to update your CI pipeline to test that your code also runs in iOS-like environments.
+
+### Minimal exhaustive patterns
+
+Empty patterns can now be omitted in common cases:
```
kpreid, line 62: It seems to me that this section heading isn't very clear. It's true that the feature name is `min_exhaustive_patterns`, but the name doesn't really communicate what this is, particularly because the obvious interpretation of “exhaustive” is in the sense that “`match`es must be exhaustive”, which is *already* true, so saying “[Rust now allows] exhaustive patterns” is odd.

```suggestion
### Omitting empty types in pattern matching

Patterns which match empty (a.k.a. uninhabited) types by value can now be omitted:
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+
+```rust
+use std::convert::Infallible;
+pub fn safe_unwrap<T>(x: Result<T, Infallible>) -> T {
```
kpreid, line 66: Let’s not contribute to the confusion of “safe” with “doesn’t panic” or “correct”, and avoid using it except in the formal sense:

```suggestion
pub fn unwrap_without_panic<T>(x: Result<T, Infallible>) -> T {
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+    let Ok(x) = x; // the `Err` case does not need to appear
+    x
+}
+```
+
+This works with empty types such as a variant-less `enum Void {}`, or structs and enums with a visible empty field and no `#[non_exhaustive]` attribute. It will also be particularly useful in combination with the never type `!`, although that type is still unstable at this time.
+
+This feature is "minimal" because there are still some exclusions at this time. For reasons related to uninitialized values and unsafe code, omitting patterns is not allowed if the empty type is accessed through a reference, pointer, or union field:
```
kpreid, line 74: This suggestion goes with my earlier rewording of the section heading which removed “minimal”, and doesn't make sense otherwise:

```suggestion
There are still some cases where empty patterns must still be written. For reasons related to uninitialized values and unsafe code, omitting patterns is not allowed if the empty type is accessed through a reference, pointer, or union field:
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+pub fn safe_unwrap_ref<T>(x: &Result<T, Infallible>) -> &T {
+    match x {
+        Ok(x) => x,
+        // this branch cannot be omitted because of the reference
+        Err(infallible) => match *infallible {},
+    }
+}
+```
+
+To avoid interfering with crates that wish to support several rust versions, branches with empty patterns are not yet warned as "unreachable", despite the fact that they can be removed.
```
kpreid, line 86:
```suggestion
To avoid interfering with crates that wish to support several Rust versions, `match` arms with empty patterns are not yet reported as “unreachable code” warnings, despite the fact that they can be removed.
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+    x
+}
+```
+
+This works with empty types such as a variant-less `enum Void {}`, or structs and enums with a visible empty field and no `#[non_exhaustive]` attribute. It will also be particularly useful in combination with the never type `!`, although that type is still unstable at this time.
+
+This feature is "minimal" because there are still some exclusions at this time. For reasons related to uninitialized values and unsafe code, omitting patterns is not allowed if the empty type is accessed through a reference, pointer, or union field:
+
+```rust
+pub fn safe_unwrap_ref<T>(x: &Result<T, Infallible>) -> &T {
```
kpreid, line 77: Same rationale as previous: let’s not use “safe” informally.

```suggestion
pub fn unwrap_ref_without_panic<T>(x: &Result<T, Infallible>) -> &T {
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+    match x {
+        Ok(x) => x,
+        // this branch cannot be omitted because of the reference
```
kpreid, line 80: Syntactically, `match`es have arms, not branches; branching is what the `match` _does_. Arguably "branch" is also correct here but it's unnecessarily non-specific.

```suggestion
        // this arm cannot be omitted because of the reference
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+
+With this release, Rust standardizes on a set of rules for how NaN values behave. This set of rules is *not* fully deterministic, which means that the result of operations like `(0.0 / 0.0).is_sign_positive()` can differ depending on the hardware architecture, optimization levels, and the surrounding code. Code that aims to be fully portable should avoid using `to_bits` and should use `f.signum() == 1.0` instead of `f.is_sign_positive()`. However, the rules are carefully chosen to still allow advanced optimization techniques such as NaN boxing to be realized in Rust code. For more details on what the exact rules are, check out our [documentation](https://doc.rust-lang.org/std/primitive.f32.html#nan-bit-patterns).
```
kpreid, line 92: I think “optimization” isn’t a good word to use here because it sounds like Rust might be _automatically_ NaN-boxing things someday.

```suggestion
With this release, Rust standardizes on a set of rules for how NaN values behave. This set of rules is *not* fully deterministic, which means that the result of operations like `(0.0 / 0.0).is_sign_positive()` can differ depending on the hardware architecture, optimization levels, and the surrounding code. Code that aims to be fully portable should avoid using `to_bits` and should use `f.signum() == 1.0` instead of `f.is_sign_positive()`. However, the rules are carefully chosen to still allow advanced data representation techniques such as NaN boxing to be implemented in Rust code. For more details on what the exact rules are, check out our [documentation](https://doc.rust-lang.org/std/primitive.f32.html#nan-bit-patterns).
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+
+With the semantics for NaN values settled, this release also permits the use of floating-point operations in `const fn`. Due to the reasons described above, operations like `(0.0 / 0.0).is_sign_positive()` can produce a different result when executed at compile-time vs at run-time; this is not a bug and code must not rely on a `const fn` always producing the exact same result.
+
+### Native syntax for creating a raw pointer
+
+Unsafe code sometimes has to deal with pointers that may dangle, may be misaligned, or may not point to valid data. A common case where this comes up are packed structs. In such a case, it is important to avoid creating a reference, as that would cause undefined behavior. This means the usual `&` and `&mut` operators cannot be used, as those create a reference -- even if the reference is immediately cast to a raw pointer, it's too late to avoid the undefined behavior.
```
kpreid, line 98:
```suggestion
Unsafe code sometimes has to deal with pointers that may dangle, may be misaligned, or may not point to valid data. A common case where this comes up are `repr(packed)` structs. In such a case, it is important to avoid creating a reference, as that would cause undefined behavior. This means the usual `&` and `&mut` operators cannot be used, as those create a reference -- even if the reference is immediately cast to a raw pointer, it's too late to avoid the undefined behavior.
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
++    #[allow(unused_unsafe)]
+     let static_mut_ptr = unsafe { std::ptr::addr_of_mut!(STATIC_MUT) };
+ }
+```
+
+A future version of Rust is expected to generalize this to other expressions which would be safe in this position, not just statics.
+
+### Unsafe attributes
+
+Some Rust attributes, such as `no_mangle`, can be used to [cause Undefined Behavior without any `unsafe` block](https://github.com/rust-lang/rust/issues/28179). If this was regular code we would require them to be placed in an `unsafe {}` block, but so far attributes have not had comparable syntax. To reflect the fact that these attributes can undermine Rust's safety guarantees, they are now considered "unsafe" and should be written as follows:
```
kpreid, line 161:
```suggestion
Some Rust attributes, such as [`no_mangle`](https://doc.rust-lang.org/reference/abi.html#the-no_mangle-attribute), can be used to [cause Undefined Behavior without any `unsafe` block](https://github.com/rust-lang/rust/issues/28179). If this was regular code we would require them to be placed in an `unsafe {}` block, but so far attributes have not had comparable syntax. To reflect the fact that these attributes can undermine Rust's safety guarantees, they are now considered "unsafe" and should be written as follows:
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+<b><span style="color:#0A0">homepage:</span></b> https://github.com/rust-lang/cc-rs
+<b><span style="color:#0A0">repository:</span></b> https://github.com/rust-lang/cc-rs
+<b><span style="color:#0A0">crates.io:</span></b> https://crates.io/crates/cc/1.1.23
+<b><span style="color:#0A0">features:</span></b>
+  jobserver = []
+  parallel  = [dep:libc, dep:jobserver]
+<b><span style="color:#0AA">note</span></b><b>:</b> to see how you depend on cc, run `<b><span style="color:#0AA">cargo tree --invert --package cc@1.1.23</span></b>`
+</pre>
+
+By default `cargo info` describes the package version in the local `Cargo.lock`, if any, but as you can see it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that. The feature list also indicates a `+` next to any that are enabled.
```
kpreid, line 46:
```suggestion
By default, `cargo info` describes the package version in the local `Cargo.lock`, if any, but as you can see, it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that. The feature list also indicates a `+` next to any that are enabled by default.
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+- [`core::arch::wasm32::u16x8_relaxed_q15mulr`](https://doc.rust-lang.org/nightly/core/arch/wasm32/fn.u16x8_relaxed_q15mulr.html)
+- [`core::arch::wasm32::i16x8_relaxed_dot_i8x16_i7x16`](https://doc.rust-lang.org/nightly/core/arch/wasm32/fn.i16x8_relaxed_dot_i8x16_i7x16.html)
+- [`core::arch::wasm32::u16x8_relaxed_dot_i8x16_i7x16`](https://doc.rust-lang.org/nightly/core/arch/wasm32/fn.u16x8_relaxed_dot_i8x16_i7x16.html)
+- [`core::arch::wasm32::i32x4_relaxed_dot_i8x16_i7x16_add`](https://doc.rust-lang.org/nightly/core/arch/wasm32/fn.i32x4_relaxed_dot_i8x16_i7x16_add.html)
+- [`core::arch::wasm32::u32x4_relaxed_dot_i8x16_i7x16_add`](https://doc.rust-lang.org/nightly/core/arch/wasm32/fn.u32x4_relaxed_dot_i8x16_i7x16_add.html)
+
+These APIs are now stable in const contexts:
+
+- [`std::task::Waker::from_raw`](https://doc.rust-lang.org/nightly/std/task/struct.Waker.html#method.from_raw)
+- [`std::task::Waker::waker`](https://doc.rust-lang.org/nightly/std/task/struct.Waker.html#method.from_raw)
```
tomaka, line 308: The link is bad here, and the `Waker::waker` function doesn't seem to exist

posts/2024-10-17-Rust-1.82.0.md
Nadrieril: I agree with all your comments on this section (I'm the one who wrote the original text) 👍

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+  jobserver = []
+  parallel  = [dep:libc, dep:jobserver]
+<b><span style="color:#0AA">note</span></b><b>:</b> to see how you depend on cc, run `<b><span style="color:#0AA">cargo tree --invert --package cc@1.1.23</span></b>`
+</pre>
+
+By default `cargo info` describes the package version in the local `Cargo.lock`, if any, but as you can see it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that. The feature list also indicates a `+` next to any that are enabled.
+
+### Apple target promotions
+
+#### macOS on ARM64 is now Tier 1
```
shepmaster, line 50:
```suggestion
#### macOS on AArch64 is now Tier 1
```

So, this is an awkward one. Pretty much everywhere *Rust* refers to this architecture, we[^1] call it "aarch64". *Apple* appears to use "arm64" in code paths and "Apple Silicon" in user-facing / marketing venues.

For the purpose of this, a Rust-specific blog post, I think we should primarily refer to the architecture as `aarch64` if it's code or "AArch64" if it's human talk (according to [the Wikipedia article](https://en.wikipedia.org/wiki/AArch64)). We should then extend the aliases section below to mention "ARM64".

[^1]: If I recall correctly, this was basically a unilateral decision on my part to better match our existing targets.

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,320 @@
...
+
+The Rust target `aarch64-apple-darwin` for macOS on ARM64 (M1-family or later Apple Silicon CPUs) is now a tier 1 target, indicating our highest guarantee of working properly. As the [platform support](https://doc.rust-lang.org/stable/rustc/platform-support.html) page describes, every change in the Rust repository must pass full tests on every tier 1 target before it can be merged. This ARM64 target was added to tier 2 back in Rust 1.49, making it available in `rustup`, and this new milestone raises it on par with ARM64 Linux and the X86 macOS, Linux, and Windows targets.
```
shepmaster, line 52:
```suggestion
The Rust target `aarch64-apple-darwin` for macOS on AArch64 (also known as ARM64 or Apple Silicon) is now a tier 1 target, indicating our highest guarantee of working properly. As the [platform support](https://doc.rust-lang.org/stable/rustc/platform-support.html) page describes, every change in the Rust repository must pass full tests on every tier 1 target before it can be merged. This target was introduced as tier 2 back in Rust 1.49, making it available in `rustup`. This new milestone puts the `aarch64-apple-darwin` target on par with the AArch64 Linux and the X86 macOS, Linux, and Windows targets.
```

posts/2024-10-17-Rust-1.82.0.md
cuviper: On the other hand, our [platform support](https://doc.rust-lang.org/nightly/rustc/platform-support.html) page consistently uses "ARM64" in the target "notes" descriptions, and there are also a few in triples: T2 `arm64ec-pc-windows-msvc`, T3 `arm64e-apple-{darwin,ios,tvos}` and `arm64_32-apple-watchos`.

The [Rust 1.49.0](https://blog.rust-lang.org/2020/12/31/Rust-1.49.0.html) announcement spelled out "64-bit ARM" for both Linux and macOS -- would that be acceptable to you here?

posts/2024-10-17-Rust-1.82.0.md
cuviper: I think we're getting too many clauses in that sentence -- how about this?
```suggestion
By default, `cargo info` describes the package version in the local `Cargo.lock`, if any. As you can see, it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that. The feature list also indicates a `+` next to any that are enabled by default.
```

Regarding the features "by default", I'm actually surprised that this doesn't look at local enablement, but I just verified it's the same even if I have `default-features=false` in my dep. Using an `indexmap` dep, it looks like the actual list is just `+default`, plain white text without `+` for other features enabled by default (`std`), and gray text for remaining features (even if they are locally enabled). Maybe we should remove this feature sentence altogether?

posts/2024-10-17-Rust-1.82.0.md
cuviper: Thanks, I'm guessing that was a copy-paste error, which also needs to be fixed in the release notes.
```suggestion
```

posts/2024-10-17-Rust-1.82.0.md
shepmaster: > spelled out "64-bit ARM"

That'd OK. At least we skip picking one name or the other.

> consistently uses "ARM64"

It may be relevant that Wikipedia redirects `ARM64` to the aforementioned `AArch64` page, and only uses "ARM64" once as an "also known as". Perhaps we should follow suit?

> and there are also a few in triples: T2 `arm64ec-pc-windows-msvc`, T3 `arm64e-apple-{darwin,ios,tvos}`

To my limited knowledge arm64**e** and arm64**ec** are ABIs, not processor architectures. It may be a sign they should not be present there in these target triple?

> and `arm64_32-apple-watchos`.

The same may be true for `arm64_32`?

cuviper: > > spelled out "64-bit ARM"
> 
> That'd be OK. At least we skip picking one name or the other.

At the very least, I think this is not time or place to litigate it. :) I'll make the change.

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,319 @@
...
+
+This code is now allowed:
+
+```rust
+static mut STATIC_MUT: Type = Type::new();
+extern "C" {
+    static EXTERN_STATIC: Type;
+}
+fn main() {
+     let static_mut_ptr = std::ptr::addr_of_mut!(STATIC_MUT);
```
traviscross, line 140:
```suggestion
     let static_mut_ptr = &raw mut STATIC_MUT;
```

Might as well use the new syntax here.

cuviper: I think the intention was to show that both ways are fine, but the later compatibility example also shows `addr_of_mut`, so I guess we don't need it here.

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,319 @@
...
+
+In an expression context, `STATIC_MUT` and `EXTERN_STATIC` are [place expressions](https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions). Previously, the compiler's safety checks were not aware that the raw ref operator did not actually affect the operand's place, treating it as a possible read or write to a pointer. No unsafety is actually present, however, as it just creates a pointer.
+
+Relaxing this may cause problems where some unsafe blocks are now reported as unused if you deny the `unused_unsafe` lint, but they are now only useful on older versions. Annotate these unsafe blocks with `#[allow(unused_unsafe)]` if you wish to support multiple versions of Rust, as in this example diff:
+
+```diff
+ static mut STATIC_MUT: Type = Type::new();
+ fn main() {
++    #[allow(unused_unsafe)]
+     let static_mut_ptr = unsafe { std::ptr::addr_of_mut!(STATIC_MUT) };
```
traviscross, line 153:
```suggestion
     let static_mut_ptr = unsafe { &raw mut STATIC_MUT };
```

cuviper: This one should *not* change, because if you're writing code that's lint-compatible with older versions, then you can't use `&raw` yet either.

compiler-errors:
I believe https://github.com/rust-lang/rust/issues/129577 is missing from the relnotes. It's currently lang-nominated, but I assume T-lang may not have the time to draft relnotes sections given a deadline. 

I'm happy to help write something, though I'm very much not a good writer. Perhaps we could lift some verbiage from https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules.html#impl-traits-can-include-a-use-bound-to-specify-precisely-which-generic-types-and-lifetimes-they-use.

Also, process-wise, is there something we can do to ensure that things like this don't slip through the cracks?

cuviper:
@compiler-errors precise capturing is listed in the release notes, and it was on our candidate list for the blog, but I dropped it because I wasn't sure what to say about it -- and we already have a lot more content than usual. It does have a place in the [RPIT captures blog post](https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules.html), as far as it pertains to 2024.

For stable editions now, does it actually enable anything new though? AIUI you currently have to write *all* type parameters, and writing lifetime parameters like `+ use<'a>` is the same as `+ 'a`, isn't it? If there's more nuance, then that's what I need help writing. I do understand the need after the 2024 change, but that's not stable yet.

compiler-errors:
Well, so `+ use<'a>` is almost synonymous with `+ 'a`, though `+ use<'a, 'b>` is not the same as `+ 'a + 'b`. I believe it's useful in 1.82 even in edition 2021 as a replacement for the [`Captures` trick](https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html#the-captures-trick), and `+ use<>` is a slightly more intentional way of writing `+ 'static` in some cases. Not totally certain if that's *significant enough*, though.

cuviper:
Hmm, OK, I think that earlier section on [outlives](https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html#the-outlives-trick) has the extra nuance I was missing -- why the current syntax is insufficient.

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,319 @@
...
+<b><span style="color:#0A0">repository:</span></b> https://github.com/rust-lang/cc-rs
+<b><span style="color:#0A0">crates.io:</span></b> https://crates.io/crates/cc/1.1.23
+<b><span style="color:#0A0">features:</span></b>
+  jobserver = []
+  parallel  = [dep:libc, dep:jobserver]
+<b><span style="color:#0AA">note</span></b><b>:</b> to see how you depend on cc, run `<b><span style="color:#0AA">cargo tree --invert --package cc@1.1.23</span></b>`
+</pre>
+
+By default `cargo info` describes the package version in the local `Cargo.lock`, if any, but as you can see it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that. The feature list also indicates a `+` next to any that are enabled.
+
```
traviscross, line 47:
````suggestion

### Replacing `Captures` trick with `use<..>` syntax

Rust now supports `use<..>` syntax within certain impl Trait bounds to control which generic lifetime parameters are captured.

Return-position impl Trait (RPIT) types in Rust *capture* certain generic parameters.  Capturing a generic parameter allows that parameter to be used in the hidden type.  That in turn affects borrow checking.

In Rust 2021 and earlier editions, lifetime parameters are not captured in opaque types on bare functions and on functions and methods of inherent impls unless those lifetime parameters are mentioned syntactically in the opaque type.  E.g., this is an error:

```rust
//@ edition: 2021
fn f(x: &()) -> impl Sized { x }
```

```
error[E0700]: hidden type for `impl Sized` captures lifetime that does not appear in bounds
 --> src/main.rs:1:30
  |
1 | fn f(x: &()) -> impl Sized { x }
  |         ---     ----------   ^
  |         |       |
  |         |       opaque type defined here
  |         hidden type `&()` captures the anonymous lifetime defined here
  |
help: add a `use<...>` bound to explicitly capture `'_`
  |
1 | fn f(x: &()) -> impl Sized + use<'_> { x }
  |                            +++++++++
```

With the new `use<..>` syntax, we can fix this, as suggested in the error, by writing:

```rust
fn f(x: &()) -> impl Sized + use<'_> { x }
```

Previously, correctly fixing this class of error required defining a dummy trait, conventionally called `Captures`, and using it as follows:

```rust
trait Captures<T: ?Sized> {}
impl<T: ?Sized, U: ?Sized> Captures<T> for U {}

fn f(x: &()) -> impl Sized + Captures<&'_ ()> { x }
```

That was called "the `Captures` trick", and it was a bit baroque and subtle.  It's no longer needed.

There are some limitations to what is being stabilized.  The `use<..>` syntax cannot currently appear within traits or within trait impls, and it must list all in-scope generic type and const parameters.  We hope to lift these restrictions over time.

Note that in Rust 2024, the example above will "just work" without needing `use<..>` syntax (or any tricks).  This is because in the new edition, opaque types will automatically capture all lifetime parameters in scope.  This is a better default, and we've seen a lot of evidence about how this cleans up code.  In Rust 2024, `use<..>` syntax will serve as an important way of opting-out of that default.

For more details about `use<..>` syntax, capturing, and how this applies to Rust 2024, see the ["RPIT lifetime capture rules"](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/rpit-lifetime-capture.html) chapter of the edition guide.  For details about the overall direction, see our recent blog post, ["Changes to `impl Trait` in Rust 2024"](2024-09-05-impl-trait-capture-rules.md).

````

Comment (traviscross) [reply posts/2024-10-17-Rust-1.82.0.md:47]: Apologies in advance about the length; this is as short as I could make it.

Comment (traviscross) [reply posts/2024-10-17-Rust-1.82.0.md:47]: Rust 1.82 is going to be an unusually large release in terms of user-facing things, so it probably makes sense it will be a longer-than-usual blog post.

File: posts/2024-10-17-Rust-1.82.0.md [hunk 18]
```diff
@@ -0,0 +1,319 @@
...
+    let ptr = &raw const p.not_aligned_field;
+
+    // Accessing the pointer has not changed.
+    // Note that `val = *ptr` would be undefined behavior because
+    // the pointer is not aligned!
+    let val = unsafe { ptr.read_unaligned() };
+}
+```
+
+The native syntax makes it more clear that the operand expression of these operators is interpreted as a [place expression](https://www.ralfj.de/blog/2024/08/14/places.html). It also avoids the term "address-of" when referring to the action of creating a pointer. A pointer is [more than just an address](https://rust-lang.github.io/rfcs/3559-rust-has-provenance.html), so Rust is moving away from terms like "address-of" that reaffirm a false equivalence of pointers and addresses.
```
traviscross, line 128: This is a fairly large user-facing change.  I'd suggest moving this section further toward the top, perhaps after `use<..>`.

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,319 @@
...
+#[unsafe(no_mangle)]
+pub fn my_global_function() { }
+```
+
+The old form of the attribute (without `unsafe`) is currently still accepted, but might be linted against at some point in the future, and will be a hard error in a future edition.
+
+This affects the following attributes:
+- `no_mangle`
+- `link_section`
+- `export_name`
```
traviscross, line 173: This too is a fairly large user-facing change.  Maybe let's move this up after the section about `&raw`.

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,319 @@
...
+```diff
+ static mut STATIC_MUT: Type = Type::new();
+ fn main() {
++    #[allow(unused_unsafe)]
+     let static_mut_ptr = unsafe { std::ptr::addr_of_mut!(STATIC_MUT) };
+ }
+```
+
+A future version of Rust is expected to generalize this to other expressions which would be safe in this position, not just statics.
+
```
traviscross, line 158:
````suggestion

### Unsafe extern

Rust code can use functions and statics from foreign code.  The type signatures of these foreign items are provided in `extern` blocks.  Historically, all items within `extern` blocks have been unsafe to call, but we didn't have to write `unsafe` anywhere on the `extern` block itself.

However, if a signature within the `extern` block is incorrect, then using that item will result in undefined behavior.  Would that be the fault of the person who wrote the `extern` block, or the person who used that item?

We've decided that it's the responsibility of the person writing the `extern` block to ensure that all signatures contained within it are correct, and so we now allow writing `unsafe extern`:

```rust
unsafe extern {
    pub safe fn sqrt(x: f64) -> f64;
    pub unsafe fn strlen(p: *const u8) -> usize;
}
```

One benefit of this is that items within an `unsafe extern` block can be marked as safe to call.  In the above example, we can call `sqrt` without using `unsafe`.  Items that aren't marked with either `safe` or `unsafe` are conservatively assumed to be `unsafe`.

In future releases, we'll be encouraging the use of `unsafe extern` with lints.  Starting in Rust 2024, using `unsafe extern` will be required.

For further details, see [RFC 3484](https://github.com/rust-lang/rfcs/blob/master/text/3484-unsafe-extern-blocks.md) and the ["Unsafe extern blocks"](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-extern.html) chapter of the edition guide.

````

Comment (traviscross) [reply posts/2024-10-17-Rust-1.82.0.md:158]: Unsafe extern is another major user-visible addition in Rust 1.82.  I'd put this ahead of unsafe attributes, and move it further up along with unsafe attributes.

File: posts/2024-10-17-Rust-1.82.0.md [hunk 21]
```diff
@@ -0,0 +1,319 @@
...
+### Unsafe attributes
+
+Some Rust attributes, such as [`no_mangle`](https://doc.rust-lang.org/reference/abi.html#the-no_mangle-attribute), can be used to [cause Undefined Behavior without any `unsafe` block](https://github.com/rust-lang/rust/issues/28179). If this was regular code we would require them to be placed in an `unsafe {}` block, but so far attributes have not had comparable syntax. To reflect the fact that these attributes can undermine Rust's safety guarantees, they are now considered "unsafe" and should be written as follows:
+
+```rust
+#[unsafe(no_mangle)]
+pub fn my_global_function() { }
+```
+
+The old form of the attribute (without `unsafe`) is currently still accepted, but might be linted against at some point in the future, and will be a hard error in a future edition.
```
traviscross, line 168:
```suggestion
The old form of the attribute (without `unsafe`) is currently still accepted, but might be linted against at some point in the future, and will be a hard error in Rust 2024.  For further details, see the ["Unsafe attributes"](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-attributes.html) chapter of the edition guide.
```

posts/2024-10-17-Rust-1.82.0.md
traviscross, line 173:
```suggestion
- `export_name`

For further details, see the ["Unsafe attributes"](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-attributes.html) chapter of the edition guide.
```

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,319 @@
...
+
+### Floating-point NaN semantics and `const`
+
+Operations on floating-point values (of type `f32` and `f64`) are famously subtle. One of the reasons for this is the existence of "NaN values": this is short for "Not a Number", and is used to represent e.g. the result of `0.0 / 0.0`. What makes NaN values subtle is that more than one possible NaN value exists: a NaN value has a sign that can be checked with `f.is_sign_positive()`, and it has a "payload" that can be extracted with `f.to_bits()`. Despite very successful efforts to standardize the behavior of floating-point operations across hardware architectures, the details of when a NaN is positive or negative and what its exact payload is differ across architectures. To make matters even more complicated, Rust and its LLVM backend apply optimizations to floating-point operations when the exact numeric result is guaranteed not to change, but those optimizations can change which NaN value is produced. For instance, `f * 1.0` may be optimized to just `f`. However, if `f` is a NaN, this can change the exact bit pattern of the result!
```
traviscross, line 90:
```suggestion
Operations on floating-point values (of type `f32` and `f64`) are famously subtle. One of the reasons for this is the existence of "NaN values": this is short for "not a number", and is used to represent e.g. the result of `0.0 / 0.0`. What makes NaN values subtle is that more than one possible NaN value exists: a NaN value has a sign that can be checked with `f.is_sign_positive()`, and it has a "payload" that can be extracted with `f.to_bits()`. Despite very successful efforts to standardize the behavior of floating-point operations across hardware architectures, the details of when a NaN is positive or negative and what its exact payload is differ across architectures. To make matters even more complicated, Rust and its LLVM backend apply optimizations to floating-point operations when the exact numeric result is guaranteed not to change, but those optimizations can change which NaN value is produced. For instance, `f * 1.0` may be optimized to just `f`. However, if `f` is a NaN, this can change the exact bit pattern of the result!
```

cuviper:
@traviscross in theory, *everything* we mention in the blog is meant to be "fairly large user-facing," with the remainder left to the release notes alone. Items in the API list may be debatable on largeness, but we just dump that whole. Is there any part we included that perhaps you think is *not* blog worthy?

traviscross:
No, all the lang items there seem important to mention, in the sense that if they weren't there and someone later came back and said, "hey, how could you change that in Rust and not even bother to mention it in the release blog post?", that we'd feel they were right.

traviscross:
Here's how I'd order the relative importance of the lang items:

- `use<..>` syntax
- `&raw` syntax
- Unsafe extern
- Unsafe attributes
- `min_exhaustive_patterns`
- Floats in `const`
- `asm_const`
- Safe raw slices to statics.

posts/2024-10-17-Rust-1.82.0.md
cuviper: I think the problem here, starting from the section title, is that the "`Captures` trick" is (per RFC 3498) "not widely known, and its purpose is not commonly well understood." So I think a title like "Precise capturing with `use<>` syntax" would be more accessible, and it also needs to motivate why I shouldn't just use `+ '_`, **especially** because that's what previous versions recommended for that same example!

AIUI, the problem with outlives is that it may force an unnecessary bound on type parameters, like `T: 'a`, whereas `use<'a, T>` lets those remain independent. Is that right?

posts/2024-10-17-Rust-1.82.0.md
cuviper: Maybe we can raise the `safe` part in the section title? That's a big benefit IMO, and I missed it at first glance. How about "Unsafe extern and safe calls"?

posts/2024-10-17-Rust-1.82.0.md
cuviper:
```suggestion
By default, `cargo info` describes the package version in the local `Cargo.lock`, if any. As you can see, it will indicate when there's a newer version too, and `cargo info cc@1.1.30` would report on that.
```

posts/2024-10-17-Rust-1.82.0.md
traviscross: Good point.  Maybe "Unsafe extern and safe items".

traviscross: Or maybe more pithily, "Safe items in unsafe extern".

posts/2024-10-17-Rust-1.82.0.md
traviscross: Yes, I think in an earlier draft I'd titled the section...

```markdown
### Precise capturing `use<..>` syntax
```

...and that'd still be good.

I'll write something up about the `+ '_` bit.  I'd tried to avoid it here, as it gets somewhat detailed, but you're right about why it should be said.

cuviper:
I think since there's *so much* lang content, we should sandwich it to avoid overshadowing non-lang.

* `cargo info`
* Apple targets
* Lang stuff (in TC's order)
  ...
* Stabilized APIs

posts/2024-10-17-Rust-1.82.0.md
cuviper: Ok, I've shuffled everything else around, but I left this out for now to await your update.

Thanks a lot for your help!

cuviper: BTW, I don't think the relative `md` link for the Changes post will work in the site generator, so that should use the published URL.

traviscross: I incorporated the title change and mentioned the old compiler suggestion.  I'll try to make some other changes late too, but if not.... this would probably be OK given the links to RFC 3498 where this is discussed further.

> AIUI, the problem with outlives is that it may force an unnecessary bound on type parameters, like `T: 'a`, whereas `use<'a, T>` lets those remain independent. Is that right?

Yes... specifically think about invariant lifetimes where this ends up requiring the lifetimes be exactly equal, resulting in no good way for the caller to provide that.

posts/2024-10-17-Rust-1.82.0.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Floating-point NaN semantics and `const`
+
+Operations on floating-point values (of type `f32` and `f64`) are famously subtle. One of the reasons for this is the existence of "NaN values": this is short for "not a number", and is used to represent e.g. the result of `0.0 / 0.0`. What makes NaN values subtle is that more than one possible NaN value exists: a NaN value has a sign that can be checked with `f.is_sign_positive()`, and it has a "payload" that can be extracted with `f.to_bits()`. Despite very successful efforts to standardize the behavior of floating-point operations across hardware architectures, the details of when a NaN is positive or negative and what its exact payload is differ across architectures. To make matters even more complicated, Rust and its LLVM backend apply optimizations to floating-point operations when the exact numeric result is guaranteed not to change, but those optimizations can change which NaN value is produced. For instance, `f * 1.0` may be optimized to just `f`. However, if `f` is a NaN, this can change the exact bit pattern of the result!
```
RalfJung, line 163:
```suggestion
Operations on floating-point values (of type `f32` and `f64`) are famously subtle. One of the reasons for this is the existence of "NaN values": this is short for "not a number", and is used to represent e.g. the result of `0.0 / 0.0`. What makes NaN values subtle is that more than one possible NaN value exists: a NaN value has a sign that can be checked with `f.is_sign_positive()`, and it has a "payload" that can be extracted with `f.to_bits()` -- however, both are entirely ignored by `==` (which always returns `false` on a NaN). Despite very successful efforts to standardize the behavior of floating-point operations across hardware architectures, the details of when a NaN is positive or negative and what its exact payload is differ across architectures. To make matters even more complicated, Rust and its LLVM backend apply optimizations to floating-point operations when the exact numeric result is guaranteed not to change, but those optimizations can change which NaN value is produced. For instance, `f * 1.0` may be optimized to just `f`. However, if `f` is a NaN, this can change the exact bit pattern of the result!
```
Upon re-reading this I felt like it would be good to clarify this.

posts/2024-10-17-Rust-1.82.0.md
traviscross: I've now cleaned this up further, added an example of how `+ '_` doesn't work in general inspired by a real case in the compiler, and pushed this to the branch.

traviscross (approved):

PR #1417, opened 2024-10-18
release notes: clarify that (0.0 / 0.0).is_sign_positive() cannot yet be called in const fn

RalfJung:
When writing this description I didn't realize that the float method constification slipped into the next release. So let's clarify this before people try calling `(0.0 / 0.0).is_sign_positive()` in a const fn and are confused.

[Rendered](https://github.com/RalfJung/blog.rust-lang.org/blob/release-notes-float-const/posts/2024-10-17-Rust-1.82.0.md)


RalfJung:
Cc @traviscross

traviscross (approved):

PR #1420, opened 2024-10-25
Update Alfa Slab One to the current version

LawnGnome:
This should have been done at the same time as https://github.com/rust-lang/www.rust-lang.org/pull/1940, but I forgot that the blog had its own copy of the font files. The same reasoning that applied in that PR applies here: this significantly improves the lower case `s` glyph, and generally makes the font more readable.


ehuss (approved):
Thanks!

PR #1391, opened 2024-09-04
Post: The wasm32-wasip2 target has reached tier 2 support

yoshuawuyts:
Hey all, in Rust 1.81 the `wasm32-wasip2` target reaches tier 2 support. This is a post explaining in more detail what that means, and what people can do with that. This is a follow-up to the post earlier this year where we [introduced the new WASI targets](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html).

I meant to have this post ready a couple of days ago. However life got in the way, and I only managed to finish it basically on the same day Rust 1.81 is going to go live. I was thinking: perhaps we can publish it as a companion post to the release notes (https://github.com/rust-lang/blog.rust-lang.org/pull/1386)? @rust-lang/release how would you feel about that?

cc/ @alexcrichton I implemented all feedback from the earlier draft; can I ask you to give this another skim to make sure the rewrite is correct?

PR #1433, opened 2024-11-21
Post: The `wasm32-wasip2` target has reached tier 2 support (attempt 2) 

yoshuawuyts:
This is a re-submission of https://github.com/rust-lang/blog.rust-lang.org/pull/1391, which we had unfortunately had to unpublished soon after release as documented in https://github.com/rust-lang/blog.rust-lang.org/pull/1397. We've validated that the `wasm32-wasip2` target now actually installs in Rust 1.82, which means we can now re-publish the post.

I've made some minor tweaks to the wording in the post - most notably rewritten it to no longer be in the future tense (the target is tier 2, not will be in the future). So we should probably still have an extra pair of eyes on it to double check it. But otherwise I think this should be good to go.

I've set the date it as Thursday 2024-11-21 - I don't believe that should conflict with any other posts, but let me know if we want to change that.

cc/ @alexcrichton

### DIFF
--- a/posts/2024-11-21-wasip2-tier-2.md
+++ b/posts/2024-11-21-wasip2-tier-2.md
@@ -1,0 +1,99 @@
+---
+layout: post
+title: "The wasm32-wasip2 Target Has Reached Tier 2 Support"
+author: Yosh Wuyts
+---
+
+## Introduction
+
+In April of this year we posted an update about [Rust's WASI
+targets](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html)
+to the main Rust blog. In it we covered the rename of the `wasm32-wasi` target
+to `wasm32-wasip1`, and the introduction of the new `wasm32-wasip2` target as a
+"tier 3" target. This meant that while the target was available as part of
+`rust-lang/rustc`, it was not guaranteed to build. We're pleased to announce
+that this has changed in Rust 1.82.
+
+For those unfamiliar with WebAssembly (Wasm) components and WASI 0.2, here is a quick, simplified primer:
+
+- **Wasm** is a (virtual) instruction format for programs to be compiled into (think: x86).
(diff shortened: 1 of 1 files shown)


posts/2024-11-21-wasip2-tier-2.md
```diff
@@ -0,0 +1,99 @@
...
+[rust-lang/rust#129638](https://github.com/rust-lang/rust/pull/129638) adding
+native support for `std::net` in Rust 1.83. We expect more of these PRs to land
+through the remainder of the year.
+
+Though this doesn't need to stop users from using WASI 0.2 today. The stdlib is
+great because it provides *portable* abstractions, usually built on top of an
+operating system's `libc` or equivalent. If you want to use WASI 0.2 APIs
+directly today, you can either use the
+[wasi](https://docs.rs/wasi/latest/wasi/) crate directly. Or generate your own
+WASI  bindings from the [WASI
```
programmerjake, line 81:
```suggestion
WASI bindings from the [WASI
```

posts/2024-11-21-wasip2-tier-2.md
```diff
@@ -0,0 +1,99 @@
...
+specification's](https://github.com/WebAssembly/WASI/tree/main/wasip2) interface
+types  using [wit-bindgen](https://github.com/bytecodealliance/wit-bindgen/).
```
programmerjake, line 83:
```suggestion
types using [wit-bindgen](https://github.com/bytecodealliance/wit-bindgen/).
```

alexcrichton (approved):

yoshuawuyts:
r? @Mark-Simulacrum

I think? I'm not quite sure who I should be tagging here

Mark-Simulacrum (approved):
Modulo moving the date (maybe today?) this seems fine to me. We have a release this Thursday (1.83) so would prefer to avoid that date.

Mark-Simulacrum (approved):

PR #1436, opened 2024-11-28
Add missing word

SUPERCILEX:
[Rendered](https://github.com/SUPERCILEX/blog.rust-lang.org/blob/patch-1/posts/2024-11-27-Rust-2024-public-testing.md)

PR #1437, opened 2024-11-29
Re-tighten up some verbiage

traviscross:
We had said "further upon X".  We really meant "further X".  This was changed to "further improve upon X" instead in PR #1436, but that's a bit odd as the sentence uses "improvements" directly before it.

Let's tighten this up to just say "further X".

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/TC/retighten-verbiage-in-rust-2024-nightly-beta-announcement/posts/2024-11-27-Rust-2024-public-testing.md)

cc @Kobzol


ehuss (approved):
Thanks!

Kobzol:
Oh, I had no idea that "further" can be used as a verb. The fix looked obvious, but I should have asked, sorry.

PR #1439, opened 2024-12-02
Rustc Trait System Refactor Initiative Update: Stabilizing `-Znext-solver=coherence`

lcnr:
I've set the date to this Wednesday for now.

@rust-lang/initiative-trait-system-refactor @compiler-errors 

[Rendered](https://github.com/lcnr/blog.rust-lang.org/blob/rarw/posts/inside-rust/2024-12-04-trait-system-refactor-initiative.md)

### DIFF
--- a/posts/inside-rust/2024-12-04-trait-system-refactor-initiative.md
+++ b/posts/inside-rust/2024-12-04-trait-system-refactor-initiative.md
@@ -19,7 +19,7 @@ We then moved our focus to [bootstrap using exclusively the new solver](https://
 
 ### Looking ahead
 
-To unblock the stabilization of the next-generation trait solver we're required to fix [multiple](https://github.com/rust-lang/rust/pull/133501) [issues](https://rust.godbolt.org/z/aoseYzMx9) which also affect the current implementation. This is due to subtle differences between the implementations causing the new solver to trigger bugs which are avoided by the old solver. The cases where the new implementation works correctly while the old one does not are unfortunately far less relevant as nobody has written code while relies on the behavior of the next-generation trait solver yet.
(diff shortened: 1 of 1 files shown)


BoxyUwU:
do we definitely want to put this out 6 weeks in advance of stable, just in case anyone files a beta regression and we wind up having to revert?

posts/inside-rust/2024-12-04-trait-system-refactor-initiative.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Looking ahead
+
+To unblock the stabilization of the next-generation trait solver we're required to fix [multiple](https://github.com/rust-lang/rust/pull/133501) [issues](https://rust.godbolt.org/z/aoseYzMx9) which also affect the current implementation. This is due to subtle differences between the implementations causing the new solver to trigger bugs which are avoided by the old solver. The cases where the new implementation works correctly while the old one does not are unfortunately far less relevant as nobody has written code while relies on the behavior of the next-generation trait solver yet.
```
CleanCut, line 22:
```suggestion
To unblock the stabilization of the next-generation trait solver we're required to fix [multiple](https://github.com/rust-lang/rust/pull/133501) [issues](https://rust.godbolt.org/z/aoseYzMx9) which also affect the current implementation. This is due to subtle differences between the implementations causing the new solver to trigger bugs which are avoided by the old solver. The cases where the new implementation works correctly while the old one does not are unfortunately far less relevant as nobody has written code which relies on the behavior of the next-generation trait solver yet.
```

lcnr:
It might get more people to be aware of this on beta, causing us to detect an issue in the first place, allowing us to revert it before it hits stable :3

so yes, i'd like to do so :grin: could change the wording though

lcnr:
rdy for merge, needs approval for it though :thinking:

BoxyUwU (approved):

PR #1443, opened 2024-12-08
Avoid theme flicker on reload in dark mode

JVMerkle:
By loading the theme-switch earlier the nav bar background no longer flickers from bright to dark.

Steps to reproduce:
1. Switch to dark mode
2. Reload page (a couple of times)

### DIFF
--- a/templates/footer.hbs
+++ b/templates/footer.hbs
@@ -44,4 +44,3 @@
 
 <!-- scripts -->
 <script src="{{root}}scripts/highlight.js"></script>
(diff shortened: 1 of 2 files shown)


JVMerkle:
This just "reduces" the symptoms, but I think the theme switch should be reworked somehow, e.g. loading it as early as possible (`<head>`). I notice other pages (posts) also flicker, this is very disturbing while in dark mode.

GuillaumeGomez (approved):

PR #1448, opened 2024-12-26
Add system theme setting

senekor:
The system setting is followed if there is no item `blog-rust-lang-org-theme` in local storage. In order to switch back to the system theme, that item is deleted.

Previously, if a system preference for dark theme was detected, that was additionally stored explicitly as `blog-rust-lang-org-theme`. That was removed, because it prevents the system setting from being followed if the user switches their system back to light theme.

### DIFF
--- a/static/scripts/theme-switch.js
+++ b/static/scripts/theme-switch.js
@@ -1,10 +1,18 @@
 "use strict";
 
 function changeThemeTo(val) {
-    document.documentElement.setAttribute("data-theme", val);
-    // save theme prefs in the browser
-    if (storageAvailable("localStorage")) {
-        localStorage.setItem("blog-rust-lang-org-theme", val);
+    if (val === "system") {
+        setThemeToSystemPref();
+        // delete explicit theme pref from browser storage
+        if (storageAvailable("localStorage")) {
+            localStorage.removeItem("blog-rust-lang-org-theme");
+        }
+    } else {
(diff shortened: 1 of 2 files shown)


static/scripts/theme-switch.js
```diff
@@ -54,9 +70,8 @@ if (storageAvailable("localStorage")) {
 }
 if (savedTheme) {
     document.documentElement.setAttribute("data-theme", savedTheme);
-} else if (window.matchMedia("(prefers-color-scheme: dark)").matches) {
-    document.documentElement.setAttribute("data-theme", "dark");
```
senekor, line 58: @apiraino Do you remember the purpose of this line? I removed it because once the theme preference is stored explicitly, the system preference is not followed anymore if the user changes it later on.

In my testing, I think there was a little bit more flickering without this. Meaning the page was initially white and then quickly flashed to dark according to my system preference. But there was some amount of flickering in both cases, hard to tell the difference.

I'm not sure why that would be the case, unless `window.matchMedia` takes a noticeable amount of time to run (Firefox) ?

apiraino:
@senekor how does this relates to #1443?

GuillaumeGomez (approved):

GuillaumeGomez:
Code looks good to me, thanks!

GuillaumeGomez:
They both seem to work together correctly. This one adds the possibility to reset back to "system settings" whereas the other fixes a bug due to script loading order.

PR #1450, opened 2024-12-26
Eliminate theme flicker on reload

senekor:
previous work / discussions: #1443, #1449

Running the theme switcher JS in the header does seem to fix the problem for me.

cc @GuillaumeGomez @apiraino @JVMerkle


senekor:
@GuillaumeGomez you had concerns about the previous approach, what do you think about this one?

GuillaumeGomez (approved):

GuillaumeGomez:
This approach seems like the right one, thanks!

PR #1451, opened 2024-12-29
1.84.0 release post

Mark-Simulacrum:
[Rendered](https://github.com/Mark-Simulacrum/blog.rust-lang.org/blob/blog-1.84.0/posts/2025-01-09-Rust-1.84.0.md)

### DIFF
--- a/posts/2025-01-09-Rust-1.84.0.md
+++ b/posts/2025-01-09-Rust-1.84.0.md
@@ -69,7 +69,7 @@ automatically use older library versions compatible with their older toolchain.
 
 See the [documentation](https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version) for more considerations when deciding on an MSRV policy.
 
-### Migration to a new trait solver begins
+### Migration to the new trait solver begins
 
 The Rust compiler is in the process of moving to a new implementation for the
 trait solver. The next-generation trait solver is a reimplementation of a core
@@ -81,16 +81,15 @@ underlying type of `<Vec<T> as IntoIterator>::Item` - and equating types
 
 In 1.84, the new solver is used for checking coherence of trait impls. At a
 high level, coherence is responsible for ensuring that there is at most one
-implementation of a trait for a given type, including *globally* in not yet
-written or visible code in downstream crates from the current compilation.
+implementation of a trait for a given type while considering not yet written
+or visible code from other crates.
 
-This stabilization does include some breaking changes, primarily by fixing some
-correctness issues with the old solver. Typically these will show up as new
-"conflicting implementations of trait ..." errors that were not previously
-reported. We expect instances of this to be rare based on evaluation of
-available code through [Crater], as the soundness holes in the previous solving
-engine used relatively esoteric code. It also improves our ability to detect
-where impls do *not* overlap, allowing more code to be written in some cases.
+This stabilization fixes a few mostly theoretical correctness issues of the
+old implementation, resulting in potential "conflicting implementations of trait ..."
+errors that were not previously reported. We expect the affected patterns to be
+very rare based on evaluation of available code through [Crater]. The stabilization
+also improves our ability to prove that impls do *not* overlap, allowing more code
+to be written in some cases.
 
 For more details, see a [previous blog post](https://blog.rust-lang.org/inside-rust/2024/12/04/trait-system-refactor-initiative.html)
 and the [stabilization report](https://github.com/rust-lang/rust/pull/130654).
--- a/posts/2025-01-09-Rust-1.84.0.md
+++ b/posts/2025-01-09-Rust-1.84.0.md
@@ -19,20 +19,23 @@ If you'd like to help us out by testing future releases, you might consider upda
 
 ## What's in 1.84.0 stable
 
-### Cargo can use toolchain version for library version selection
+### Cargo considers Rust versions for dependency version selection 
 
 1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
 which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
 dependencies, if available, to improve package version selection. Rust version
 aware selection allows library authors to easily adopt newer Rust versions
-while avoiding breaking consumers needing compatibility with older toolchains.
+while avoiding breaking consumers who want to use older Rust compilers.
 
 You can opt-in to the MSRV-aware resolver via [`.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions):
+
 ```toml
 [resolver]
 incompatible-rust-versions = "fallback"
 ```
+
 Then when adding a dependency:
+
 ```console
 $ cargo add clap
     Updating crates.io index
@@ -44,6 +47,7 @@ warning: ignoring clap@4.5.23 (which requires rustc 1.74) to maintain demo's rus
 ```
 
 When [verifying the latest dependencies in CI](https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-latest-dependencies), you can override this:
+
 ```console
 $ CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=allow cargo update
     Updating crates.io index
@@ -60,10 +64,10 @@ toolchain (and avoid running `cargo update`). Now, those users will be able to
 automatically use older library versions compatible with their older toolchain.
 In the future, we expect this to provide more flexibility for library authors
 to select their preferred support strategy for Rust versions, with the toolchain
-helping users on older toolchains avoid breakage.
+helping users on older toolchains avoid breakage. See also the [detailed guidance](https://doc.rust-lang.org/beta/cargo/reference/rust-version.html#setting-and-updating-rust-version)
+on what to consider when making a choice of supported version.
 
-The new resolver will be enabled by default with the 2024 edition (expected to
-stabilize in 1.85), but can be enabled as of 1.84 by setting
+The new resolver will be enabled by default when optin into the 2024 edition (will stabilize in 1.85), but can be enabled as of 1.84 by setting
 [`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
 [`resolver.incompatible-rust-versions = "fallback"`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) in the Cargo configuration file.
 
@@ -93,7 +97,7 @@ engine used relatively esoteric code. It also improves our ability to detect
 where impls do *not* overlap, allowing more code to be written in some cases.
 
 For more details, see a [previous blog post](https://blog.rust-lang.org/inside-rust/2024/12/04/trait-system-refactor-initiative.html)
-and the [stabilization report](https://github.com/rust-lang/rust/pull/130654)
+and the [stabilization report](https://github.com/rust-lang/rust/pull/130654).
 
 [Crater]: https://github.com/rust-lang/crater/
 
--- a/posts/2025-01-09-Rust-1.84.0.md
+++ b/posts/2025-01-09-Rust-1.84.0.md
@@ -23,7 +23,7 @@ If you'd like to help us out by testing future releases, you might consider upda
 
 1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
 which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
-dependencies, if available, to improve package version selection. Rust version
+dependencies, if available, to improve package version selection. MSRV-aware
 aware selection allows library authors to easily adopt newer Rust versions
 while avoiding breaking consumers who want to use older Rust compilers.
 
--- a/posts/2025-01-09-Rust-1.84.0.md
+++ b/posts/2025-01-09-Rust-1.84.0.md
@@ -76,17 +76,16 @@ and the [stabilization report](https://github.com/rust-lang/rust/pull/130654)
 
 ### Strict provenance APIs
 
-In Rust, [pointers are not simply an “integer” or
-“address”](https://rust-lang.github.io/rfcs/3559-rust-has-provenance.html). For
-instance, it’s uncontroversial to say that a Use After Free is clearly
-Undefined Behavior, even if you “get lucky” and the freed memory gets
-reallocated before your read/write. It is also uncontroversial that writing
-through a pointer derived from an `&i32` reference is Undefined Behavior, even
+In Rust, [pointers are not simply an "integer" or
+"address"](https://rust-lang.github.io/rfcs/3559-rust-has-provenance.html). For
+instance, a "use after free" is undefined behavior even if you "get lucky" and the freed memory gets
+reallocated before your read/write. As another example, writing
+through a pointer derived from an `&i32` reference is undefined behavior, even
 if writing to the same address via a different pointer is legal. The underlying
 pattern here is that *the way a pointer is computed matters*, not just the
 address that results from this computation. For this reason, we say that
-pointers have **provenance**: to fully characterize pointer-related Undefined
-Behavior in Rust, we have to know not only the address the pointer points to,
+pointers have **provenance**: to fully characterize pointer-related undefined
+behavior in Rust, we have to know not only the address the pointer points to,
 but also track which other pointer(s) it is "derived from".
 
 Most of the time, programmers do not need to worry much about provenance, and
--- a/posts/2025-01-09-Rust-1.84.0.md
+++ b/posts/2025-01-09-Rust-1.84.0.md
@@ -128,7 +128,48 @@ For more details, see the standard library [documentation on provenance](https:/
 
 ### Stabilized APIs
 
-TODO, relnotes still in-progress for this section
+- [`Ipv6Addr::is_unique_local`](https://doc.rust-lang.org/stable/core/net/struct.Ipv6Addr.html#method.is_unique_local)
+- [`Ipv6Addr::is_unicast_link_local`](https://doc.rust-lang.org/stable/core/net/struct.Ipv6Addr.html#method.is_unicast_link_local)
+- [`core::ptr::with_exposed_provenance`](https://doc.rust-lang.org/stable/core/ptr/fn.with_exposed_provenance.html)
+- [`core::ptr::with_exposed_provenance_mut`](https://doc.rust-lang.org/stable/core/ptr/fn.with_exposed_provenance_mut.html)
+- [`<ptr>::addr`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.addr)
+- [`<ptr>::expose_provenance`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.expose_provenance)
+- [`<ptr>::with_addr`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.with_addr)
+- [`<ptr>::map_addr`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.map_addr)
+- [`<int>::isqrt`](https://doc.rust-lang.org/stable/core/primitive.i32.html#method.isqrt)
+- [`<int>::checked_isqrt`](https://doc.rust-lang.org/stable/core/primitive.i32.html#method.checked_isqrt)
+- [`<uint>::isqrt`](https://doc.rust-lang.org/stable/core/primitive.u32.html#method.isqrt)
+- [`NonZero::isqrt`](https://doc.rust-lang.org/stable/core/num/struct.NonZero.html#impl-NonZero%3Cu128%3E/method.isqrt)
+- [`core::ptr::without_provenance`](https://doc.rust-lang.org/stable/core/ptr/fn.without_provenance.html)
+- [`core::ptr::without_provenance_mut`](https://doc.rust-lang.org/stable/core/ptr/fn.without_provenance_mut.html)
+- [`core::ptr::dangling`](https://doc.rust-lang.org/stable/core/ptr/fn.dangling.html)
+- [`core::ptr::dangling_mut`](https://doc.rust-lang.org/stable/core/ptr/fn.dangling_mut.html)
+
+These APIs are now stable in const contexts
+
+- [`AtomicBool::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicBool.html#method.from_ptr)
+- [`AtomicPtr::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicPtr.html#method.from_ptr)
+- [`AtomicU8::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicU8.html#method.from_ptr)
+- [`AtomicU16::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicU16.html#method.from_ptr)
+- [`AtomicU32::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicU32.html#method.from_ptr)
+- [`AtomicU64::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicU64.html#method.from_ptr)
+- [`AtomicUsize::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicUsize.html#method.from_ptr)
+- [`AtomicI8::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicI8.html#method.from_ptr)
+- [`AtomicI16::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicI16.html#method.from_ptr)
+- [`AtomicI32::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicI32.html#method.from_ptr)
+- [`AtomicI64::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicI64.html#method.from_ptr)
+- [`AtomicIsize::from_ptr`](https://doc.rust-lang.org/stable/core/sync/atomic/struct.AtomicIsize.html#method.from_ptr)
+- [`<ptr>::is_null`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.is_null-1)
+- [`<ptr>::as_ref`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.as_ref-1)
+- [`<ptr>::as_mut`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.as_mut)
+- [`Pin::new`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.new)
+- [`Pin::new_unchecked`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.new_unchecked)
+- [`Pin::get_ref`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.get_ref)
+- [`Pin::into_ref`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.into_ref)
+- [`Pin::get_mut`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.get_mut)
+- [`Pin::get_unchecked_mut`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.get_unchecked_mut)
+- [`Pin::static_ref`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.static_ref)
+- [`Pin::static_mut`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.static_mut)
 
 ### Other changes
 
--- a/posts/2025-01-09-Rust-1.84.0.md
+++ b/posts/2025-01-09-Rust-1.84.0.md
@@ -56,6 +56,9 @@ $ CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=allow cargo update
     Updating clap v4.0.32 -> v4.5.23
 ```
 
+You can also opt-in by setting [`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file though that will require raising your MSRV to 1.84.  The new resolver will be enabled by default for projects using the 2024 edition
+(which will stabilize in 1.85).
+
 This gives library authors more flexibility when deciding
 their policy on adopting new Rust toolchain features. Previously, a library
 adopting features from a new Rust toolchain would force downstream users of
@@ -65,11 +68,6 @@ toolchain (and avoid running `cargo update`). Now, those users will be able to
 automatically use older library versions compatible with their older toolchain.
 
 See the [documentation](https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version) for more considerations when deciding on an MSRV policy.
-
-The new resolver will be enabled by default for projects using the 2024 edition
-(which will stabilize in 1.85), but can be enabled in this release via
-[`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
-[`resolver.incompatible-rust-versions = "fallback"`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) in the Cargo configuration file.
 
 ### Migration to a new trait solver begins
 
--- a/posts/2025-01-09-Rust-1.84.0.md
+++ b/posts/2025-01-09-Rust-1.84.0.md
@@ -42,8 +42,6 @@ stabilize in 1.85), but can be enabled as of 1.84 by setting
 [`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
 [`resolver.incompatible-rust-versions = "fallback"`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) in the Cargo configuration file.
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
cc https://github.com/rust-lang/rust/issues/133210 @lcnr @rust-lang/types -- I expanded/reworded a bit the stub you had started for new trait solver in coherence, would definitely appreciate review to make sure I didn't get the details too wrong :)

cc @joshtriplett  @rust-lang/cargo for the section on MSRV-aware resolver, I sort of semi-randomly came up with recommendations and general prose but very happy to replace it if there's something better or actual recommendations the team wants to offer :)

cc @rust-lang/opsem for the strict provenance section -- I tried to find a balance between copying in the whole contents of the standard library docs and referencing them, very open to changes though.

Mark-Simulacrum:
cc @rust-lang/release as well...

jieyouxu:
cc @alex-semenyuk @traviscross for relnotes blog post (I am triagebot)

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,110 @@
...
+requiring an older Rust version to avoid running `cargo update` and/or require
+pinning the version of the library in use, but now those consumers will be able
+to automatically avoid pulling in new library versions if they want
+compatibility with an older toolchain. We expect this to provide more options
+for library authors for picking their preferred support strategy for Rust
+versions.
+
+The new resolver will be enabled by default with the 2024 edition (expected to
+stabilize in 1.85), but can be enabled as of 1.84 by setting
+`resolver.incompatible-rust-versions = "fallback"`.
```
weihanglo, line 42: To me, setting it in manifest is a bit better than in config file. Not every project checks in `.cargo/config.toml`.

```suggestion
[`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
[`resolver.incompatible-rust-versions = "fallback"`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) in the Cargo configuration file.
```

pietroalbini, line 42: Would be nice to list where.

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,110 @@
...
+### Cargo can use toolchain version for library version selection
+
+1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
+which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
+dependencies, if available, to improve package version selection. Rust version
+aware selection allows library authors to easily adopt newer Rust versions
+while providing library consumers a way to opt-in to using new package versions
+if they need compatibility with older toolchains.
+
+Library authors should start declaring their MSRV. Previously, bumping the
```
weihanglo, line 31: Instead of urging people setting MSRV, I might put focus on thinking their MSRV “policy” under a reasonable consideration, and then set the value.
Cargo doc does provide [expectations and recommendations](https://doc.rust-lang.org/cargo/reference/rust-version.html) to help lib authors find their own policies, though I don't know how to put that properly in this paragraph 😞.

Anyway, thanks for writing this!

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,110 @@
...
+
+## What's in 1.84.0 stable
+
+### Cargo can use toolchain version for library version selection
+
+1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
+which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
+dependencies, if available, to improve package version selection. Rust version
+aware selection allows library authors to easily adopt newer Rust versions
+while providing library consumers a way to opt-in to using new package versions
```
joshtriplett, line 28:
```suggestion
while providing library consumers a way to opt-in to using old package versions
```

RalfJung:
@Mark-Simulacrum I already spent some effort on Strict Provenance release notes in https://github.com/rust-lang/rust/issues/131996, did you see that? I thought that is the intended place for such text.

joshtriplett:
cc @epage

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,110 @@
...
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.84.0](https://doc.rust-lang.org/stable/releases.html#version-1840-2025-01-09).
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.84.0 stable
+
+### Cargo can use toolchain version for library version selection
+
+1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
```
joshtriplett, line 24:
```suggestion
1.84.0 stabilizes the MSRV-aware resolver,
```

The acronym is explained by the link later in the sentence, and avoiding the expansion *here* allows hyphenating "MSRV-aware".

posts/2025-01-09-Rust-1.84.0.md
pietroalbini: Covered in https://github.com/rust-lang/blog.rust-lang.org/pull/1451#discussion_r1899161739.

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,110 @@
...
+if they need compatibility with older toolchains.
+
+Library authors should start declaring their MSRV. Previously, bumping the
+version at each release to latest stable would force downstream consumers
+requiring an older Rust version to avoid running `cargo update` and/or require
+pinning the version of the library in use, but now those consumers will be able
+to automatically avoid pulling in new library versions if they want
+compatibility with an older toolchain. We expect this to provide more options
+for library authors for picking their preferred support strategy for Rust
+versions.
```
joshtriplett, line 38:
```suggestion
Library authors should take the MSRV-aware resolver into account when deciding their policy on adopting new Rust toolchain features. Previously, a library adopting features
from a new Rust toolchain would force downstream users of that library who have an older Rust version to either upgrade their toolchain or manually select an old version of the library compatible with their toolchain (and avoid running `cargo update`). Now, those users will be able to automatically use older library versions compatible with their older toolchain. In the future, we expect this to provide more flexibility for library authors to select their preferred support strategy for Rust versions, without worrying about users on older toolchains.
```

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,110 @@
...
+
+The new resolver will be enabled by default with the 2024 edition (expected to
+stabilize in 1.85), but can be enabled as of 1.84 by setting
+`resolver.incompatible-rust-versions = "fallback"`.
+
+TODO: Should we talk about resolver v3 and/or recommend that as the path to enabling instead?
+
```
joshtriplett, line 45:
```suggestion
```

@weihanglo's suggested change addresses this.

Mark-Simulacrum:
> @Mark-Simulacrum I already spent some effort on Strict Provenance release notes in [rust-lang/rust#131996](https://github.com/rust-lang/rust/issues/131996), did you see that? I thought that is the intended place for such text.

Thanks, I had seen it previously but missed it when putting together the blog (just my bad). Copied that text in, it looks good to me, thanks for writing it up!

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.84.0](https://doc.rust-lang.org/stable/releases.html#version-1840-2025-01-09).
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.84.0 stable
+
+### Cargo can use toolchain version for library version selection
```
shepmaster, line 22: “A toolchain”? “The toolchain”?

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Migration to a new trait solver begins
+
+The Rust compiler is in the process of moving to a new implementation for the
+trait solver. The next-generation trait solver is a reimplementation of a core
+component of Rust's type system. It is not only responsible for checking
+whether trait-bounds - e.g. `Vec<T>: Clone` - hold, but is also used by many
+other parts of the type system, such as normalization - figuring out the
+underlying type of `<Vec<T> as IntoIterator>::Item` - and equating types
+(checking whether T and U are the same).
```
shepmaster, line 57:
```suggestion
(checking whether `T` and `U` are the same).
```

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.84.0 stable
+
+### Cargo can use toolchain version for library version selection
+
+1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
+which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
+dependencies, if available, to improve package version selection. Rust version
```
alex-semenyuk, line 26: MSRV-aware

jieyouxu: Already mentioned here https://github.com/rust-lang/blog.rust-lang.org/pull/1451#discussion_r1899170074

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+or manually select an old version of the library compatible with their
+toolchain (and avoid running `cargo update`). Now, those users will be able to
+automatically use older library versions compatible with their older toolchain.
+In the future, we expect this to provide more flexibility for library authors
+to select their preferred support strategy for Rust versions, with the toolchain
+helping users on older toolchains avoid breakage.
+
+The new resolver will be enabled by default with the 2024 edition (expected to
+stabilize in 1.85), but can be enabled as of 1.84 by setting
+[`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
```
alex-semenyuk, line 44: It's nothing mentioned at the link about the version 3 feature resolver but there is about version 2 https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2 might be good idea to add there in future

weihanglo: When beta becomes stable, it will be <https://doc.rust-lang.org/beta/cargo/reference/resolver.html#resolver-versions>.

posts/2025-01-09-Rust-1.84.0.md
alex-semenyuk: Yes, I meant line 26 as well 

```suggestion
dependencies, if available, to improve package version selection. MSRV-aware
```

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+instance, it’s uncontroversial to say that a Use After Free is clearly
+Undefined Behavior, even if you “get lucky” and the freed memory gets
+reallocated before your read/write. It is also uncontroversial that writing
+through a pointer derived from an `&i32` reference is Undefined Behavior, even
+if writing to the same address via a different pointer is legal. The underlying
+pattern here is that *the way a pointer is computed matters*, not just the
+address that results from this computation. For this reason, we say that
+pointers have **provenance**: to fully characterize pointer-related Undefined
+Behavior in Rust, we have to know not only the address the pointer points to,
+but also track which other pointer(s) it is "derived from".
```
RalfJung, line 90: Some word-smithing on a re-read.
```suggestion
In Rust, [pointers are not simply an “integer” or
“address”](https://rust-lang.github.io/rfcs/3559-rust-has-provenance.html). For
instance, a Use After Free is Undefined Behavior even if you “get lucky” and the freed memory gets
reallocated before your read/write. As another example, writing
through a pointer derived from an `&i32` reference is Undefined Behavior, even
if writing to the same address via a different pointer is legal. The underlying
pattern here is that *the way a pointer is computed matters*, not just the
address that results from this computation. For this reason, we say that
pointers have **provenance**: to fully characterize pointer-related Undefined
Behavior in Rust, we have to know not only the address the pointer points to,
but also track which other pointer(s) it is "derived from".
```

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+Library authors should take the MSRV-aware resolver into account when deciding
+their policy on adopting new Rust toolchain features. Previously, a library
+adopting features from a new Rust toolchain would force downstream users of
+that library who have an older Rust version to either upgrade their toolchain
+or manually select an old version of the library compatible with their
+toolchain (and avoid running `cargo update`). Now, those users will be able to
+automatically use older library versions compatible with their older toolchain.
+In the future, we expect this to provide more flexibility for library authors
+to select their preferred support strategy for Rust versions, with the toolchain
+helping users on older toolchains avoid breakage.
```
epage, line 40: If we want to provide more context on setting an MSRV, https://doc.rust-lang.org/cargo/reference/rust-version.html is where we have that documented and is fairly new so people are likely to have missed it

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+
+The new resolver will be enabled by default with the 2024 edition (expected to
+stabilize in 1.85), but can be enabled as of 1.84 by setting
+[`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
+[`resolver.incompatible-rust-versions = "fallback"`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) in the Cargo configuration file.
```
epage, line 45: *(opening a new thread since https://github.com/rust-lang/blog.rust-lang.org/pull/1451#discussion_r1899161739 is resolved and don't want this missed)*

> TODO: Should we talk about resolver v3 and/or recommend that as the path to enabling instead?

@weihanglo in https://github.com/rust-lang/blog.rust-lang.org/pull/1451#discussion_r1899161739

> To me, setting it in manifest is a bit better than in config file. Not every project checks in `.cargo/config.toml`.

The emphasis should be on the config but with specific context.

It is true that we generally prefer project state to be in the manifest and transient state to be in the config file.  This would generally be a project default that people would want to create a happy path for contributors.

However, if we only talk about the manifest, people will assume they can only use this feature if they bump their MSRV which would severely hamper the adoption of this feature as people with an MSRV are unlikely to bump it to gain access to this.  We also need to acknowledge the config to ensure people know they can override this in CI for testing the [latest dependencies](https://doc.rust-lang.org/nightly/cargo/guide/continuous-integration.html#verifying-latest-dependencies).

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,117 @@
...
+### Cargo can use toolchain version for library version selection
+
+1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
+which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
+dependencies, if available, to improve package version selection. Rust version
+aware selection allows library authors to easily adopt newer Rust versions
+while allowing consumers of the library to automatically use old versions
+if they need compatibility with older toolchains.
+
+Library authors should take the MSRV-aware resolver into account when deciding
```
epage, line 31: This section is a wall of text and I feel like demonstrating what was just discussed would be a big help.

````suggestion
if they need compatibility with older toolchains.

You can opt-in to the MSRV-aware resolver via [`.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions):
```toml
[resolver]
incompatible-rust-versions = "fallback"
```
Then when adding a dependency:
```console
$ cargo add clap
    Updating crates.io index
warning: ignoring clap@4.5.23 (which requires rustc 1.74) to maintain demo's rust-version of 1.60
      Adding clap v4.0.32 to dependencies
    Updating crates.io index
     Locking 33 packages to latest Rust 1.60 compatible versions
      Adding clap v4.0.32 (available: v4.5.23, requires Rust 1.74)
```

When [verifying the latest dependencies in CI](https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-latest-dependencies), you can override this:
```console
$ CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=allow cargo update
    Updating crates.io index
     Locking 12 packages to latest compatible versions
    Updating clap v4.0.32 -> v4.5.23
```

Library authors should take the MSRV-aware resolver into account when deciding
````
Notes:
- `cargo add` using MSRV for selecting version requirements isn't new (the warning) but the `Locking to latest Rust 1.60 compatible version` part is
- I edited the output for brevity

File: posts/2025-01-09-Rust-1.84.0.md (hunk 11 above)
Comment (traviscross) [reply posts/2025-01-09-Rust-1.84.0.md:90]:
```suggestion
In Rust, [pointers are not simply an “integer” or
“address”](https://rust-lang.github.io/rfcs/3559-rust-has-provenance.html). For
instance, it’s uncontroversial to say that a "use after free" is clearly
undefined behavior, even if you “get lucky” and the freed memory gets
reallocated before your read/write. It is also uncontroversial that writing
through a pointer derived from an `&i32` reference is undefined behavior, even
if writing to the same address via a different pointer is legal. The underlying
pattern here is that *the way a pointer is computed matters*, not just the
address that results from this computation. For this reason, we say that
pointers have **provenance**: to fully characterize pointer-related undefined
behavior in Rust, we have to know not only the address the pointer points to,
but also track which other pointer(s) it is "derived from".
```

RalfJung: That conflicts with my edits...

traviscross: Ah, oops, yes...

traviscross:
```suggestion
In Rust, [pointers are not simply an "integer" or
"address"](https://rust-lang.github.io/rfcs/3559-rust-has-provenance.html). For
instance, a "use after free" is undefined behavior even if you "get lucky" and the freed memory gets
reallocated before your read/write. As another example, writing
through a pointer derived from an `&i32` reference is undefined behavior, even
if writing to the same address via a different pointer is legal. The underlying
pattern here is that *the way a pointer is computed matters*, not just the
address that results from this computation. For this reason, we say that
pointers have **provenance**: to fully characterize pointer-related undefined
behavior in Rust, we have to know not only the address the pointer points to,
but also track which other pointer(s) it is "derived from".
```

posts/2025-01-09-Rust-1.84.0.md
Mark-Simulacrum: Typically you want to introduce an acronym *before* you use it, right? I think if we change this we'd lose any direct equivalence statement (i.e., "expanded" = "MSRV") -- you'd have to click through to figure it out. MSRV doesn't feel *that* well known a term to me, so I think writing it out is better for the first reference.

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,140 @@
...
+## What's in 1.84.0 stable
+
+### Cargo can use toolchain version for library version selection
+
+1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
+which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
+dependencies, if available, to improve package version selection. Rust version
+aware selection allows library authors to easily adopt newer Rust versions
+while allowing consumers of the library to automatically use old versions
+if they need compatibility with older toolchains.
```
Mark-Simulacrum, line 29:
```suggestion
while avoiding breaking consumers needing compatibility with older toolchains.
```

Rewording slightly to avoid "allows ... allowing".

posts/2025-01-09-Rust-1.84.0.md
Mark-Simulacrum: I think we link that already in the first sentence?

posts/2025-01-09-Rust-1.84.0.md
Mark-Simulacrum: This header isn't ideal IMO -- but I'm not sure adding an article will help. It's hard for me to succinctly describe what this feature is without using jargon (e.g., MSRV), especially since it's *not* on by default.

posts/2025-01-09-Rust-1.84.0.md
epage: The difference is that is linking to it to provide context for the field while if we linked to it here, it would be to call out that guidance on MSRV policy exists.  While the lead up sections are helpful, we could even link directly to https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version

As I said, "if we want to provide more context".

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,139 @@
...
+toolchain (and avoid running `cargo update`). Now, those users will be able to
+automatically use older library versions compatible with their older toolchain.
+In the future, we expect this to provide more flexibility for library authors
+to select their preferred support strategy for Rust versions, with the toolchain
+helping users on older toolchains avoid breakage.
+
+The new resolver will be enabled by default with the 2024 edition (expected to
+stabilize in 1.85), but can be enabled as of 1.84 by setting
+[`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
+[`resolver.incompatible-rust-versions = "fallback"`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) in the Cargo configuration file.
```
epage, line 68: With the new example, I feel like we can trim this down
```suggestion
The new resolver will be enabled by default with the 2024 edition (expected to
stabilize in 1.85).
```

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,139 @@
...
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.84.0 stable
+
+### Cargo can use toolchain version for library version selection
+
+1.84.0 stabilizes the minimum supported Rust version (MSRV) aware resolver,
+which uses the declared [minimum supported Rust version](https://doc.rust-lang.org/cargo/reference/rust-version.html) from
+dependencies, if available, to improve package version selection. Rust version
```
epage, line 26: We've found some people assuming the MSRV-aware resolver will respect your current Rust version and not your `package.rust-version`.  Instead your toolchain version is only a fallback for when your you don't set a `package.rust-version`.

```suggestion
which prefers dependency versions that have declared a [MSRV](https://doc.rust-lang.org/cargo/reference/rust-version.html) that is compatible with your project's MSRV. Rust version
```
(switched the link to use the acronym since it is now introduced earlier)

Similarly, I wonder if the title will encourage this misunderstanding but don't have much of a suggestion.

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,139 @@
...
+aware selection allows library authors to easily adopt newer Rust versions
+while avoiding breaking consumers needing compatibility with older toolchains.
```
epage, line 28: The MSRV section focuses only on the effect on library authors and only in adopting newer MSRVs.  This also helps those still supporting old versions and for application authors who need to target older toolchains.  Since we cover the freedom this buys library authors later on, maybe we can remove this section for focus on the more immediate improvement.
```suggestion
dependencies, if available, to improve package version selection. With MSRV-aware version selection, the toil is reduced for maintainers to support older toolchains by not needing to manually select older versions for each dependency.
```

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,139 @@
...
+Library authors should take the MSRV-aware resolver into account when deciding
+their policy on adopting new Rust toolchain features. Previously, a library
+adopting features from a new Rust toolchain would force downstream users of
+that library who have an older Rust version to either upgrade their toolchain
+or manually select an old version of the library compatible with their
+toolchain (and avoid running `cargo update`). Now, those users will be able to
+automatically use older library versions compatible with their older toolchain.
+In the future, we expect this to provide more flexibility for library authors
+to select their preferred support strategy for Rust versions, with the toolchain
+helping users on older toolchains avoid breakage.
```
epage, line 63: While I would love for library authors to bump their MSRV more aggressively (maybe some packages can finally drop Rust 1.31 support...), I feel like this prescriptive language might be off putting to people who support an older MSRV.

Maybe if we replace this with the talk of flexibility that comes later
```suggestion
This gives library authors more flexibility when deciding
their policy on adopting new Rust toolchain features. Previously, a library
adopting features from a new Rust toolchain would force downstream users of
that library who have an older Rust version to either upgrade their toolchain
or manually select an old version of the library compatible with their
toolchain (and avoid running `cargo update`). Now, those users will be able to
automatically use older library versions compatible with their older toolchain.
See the [documentation](https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version) for more considerations when deciding on an MSRV policy.
```

posts/2025-01-09-Rust-1.84.0.md
Mark-Simulacrum: I think the example doesn't reference Cargo.toml setting so this is probably still needed in the expanded form?

posts/2025-01-09-Rust-1.84.0.md
Mark-Simulacrum: It's not clear to me what changes you're specifically looking for here (especially with a bunch of the rewording already done) -- maybe you can suggest something specific?

Mark-Simulacrum:
OK, I think that incorporates the feedback I have clear solutions for. I'm still not super happy with the header for the MSRV section, but I don't have any better ideas than the tweak I applied just now.

posts/2025-01-09-Rust-1.84.0.md
epage: The main focus of my comment is on emphasizing the config which is now done so people find a way to use this without bumping MSRV.

It would be nice to find a way to tell people that they don't need to bump MSRV to use this feature but I can't think of a succinct enough way of doing it.

We can probably mark this as resolved.

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,182 @@
...
+or manually select an old version of the library compatible with their
+toolchain (and avoid running `cargo update`). Now, those users will be able to
+automatically use older library versions compatible with their older toolchain.
+
+See the [documentation](https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version) for more considerations when deciding on an MSRV policy.
+
+The new resolver will be enabled by default for projects using the 2024 edition
+(which will stabilize in 1.85), but can be enabled in this release via
+[`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file, or
+[`resolver.incompatible-rust-versions = "fallback"`](https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions) in the Cargo configuration file.
```
epage, line 72: An alternative to https://github.com/rust-lang/blog.rust-lang.org/pull/1451#discussion_r1900280569

```suggestion
You can also opt-in by setting [`package.resolver = "3"`](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) in the Cargo.toml manifest file though that will require raising your MSRV to 1.84.  The new resolver will be enabled by default for projects using the 2024 edition
(which will stabilize in 1.85).

This gives library authors more flexibility when deciding
their policy on adopting new Rust toolchain features. Previously, a library
adopting features from a new Rust toolchain would force downstream users of
that library who have an older Rust version to either upgrade their toolchain
or manually select an old version of the library compatible with their
toolchain (and avoid running `cargo update`). Now, those users will be able to
automatically use older library versions compatible with their older toolchain.

See the [documentation](https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version) for more considerations when deciding on an MSRV policy.
```
This removes the redundant talk of the config and groups the ways that this will be enabled.

posts/2025-01-09-Rust-1.84.0.md
epage: Saying that it can opt-in with the config is redundant with the earlier section.  That just leaves `package.resolver` which feels odd to leave dangling.

I posted an alternative suggestion at https://github.com/rust-lang/blog.rust-lang.org/pull/1451/files#r1902126387

Feel free to take that or close both

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,180 @@
...
+"conflicting implementations of trait ..." errors that were not previously
+reported. We expect instances of this to be rare based on evaluation of
+available code through [Crater], as the soundness holes in the previous solving
+engine used relatively esoteric code. It also improves our ability to detect
+where impls do *not* overlap, allowing more code to be written in some cases.
+
+For more details, see a [previous blog post](https://blog.rust-lang.org/inside-rust/2024/12/04/trait-system-refactor-initiative.html)
+and the [stabilization report](https://github.com/rust-lang/rust/pull/130654).
+
+[Crater]: https://github.com/rust-lang/crater/
```
lcnr, line 98:
```suggestion
### Migration to the new trait solver begins

The Rust compiler is in the process of moving to a new implementation for the
trait solver. The next-generation trait solver is a reimplementation of a core
component of Rust's type system. It is not only responsible for checking
whether trait-bounds - e.g. `Vec<T>: Clone` - hold, but is also used by many
other parts of the type system, such as normalization - figuring out the
underlying type of `<Vec<T> as IntoIterator>::Item` - and equating types
(checking whether `T` and `U` are the same).

In 1.84, the new solver is used for checking coherence of trait impls. At a
high level, coherence is responsible for ensuring that there is at most one
implementation of a trait for a given type while considering not yet written
or visible code from other crates.

This stabilization fixes a few mostly theoretical correctness issues of the
old implementation, resulting in potential "conflicting implementations of trait ..."
errors that were not previously reported. We expect the affected patterns to be
very rare based on evaluation of available code through [Crater]. The stabilization
also improves our ability to prove that impls do *not* overlap, allowing more code
to be written in some cases.

For more details, see a [previous blog post](https://blog.rust-lang.org/inside-rust/2024/12/04/trait-system-refactor-initiative.html)
and the [stabilization report](https://github.com/rust-lang/rust/pull/130654).

[Crater]: https://github.com/rust-lang/crater/
```
Thank you @Mark-Simulacrum for fleshing this out and sorry for not finishing it myself :sweat_smile: ended up sick before the holidays and forgot about it xx

some small (mostly vibe based) nits, but the summary is already very good

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,179 @@
...
+to be written in some cases.
+
+For more details, see a [previous blog post](https://blog.rust-lang.org/inside-rust/2024/12/04/trait-system-refactor-initiative.html)
+and the [stabilization report](https://github.com/rust-lang/rust/pull/130654).
+
+[Crater]: https://github.com/rust-lang/crater/
+
+### Strict provenance APIs
+
+In Rust, [pointers are not simply an "integer" or
```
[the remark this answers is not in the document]
Mark-Simulacrum: My preference is to leave this just talking about Rust, I think. IIRC, C has said both more and less on this depending on the specific place you read, and I'd rather avoid people reading too much into the paragraph here.

(In particular, we have not yet settled on any concrete rules to my knowledge about provenance, though they're likely to draw from a similar space as c does).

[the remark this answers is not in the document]
Urgau: Fair enough.

pietroalbini (approved):

epage (approved):

traviscross (approved):

posts/2025-01-09-Rust-1.84.0.md
```diff
@@ -0,0 +1,179 @@
...
+- [`<ptr>::with_addr`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.with_addr)
+- [`<ptr>::map_addr`](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.map_addr)
+- [`<int>::isqrt`](https://doc.rust-lang.org/stable/core/primitive.i32.html#method.isqrt)
+- [`<int>::checked_isqrt`](https://doc.rust-lang.org/stable/core/primitive.i32.html#method.checked_isqrt)
+- [`<uint>::isqrt`](https://doc.rust-lang.org/stable/core/primitive.u32.html#method.isqrt)
+- [`NonZero::isqrt`](https://doc.rust-lang.org/stable/core/num/struct.NonZero.html#impl-NonZero%3Cu128%3E/method.isqrt)
+- [`core::ptr::without_provenance`](https://doc.rust-lang.org/stable/core/ptr/fn.without_provenance.html)
+- [`core::ptr::without_provenance_mut`](https://doc.rust-lang.org/stable/core/ptr/fn.without_provenance_mut.html)
+- [`core::ptr::dangling`](https://doc.rust-lang.org/stable/core/ptr/fn.dangling.html)
+- [`core::ptr::dangling_mut`](https://doc.rust-lang.org/stable/core/ptr/fn.dangling_mut.html)
```
traviscross, line 145:
```suggestion
- [`core::ptr::dangling_mut`](https://doc.rust-lang.org/stable/core/ptr/fn.dangling_mut.html)
- [`Pin::as_deref_mut`](https://doc.rust-lang.org/stable/core/pin/struct.Pin.html#method.as_deref_mut)
```

See:

- https://github.com/rust-lang/rust/pull/135275

PR #1462, opened 2025-01-23
Update serde_yaml to 0.9.34-deprecated

senekor:
The renovate bot is having trouble with the `-deprecated` marker:
https://github.com/rust-lang/blog.rust-lang.org/pull/1286

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1855,9 +1855,9 @@ dependencies = [
 
 [[package]]
 name = "serde_yaml"
-version = "0.9.33"
+version = "0.9.34-deprecated"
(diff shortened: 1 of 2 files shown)


Turbo87:
> The renovate bot is having trouble with the `-deprecated` marker

the "problem" is that `-deprecated` is treated as a pre-release indicator and renovate by default is not updating from a regular release to a pre-release. we could disable https://docs.renovatebot.com/configuration-options/#ignoreunstable for `serde-yaml`, but since this version is most likely the last anyway it's probably not worth it :D

Turbo87 (approved):

PR #1473, opened 2025-02-05
Revert "style: wrap text code to prevent content overflow-x"

peterpeterparker:
# Motivation

Revert PR #1325.

PR was 7 months old, and other CSS changes happened in the meantime, which made its style break some styles for the `<code>` block once applied today, as reported by @weihanglo.

Before #1325:

image

After:

image

# Notes

I had a quick look, and it seems that the original issue addressed by #1325 in May 2024 still occurs on main when a long function is inside a code block within a list. This can now potentially be prevented with #1474.

PR #1474, opened 2025-02-05
style: word wrap code to prevent content overflow-x

peterpeterparker:
# Motivation

While I reverted #1473—a change originally intended to prevent overflow-x on mobile—I noticed that, despite CSS evolving, the issue still occasionally occurs when a code block inside a link within a list has text long enough to overflow the screen. Wrapping the text forces it to break onto a new line.

Before:

Capture d’écran 2025-02-05 à 06 31 38

After:

after

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -225,6 +225,14 @@ div.brand {
     white-space: pre-wrap;
   }
 
+  pre code {
+    white-space: inherit;
+  }
+
+  ul code {
+    display: block;
+  }
+
   a.anchor::before {
     content: "§";
     display: none;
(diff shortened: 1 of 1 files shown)


src/styles/app.scss
```diff
@@ -120,6 +120,10 @@ code {
   overflow: auto;
 }
 
+ul a code {
```
Turbo87, line 123: this seems like quite a specific selector 😅 

while this would probably fix the issue for the specific blog post you took a screenshot of, I assume it won't work for `ol` lists, for `code` blocks that are not links, and for `code` blocks that are not in lists.

do you think we can find a solution that would work for all of these cases? :)

peterpeterparker: Fair point, I agree with you. That was a quick fix. I'll make a note and tackle this properly when I have more time. Meanwhile, I'll close the PR.

PR #1476, opened 2025-02-05
Address overlooked feedback

oli-obk:
The sentence carries no meaningful information for readers and the tone is not great


traviscross (approved):
Approving as requested by the moderation team.

PR #1505, opened 2025-03-08
Fix RSS feed

senekor:
This was broken in the transition from Hanlebars to Tera. It was excluded from the snapshot tests, because it contains a non-deterministic timestamp.

reported here: https://github.com/rust-lang/blog.rust-lang.org/pull/1502#issuecomment-2707907095

### DIFF
--- a/templates/feed.tera
+++ b/templates/feed.tera
@@ -14,16 +14,16 @@
 
     {% for post in posts %}
     <entry>
-        <title>{{post.title}}</title>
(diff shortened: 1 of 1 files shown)


senekor:
cargo-insta has a feature called [Redactions](https://insta.rs/docs/redactions/) to ignore non-deterministic parts on a fine-grained level, I should've used that instead of excluding the whole file.

Manishearth (approved):

PR #1506, opened 2025-03-08
Add snapshot testing setup

senekor:


senekor:
I added specific tests for non-deterministic files with the timestamps filtered out, to prevent mistakes like #1505 in the future.

Manishearth (approved):

PR #1508, opened 2025-03-08
Replace yaml with toml front matter

senekor:
I snapshot tested this locally, the output should be identical. One commit contains the changes to the header of every post, which is a big, scripted diff. I also snuck in an unrelated change by fixing the windows line endings on a few posts, which looks like the entire file was changed. The other commits have a more readable diff.

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/senekor/toml-front-matter/posts/2014-09-15-Rust-1.0.md)


senekor:
A tiny regression seems to be that the preview rendered by triagebot doesn't look the same. Yaml front matter results in a nice table while toml front matter is left as-is.

Manishearth (approved):
Thought: What if we had a CI job that ran snapshot tests against the merge-base or trunk IF the PR body had "RUN_SNAPSHOT_TESTS" in it?

senekor:
Great idea! Opened an [issue](https://github.com/rust-lang/blog.rust-lang.org/issues/1511) for it so I don't forget :)

PR #1519, opened 2025-03-10
Fix file name conflict on case-insensitive FS

### LINKED ISSUE #1516: Repo broken for case-insensitive filesystems
#1510 seems to be causing a problem for case-insensitive filesystems. There are files that now differ only by case, which causes git to essentially be broken. They are:

posts/inside-rust/Goverance-wg-cfp.md
posts/inside-rust/Governance-wg.md
posts/inside-rust/Lang-Team-Meeting.md

senekor:
fixes #1516 

snapshot tested locally

cc @ehuss 

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/senekor/wxrkuuzuvwwk/posts/inside-rust/Goverance-wg-cfp@0.md)


ehuss (approved):
Thanks!

PR #1520, opened 2025-03-10
Improve error message for bad front matter

### LINKED ISSUE #1517: Confusing errors for TOML front-matter issues
I'm running into some problems with the TOML front-matter checks. They look roughly like this:

```
Error:
   0: markdown file must start with the line `+++`

Location:
   front_matter/src/lib.rs:24
```

The problem is that it doesn't tell me which file is having the problem, which makes me need to hunt down where it was having an issue (it was not with the file I was currently working on). The same goes for other kinds of errors like:

```
Error:
   0: TOML parse error at line 1, column 1
   0:   |
   0: 1 | layout = "post"
   0:   | ^^^^^^^^^^^^^^^
   0: missing field `date`
   0:

Location:
   front_matter/src/lib.rs:31
```

senekor:
fixes #1517 

cc @ehuss

### DIFF
--- a/front_matter/src/lib.rs
+++ b/front_matter/src/lib.rs
@@ -1,4 +1,4 @@
-use eyre::bail;
+use eyre::{ContextCompat, bail};
 use serde::{Deserialize, Serialize};
 use toml::value::Date;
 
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

PR #1534, opened 2025-03-25
Announcing Rust 1.86.0

BoxyUwU:
cc @rust-lang/release

### DIFF
--- a/posts/Rust-1.86.0.md
+++ b/posts/Rust-1.86.0.md
@@ -1,0 +1,174 @@
++++
+layout = "post"
+date = 2025-04-03
+title = "Announcing Rust 1.86.0"
+author = "The Rust Release Team"
+release = true
++++
+
+The Rust team is happy to announce a new version of Rust, 1.86.0. Rust is a programming language empowering everyone to build reliable and efficient software.
+
+If you have a previous version of Rust installed via `rustup`, you can get 1.86.0 with:
+
+```console
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.86.0](https://doc.rust-lang.org/stable/releases.html#version-1860-2025-04-03).
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.86.0 stable
+
+### Trait upcasting
+
+This release includes a long awaited feature — the ability to upcast trait objects.
+If a trait has a [supertrait](https://doc.rust-lang.org/reference/items/traits.html#supertraits) you can coerce a reference to said trait object to a reference to a trait object of the supertrait:
+
+```rust
+trait Trait: Supertrait {}
+trait Supertrait {}
+
+fn upcast(x: &dyn Trait) -> &dyn Supertrait {
+    x
+}
+```
+
+The same would work with any other kind of (smart-)pointer, like `Arc<dyn Trait> -> Arc<dyn Supertrait>` or `*const dyn Trait -> *const dyn Supertrait`.
+
+Previously this would have required a workaround in the form of an `upcast` method in the `Trait` itself, for example `fn as_supertrait(&self) -> &dyn Supertrait`, and this would work only for one kind of reference/pointer. Such workarounds are not necessary anymore. 
+
+Note that this means that raw pointers to trait objects carry a non-trivial invariant: "leaking" a raw pointer to a trait object with an invalid vtable into safe code may lead to undefined behavior. It is not decided yet whether creating such a raw pointer temporarily in well-controlled circumstances causes immediate undefined behavior, so code should refrain from creating such pointers under any conditions (and Miri enforces that).
+
+Trait upcasting may be especially useful with the `Any` trait, as it allows upcasting your trait object to `dyn Any` to call `Any`'s downcast methods, without adding any trait methods or using external crates.
+
+```rust
+use std::any::Any;
+
+trait MyAny: Any {}
+
+impl dyn MyAny {
+    fn downcast_ref<T>(&self) -> Option<&T> {
+        (self as &dyn Any).downcast_ref()
+    }
+}
+```
+
+You can [learn more about trait upcasting in the Rust reference](https://doc.rust-lang.org/reference/type-coercions.html#unsized-coercions).
+
+### `HashMap`s and slices now support indexing multiple elements mutably
+
+The borrow checker prevents simultaneous usage of references obtained from repeated calls to `get_mut` methods. To safely support this pattern the standard library now provides a `get_disjoint_mut` helper on slices and `HashMap` to retrieve mutable references to multiple elements simultaneously. See the following example taken from the API docs of [`slice::get_disjoint_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_disjoint_mut):
+```rust
+let v = &mut [1, 2, 3];
+if let Ok([a, b]) = v.get_disjoint_mut([0, 2]) {
+    *a = 413;
+    *b = 612;
+}
+assert_eq!(v, &[413, 2, 612]);
+
+if let Ok([a, b]) = v.get_disjoint_mut([0..1, 1..3]) {
+    a[0] = 8;
+    b[0] = 88;
+    b[1] = 888;
+}
+assert_eq!(v, &[8, 88, 888]);
+
+if let Ok([a, b]) = v.get_disjoint_mut([1..=2, 0..=0]) {
+    a[0] = 11;
+    a[1] = 111;
+    b[0] = 1;
+}
+assert_eq!(v, &[1, 11, 111]);
+```
+
+### Allow safe functions to be marked with the `#[target_feature]` attribute.
+
+Previously only `unsafe` functions could be marked with the `#[target_feature]` attribute as it is unsound to call such functions without the target feature being enabled. This release stabilizes the `target_feature_11` feature, allowing *safe* functions to be marked with the `#[target_feature]` attribute.
+
+Safe functions marked with the target feature attribute can only be safely called from other functions marked with the target feature attribute. However, they cannot be passed to functions accepting generics bounded by the `Fn*` traits and only support being coerced to function pointers inside of functions marked with the `target_feature` attribute.
+
+Inside of functions not marked with the target feature attribute they can be called inside of an `unsafe` block, however it is the callers responsibility to ensure that the target feature is available.
+
+```rust
+#[target_feature(enable = "avx2")]
+fn requires_avx2() {
+    // ... snip
+}
+
+#[target_feature(enable = "avx2")]
+fn safe_callsite() {
+    // Calling `requires_avx2` here is safe as `bar`
+    // requires the `avx2` feature itself.
+    requires_avx2();
+}
+
+fn unsafe_callsite() {
+    // Calling `requires_avx2` here is unsafe, as we must
+    // ensure that the `avx2` feature is available first.
+    if is_x86_feature_detected!("avx2") {
+        unsafe { requires_avx2() };
+    }
+}
+```
+
+You can check the [`target_features_11`](https://github.com/rust-lang/rfcs/blob/master/text/2396-target-feature-1.1.md) RFC for more information.
+
+### Debug assertions that pointers are non-null when required for soundness
+
+The compiler will now insert debug assertions that a pointer is not null upon non-zero-sized reads and writes, and also when the pointer is reborrowed into a reference. For example, the following code will now produce a non-unwinding panic when debug assertions are enabled:
+```rust
+let _x = *std::ptr::null::<u8>();
+let _x = &*std::ptr::null::<u8>();
+```
+Trivial examples like this have produced a warning since Rust 1.53.0, the new runtime check will detect these scenarios regardless of complexity.
+
+These assertions only take place when debug assertions are enabled which means that they **must not** be relied upon for soundness. This also means that dependencies which have been compiled with debug assertions disabled (e.g. the standard library) will not trigger the assertions even when called by code with debug assertions enabled.
+
+### Make `missing_abi` lint warn by default
+
+Omitting the ABI in extern blocks and functions (e.g. `extern {}` and `extern fn`) will now result in a warning (via the `missing_abi` lint). Omitting the ABI after the `extern` keyword has always implicitly resulted in the `"C"` ABI. It is now recommended to explicitly specify the `"C"` ABI (e.g. `extern "C" {}` and `extern "C" fn`).
+
+You can check the [Explicit Extern ABIs RFC](https://rust-lang.github.io/rfcs/3722-explicit-extern-abis.html) for more information.
+
+### Target deprecation warning for 1.87.0
+
+The tier-2 target `i586-pc-windows-msvc` will be removed in the next version of Rust, 1.87.0. Its difference to the much more popular `i686-pc-windows-msvc` is that it does not require SSE2 instruction support, but Windows 10, the minimum required OS version of all `windows` targets (except the `win7` targets), requires SSE2 instructions itself.
+
+All users currently targeting `i586-pc-windows-msvc` should migrate to `i686-pc-windows-msvc` before the `1.87.0` release.
+
+You can check the [Major Change Proposal](https://github.com/rust-lang/compiler-team/issues/840) for more information.
+
+### Stabilized APIs
+
+- [`{float}::next_down`](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.next_down)
+- [`{float}::next_up`](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.next_up)
+- [`<[_]>::get_disjoint_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_disjoint_mut)
+- [`<[_]>::get_disjoint_unchecked_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_disjoint_unchecked_mut)
+- [`slice::GetDisjointMutError`](https://doc.rust-lang.org/stable/std/slice/enum.GetDisjointMutError.html)
+- [`HashMap::get_disjoint_mut`](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.get_disjoint_mut)
+- [`HashMap::get_disjoint_unchecked_mut`](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.get_disjoint_unchecked_mut)
+- [`NonZero::count_ones`](https://doc.rust-lang.org/stable/std/num/struct.NonZero.html#method.count_ones)
+- [`Vec::pop_if`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.pop_if)
+- [`sync::Once::wait`](https://doc.rust-lang.org/stable/std/sync/struct.Once.html#method.wait)
+- [`sync::Once::wait_force`](https://doc.rust-lang.org/stable/std/sync/struct.Once.html#method.wait_force)
+- [`sync::OnceLock::wait`](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html#method.wait)
+
+These APIs are now stable in const contexts:
+
+- [`hint::black_box`](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html)
+- [`io::Cursor::get_mut`](https://doc.rust-lang.org/stable/std/io/struct.Cursor.html#method.get_mut)
+- [`io::Cursor::set_position`](https://doc.rust-lang.org/stable/std/io/struct.Cursor.html#method.set_position)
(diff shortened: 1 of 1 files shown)


posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+        (self as &dyn Any).downcast_ref()
+    }
+}
+```
+
+You can [learn more about trait upcasting in the Rust reference](https://doc.rust-lang.org/reference/type-coercions.html#unsized-coercions).
+
+### `HashMap`s and slices now support indexing multiple elements mutably
+
+The borrow checker will prevent simaltaneous usage of references returned from `get_mut` methods. To get around this the standard library now provides a `get_disjoint_mut` helper on slices and `HashMap` to retrieve mutable references to multiple elements simaltaneously. See the following example taken from the API docs of [`slice::get_disjoint_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_disjoint_mut):
```
programmerjake, line 61:
```suggestion
The borrow checker will prevent simultaneous usage of references returned from `get_mut` methods. To get around this the standard library now provides a `get_disjoint_mut` helper on slices and `HashMap` to retrieve mutable references to multiple elements simultaneously. See the following example taken from the API docs of [`slice::get_disjoint_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_disjoint_mut):
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+#[target_feature(enable = "avx2")]
+fn bar() {
+    // Calling `avx2` here is safe.
+    avx2();
+}
+```
+
+### Debug assert pointers are non-null when required for soundness
+
+The compiler will now insert checks that a pointer is not null upon non-zero-sized reads and writes, and also when the pointer is reborrowed into a reference. For example, the following code will now produce a non-unwinding panic:
```
programmerjake, line 113: should probably mention in the paragraph that the asserts are only inserted when debug assertions are enabled, which they aren't by default in release mode. maybe also that dereferencing null pointers is still usually UB.

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+```rust
+trait Trait: Supertrait {}
+trait Supertrait {}
+
+fn upcast(x: &dyn Trait) -> &dyn Supertrait {
+    x
+}
+```
+
+The same would work with any other kind of (smart)-pointer, like `Arc<dyn Trait> -> Arc<dyn Supertrait>` or `*const dyn Trait -> *const dyn Supertrait`.
```
traviscross, line 37:
```suggestion
The same would work with any other kind of (smart-)pointer, like `Arc<dyn Trait> -> Arc<dyn Supertrait>` or `*const dyn Trait -> *const dyn Supertrait`.
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+
+Previously this would require a workaround in the form of an `upcast` method in the `Trait` itself (and that would work only for one kind of reference/pointer). Such workarounds are not necessary anymore.
```
traviscross, line 39:
```suggestion
Previously this would have required a workaround in the form of an `upcast` method in the `Trait` itself (and that would work only for one kind of reference/pointer). Such workarounds are not necessary anymore. 
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+fn upcast(x: &dyn Trait) -> &dyn Supertrait {
+    x
+}
+```
+
+The same would work with any other kind of (smart)-pointer, like `Arc<dyn Trait> -> Arc<dyn Supertrait>` or `*const dyn Trait -> *const dyn Supertrait`.
+
+Previously this would require a workaround in the form of an `upcast` method in the `Trait` itself (and that would work only for one kind of reference/pointer). Such workarounds are not necessary anymore. 
+
+Note that this adds a new _safety invariant_  to raw pointers — if a pointer points to a trait object, vtable must be valid for that trait. This means that "leaking" a raw pointer to a trait object with an invalid vtable into safe code may lead to undefined behavior.
```
CleanCut, line 41:
```suggestion
Note that this adds a new _safety invariant_  to raw pointers — if a pointer points to a trait object, the vtable must be valid for that trait. This means that "leaking" a raw pointer to a trait object with an invalid vtable into safe code may lead to undefined behavior.
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+```
+Trivial examples like this have produced a warning since Rust 1.53.0, the new runtime check will detect these scenarios regardless of complexity.
+
+### Make `missing_abi` lint warn by default
+
+Omitting the ABI in extern blocks and functions (e.g. `extern {}` and `extern fn`) will now result in a warning (via the `missing_abi` lint). Omitting the ABI after the `extern` keyword has always implicitly resulted in the `"C"` ABI. It is now recommended to explicitly specify the `"C"` ABI (e.g. `extern "C" {}` and `extern "C" fn`).
+
+### Target deprecation warning for 1.87.0
+
+The target `i586-pc-windows-msvc` will be removed in the next version of Rust, 1.87.0. This is because this target is effectively useless. Its difference to the much more popular `i686-pc-windows-msvc` is that it does not require SSE2 instruction support, but Windows 10, the minimum required OS version of all `windows` targets (except the `win7` targets), requires SSE2 instructions itself, so this target does not fulfill a purpose.
```
CleanCut, line 126: Would it make sense to make an explicit recommendation to migrate?

```suggestion
The target `i586-pc-windows-msvc` will be removed in the next version of Rust, 1.87.0. This is because this target is effectively useless. Its difference to the much more popular `i686-pc-windows-msvc` is that it does not require SSE2 instruction support, but Windows 10, the minimum required OS version of all `windows` targets (except the `win7` targets), requires SSE2 instructions itself, so this target does not fulfill a purpose.

Users currently targeting `i586-pc-windows-msvc` should migrate to `i686-pc-windows-msvc` before the 1.87.0 release.
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+### Debug assert pointers are non-null when required for soundness
+
+The compiler will now insert checks that a pointer is not null upon non-zero-sized reads and writes, and also when the pointer is reborrowed into a reference. For example, the following code will now produce a non-unwinding panic:
+```rust
+let _x = *std::ptr::null::<u8>();
+let _x = &*std::ptr::null::<u8>();
+```
+Trivial examples like this have produced a warning since Rust 1.53.0, the new runtime check will detect these scenarios regardless of complexity.
```
CleanCut, line 118:
```suggestion
Trivial examples like this have produced a warning since Rust 1.53.0. The new runtime check will detect these scenarios regardless of complexity.
```

CleanCut (commented):
Some minor suggestions. Feel free to take them or leave them. 😄

posts/Rust-1.86.0.md
traviscross, line 41:
```suggestion
Note that this adds a new _safety invariant_ to raw pointers — if a pointer points to a trait object, the vtable must be valid for that trait. This means that "leaking" a raw pointer to a trait object with an invalid vtable into safe code may lead to undefined behavior.
```

This refers to: https://github.com/rust-lang/rust/issues/101336

Perhaps we may want to reword this a bit to suggest that we're now leaning on this invariant rather than just now adding it, but perhaps it's the same difference.

cc @RalfJung @nikomatsakis

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.86.0](https://doc.rust-lang.org/stable/releases.html#version-1860-2025-04-03).
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.86.0 stable
+
+### Trait upcasting
+
+This release includes a long awaited feature — the ability to upcast trait objects.
+If a trait has a [supertrait](https://doc.rust-lang.org/reference/items/traits.html#supertraits) you can coerce a reference to said trait object to a reference to a trait object of the super trait:
```
kpreid, line 26:
```suggestion
If a trait has a [supertrait](https://doc.rust-lang.org/reference/items/traits.html#supertraits) you can coerce a reference to said trait object to a reference to a trait object of the supertrait:
```

posts/Rust-1.86.0.md
kpreid, line 39:
```suggestion
Previously this would require a workaround in the form of an upcast method in the `Trait` itself (and that would work only for one kind of reference/pointer), like `fn as_supertrait(&self) -> &dyn Supertrait`. Such workarounds are not necessary anymore. 
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+The same would work with any other kind of (smart)-pointer, like `Arc<dyn Trait> -> Arc<dyn Supertrait>` or `*const dyn Trait -> *const dyn Supertrait`.
+
+Previously this would require a workaround in the form of an `upcast` method in the `Trait` itself (and that would work only for one kind of reference/pointer). Such workarounds are not necessary anymore. 
+
+Note that this adds a new _safety invariant_  to raw pointers — if a pointer points to a trait object, vtable must be valid for that trait. This means that "leaking" a raw pointer to a trait object with an invalid vtable into safe code may lead to undefined behavior.
+
+Trait upcasting may be especially useful with the `Any` trait, as it allows upcasting your trait object to `dyn Any` to call the downcast methods. Whereas before you'd have to write workarounds for this or use external crates.
```
kpreid, line 43:
```suggestion
Trait upcasting may be especially useful with the `Any` trait, as it allows upcasting your trait object to `dyn Any` to call `Any`’s downcast functions, without adding any trait methods or using external crates.
```

posts/Rust-1.86.0.md
kpreid, line 61: “The borrow checker will prevent…” could be read as saying that in this and future versions, it will *start to* prevent something. Also, “get around this” sounds a bit sketchy. How about:

```suggestion
The borrow checker prevents simultaneous usage of references obtained from more than one call to `get_mut` methods. To allow simultaneous mutable access to multiple `HashMap` or slice entries, the standard library now provides `get_disjoint_mut` methods to obtain multiple references in a single operation. See the following example taken from the API docs of [`slice::get_disjoint_mut`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_disjoint_mut):
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+
+if let Ok([a, b]) = v.get_disjoint_mut([1..=2, 0..=0]) {
+    a[0] = 11;
+    a[1] = 111;
+    b[0] = 1;
+}
+assert_eq!(v, &[1, 11, 111]);
+```
+
+### Allow safe functions to be marked with the #[target_feature] attribute.
```
kpreid, line 85:
```suggestion
### Allow safe functions to be marked with the `#[target_feature]` attribute
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+
+Functions marked with `#[target_feature]` are generally considered as unsafe functions: they are unsafe to call, cannot *generally* be assigned to safe function pointers, and don't implement the `Fn*` traits.
+
+However, calling them from other `#[target_feature]` functions with a superset of features is safe.
```
kpreid, line 89: It feels unclear what part of this section is the new feature and what part is background information.

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+}
+
+#[target_feature(enable = "avx2")]
+fn bar() {
+    // Calling `avx2` here is safe.
+    avx2();
+}
+```
+
+### Debug assert pointers are non-null when required for soundness
```
kpreid, line 111: On first reading I thought this was a sentence about "debug assert pointers", whatever those are.

```suggestion
### Debug assertions that pointers are non-null when required for soundness
```

posts/Rust-1.86.0.md
RalfJung: The term "safety invariant" has not been used in official docs so far, and is under discussion at https://github.com/rust-lang/unsafe-code-guidelines/issues/539.

Regarding your comment Travis, I think ideally we'd word this in a way that does not make this sound like a breaking change. ;)  Maybe something like:

> Note that this means that raw pointers to trait objects carry a non-trivial invariant: "leaking" a raw pointer to a trait object with an invalid vtable into safe code may lead to undefined behavior. It is not decided yet whether creating such a raw pointer temporarily in well-controlled circumstances causes immediate undefined behavior, so code should refrain from creating such pointers under any conditions (and Miri enforces that).

posts/Rust-1.86.0.md
jieyouxu: It may further be useful to explicitly stress that the user *cannot* rely on this behavior to catch/prevent UB, the same level of care is necessary as before.

posts/Rust-1.86.0.md
jieyouxu: This migration is not always possible, maybe qualify it with "if SSE2 instructions are supported" then `i686-pc-windows-msvc` is a viable migration target. Otherwise, the user will likely have to build their own `i586-pc-windows-msvc`-esque toolchain.

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,163 @@
...
+The compiler will now insert checks that a pointer is not null upon non-zero-sized reads and writes, and also when the pointer is reborrowed into a reference. For example, the following code will now produce a non-unwinding panic:
+```rust
+let _x = *std::ptr::null::<u8>();
+let _x = &*std::ptr::null::<u8>();
+```
+Trivial examples like this have produced a warning since Rust 1.53.0, the new runtime check will detect these scenarios regardless of complexity.
+
+### Make `missing_abi` lint warn by default
+
+Omitting the ABI in extern blocks and functions (e.g. `extern {}` and `extern fn`) will now result in a warning (via the `missing_abi` lint). Omitting the ABI after the `extern` keyword has always implicitly resulted in the `"C"` ABI. It is now recommended to explicitly specify the `"C"` ABI (e.g. `extern "C" {}` and `extern "C" fn`).
```
jieyouxu, line 122: Suggestion: Worth elaborating on the *why*. Or at least, backlink to [RFC 3722 Explicit Extern ABIs](https://rust-lang.github.io/rfcs/3722-explicit-extern-abis.html).

> Originally, "C" was a very reasable default for extern. However, with work ongoing to add other ABIs to Rust, it is no longer obvious that "C" should forever stay the default.
>
> By making the ABI explicit, it becomes much clearer that "C" is just one of the possible choices, rather than the “standard” way for external functions. Removing the default makes it easier to add a new ABI on equal footing as "C".

Another argument in favor is making the extern ABI *explicit* minimizes room for misunderstanding/confusion by explicitly expressing user intention, especially in potentially-FFI code where there may be a lot of contracts/invariants that need to be upheld.

jieyouxu (approved):
Looks good for the most part :D

posts/Rust-1.86.0.md
cuviper: Alternatively, hyphenating "Debug-assert" would also clarify the compound meaning.

posts/Rust-1.86.0.md
BoxyUwU: Added this, I feel like it reads a bit weird to qualify the migration and then not give an alternative but it also feels a bit weird to suggest people maintain their own toolchain lol

posts/Rust-1.86.0.md
BoxyUwU: Yeah it read kind of awkward to me too. Admittedly I don't really have much of an understanding of how this feature worked so just copied it from the tracking issue and didn't edit it :thinking:

posts/Rust-1.86.0.md
RalfJung: The existing target already requires Windows 10 which requires SSE2. So migration should be always possible. IOW, the target made no logical sense. ;)

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,169 @@
...
+#[target_feature(enable = "avx2")]
+fn bar() {
+    // Calling `avx2` here is safe.
+    avx2();
+}
+```
+
+### Debug assertions that pointers are non-null when required for soundness
+
+The compiler will now insert checks that a pointer is not null upon non-zero-sized reads and writes, and also when the pointer is reborrowed into a reference. For example, the following code will now produce a non-unwinding panic:
```
joshtriplett, line 113:
```suggestion
In debug mode, the compiler will now insert checks that a pointer is not null upon non-zero-sized reads and writes, and also when the pointer is reborrowed into a reference. For example, the following code will now produce a non-unwinding panic:
```

(The section title does say this already, as does the *last* paragraph, but it seems worth reiterating here for clarity.)

posts/Rust-1.86.0.md
BoxyUwU: That's what I had thought but the "except the win7 targets" makes it seem like that's an exception. But also we say here that windows 10 is the minimum required version for windows targets.... So I'm not really sure how we simultaneously can have win7 targets that need to be called out as an exception, but also we require atleast windows 10 :thinking:

RalfJung: "Windows 10, the minimum required OS version of all `windows` targets (except the `win7` targets), requires SSE2 instructions itself" is a universal statement and correct. However it is confusing here because we only care about one target, `i586-pc-windows-msvc`, which is *not* a `win7` target.

BoxyUwU: Oh I see, the win7 targets are named as `*-win7-windows.msvc`. I guess I got a bit mixed up between terms of art and the colloquial meaning of a "windows target" ^^

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,169 @@
...
+
+Omitting the ABI in extern blocks and functions (e.g. `extern {}` and `extern fn`) will now result in a warning (via the `missing_abi` lint). Omitting the ABI after the `extern` keyword has always implicitly resulted in the `"C"` ABI. It is now recommended to explicitly specify the `"C"` ABI (e.g. `extern "C" {}` and `extern "C" fn`).
+
+You can check the [Explicit Extern ABIs RFC](https://rust-lang.github.io/rfcs/3722-explicit-extern-abis.html) for more information.
+
+### Target deprecation warning for 1.87.0
+
+The target `i586-pc-windows-msvc` will be removed in the next version of Rust, 1.87.0. Its difference to the much more popular `i686-pc-windows-msvc` is that it does not require SSE2 instruction support, but Windows 10, the minimum required OS version of all `windows` targets (except the `win7` targets), requires SSE2 instructions itself.
+
+If SSE2 instructions are supported then users currently targeting `i586-pc-windows-msvc` should migrate to `i686-pc-windows-gnu` before the `1.87.0` release.
```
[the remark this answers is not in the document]
BoxyUwU: ah :woman_facepalming: typo

alex-semenyuk (approved):

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Make `missing_abi` lint warn by default
+
+Omitting the ABI in extern blocks and functions (e.g. `extern {}` and `extern fn`) will now result in a warning (via the `missing_abi` lint). Omitting the ABI after the `extern` keyword has always implicitly resulted in the `"C"` ABI. It is now recommended to explicitly specify the `"C"` ABI (e.g. `extern "C" {}` and `extern "C" fn`).
+
+You can check the [Explicit Extern ABIs RFC](https://rust-lang.github.io/rfcs/3722-explicit-extern-abis.html) for more information.
+
+### Target deprecation warning for 1.87.0
+
+The target `i586-pc-windows-msvc` will be removed in the next version of Rust, 1.87.0. Its difference to the much more popular `i686-pc-windows-msvc` is that it does not require SSE2 instruction support, but Windows 10, the minimum required OS version of all `windows` targets (except the `win7` targets), requires SSE2 instructions itself.
```
cuviper, line 136: Might be worth specifying tier-2:
```suggestion
The tier-2 target `i586-pc-windows-msvc` will be removed in the next version of Rust, 1.87.0. Its difference to the much more popular `i686-pc-windows-msvc` is that it does not require SSE2 instruction support, but Windows 10, the minimum required OS version of all `windows` targets (except the `win7` targets), requires SSE2 instructions itself.
```

posts/Rust-1.86.0.md
```diff
@@ -0,0 +1,175 @@
...
+
+All users currently targeting `i586-pc-windows-msvc` should migrate to `i686-pc-windows-msvc` before the `1.87.0` release.
+
+You can check the [Major Change Proposal](https://github.com/rust-lang/compiler-team/issues/840) for more information.
+
+### Stabilized APIs
```
cuviper, line 142: FYI, the original order of items seen here was whatever the relnotes tool generated, but I tweaked that for the final relnotes PR to make more sense (IMO). So you may want to repeat that change here.

cuviper (approved):
A couple nits, but no big deal either way. LGTM!

cuviper (approved):

traviscross (approved):

PR #1538, opened 2025-03-27
Project director update on Feb's board meeting

carols10cents:
[Rendered](https://github.com/integer32llc/blog.rust-lang.org/blob/project-director-update-2025-03/content/inside-rust/project-director-update@3.md)

### DIFF
--- a/content/inside-rust/project-director-update@3.md
+++ b/content/inside-rust/project-director-update@3.md
@@ -1,6 +1,6 @@
 +++
 layout = "post"
-date = 2025-03-27
+date = 2025-04-03
 title = "March 2025 Project Director Update"
 author = "Carol Nichols"
 team = "Rust Foundation Project Directors <https://foundation.rust-lang.org/about/>"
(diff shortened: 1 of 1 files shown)


senekor:
Note that posts were moved to the `content/` directory, sorry for the churn 🙂

senekor:
I don't review content, I'm just the janitor around here 😄 

I think usually some team lead does the content review. If there is no obviously responsible person, maybe the team `rust-lang/inside-rust-reviewers` would be appropriate to ping for review?

ehuss:
@carols10cents can you update the date to today, and I'll merge it?
(It looks like you've locked out edits so I can't change it.)

We really should figure out a better process here so that it isn't always on Mark to do everything. Do your fellow board directors want to review these posts at all? Can we set up some system so one of them can approve?

spastorino:
@ehuss exactly and totally agreed. My point was ... whoever sees these PRs and has rights to merge, should just do a minimal review and merge. If more people have rights to merge things better.
There's a separate discussion about who should have merge rights. I have a very relaxed opinion about this which is that probably every project member should be able to merge blog posts ;).

carols10cents:
@ehuss I've updated the date! Thank you!

ehuss (approved):

PR #1542, opened 2025-03-28
Replace tera filters with macros

senekor:
Zola doesn't support adding custom filters, but macros should work fine.

RUN_SNAPSHOT_TESTS


senekor:
@Manishearth #1543 was stacked on this, so the merge didn't end up on master. Maybe you missed this PR? Thanks.

Manishearth (approved):

PR #1560, opened 2025-04-06
Match Zolas URLs scheme

senekor:
This should not break permalinks at all and make it easier to compare the "before and after" of a Zola migration.


senekor:
I guess if we can use the `aliases` feature with a better redirect template ([see here](https://github.com/rust-lang/blog.rust-lang.org/issues/1522#issuecomment-2782500392)) this would not be a good idea after all.

PR #1565, opened 2025-04-13
Add This Month In Our Test Infra March 2025 issue

jieyouxu:
Date range: 2025-03-11 to 2025-04-13.

r? @onur-ozkan (or @Mark-Simulacrum, or compiler leads as fallback reviewers since this mostly has notable changes for `compiletest`)

[Rendered](https://github.com/jieyouxu/blog.rust-lang.org/blob/tmioti-mar-2025/content/inside-rust/test-infra-mar-2025.md)

### DIFF
--- a/content/inside-rust/test-infra-mar-2025.md
+++ b/content/inside-rust/test-infra-mar-2025.md
@@ -1,0 +1,266 @@
++++
+layout = "post"
+date = 2025-04-13
+title = "This Month in Our Test Infra: March 2025 issue"
+author = "Jieyou Xu"
+team = "the Bootstrap Team <https://www.rust-lang.org/governance/teams/infra#team-bootstrap>"
++++
+
+# This Month in Our Test Infra: March 2025 issue
+
+<!-- time period: 2025-03-11 through 2025-04-13 -->
+
+This is a quick summary of the changes in the test infrastructure for the [rust-lang/rust] repository[^scope]
+between **2025-03-11** and **2025-04-13**.
+
+During this period, there are some significant changes to [compiletest] and [bootstrap] that contributors should be aware of.
+
+[^scope]: The test infra here refers to the test harness [compiletest] and supporting components in our build system [bootstrap]. This test infra is used mainly by rustc and rustdoc. Other tools like cargo, miri or rustfmt maintain their own test infra.
+
+As usual, if you encounter bugs or UX issues when using our test infrastructure, please [file an issue][new-issue]. Bugs and papercuts can't be fixed if we don't know about them!
+
+**Thanks to everyone who contributed to our test infra!**
+
+## Highlights
+
+### `compiletest` now supports matching diagnostics on lines below
+
+UI error annotations previously could match against diagnostics emitted for previous lines (`//~^`) or same line (`//~`), but couldn't express matching against diagnostics emitted for subsequent lines. Following [#139100][pr-139100], test writers can now use `//~v` to match against diagnostics emitted for subsequent lines. This is particularly useful for e.g. parser tests where the test file may be exercising lexer/parser diagnostics and the test file itself is not syntactically valid (so can't use comments *afterwards* to match against *previous* lines).
+
+Example:
+
+```rs
+// tests/ui/parser/issues/issue-103451.rs
+
+struct R { }
+//~vv ERROR this file contains an unclosed delimiter
+struct S {
+    x: [u8; R
+```
+
+[pr-139100]: https://github.com/rust-lang/rust/pull/139100
+
+### `compiletest` now supports precise and exhaustive matching on diagnostics without line information associated with the main test file
+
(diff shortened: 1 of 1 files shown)


content/inside-rust/test-infra-mar-2025.md
```diff
@@ -0,0 +1,264 @@
...
+//~vv ERROR this file contains an unclosed delimiter
+struct S {
+    x: [u8; R
+```
+
+[pr-139100]: https://github.com/rust-lang/rust/pull/139100
+
+### `compiletest` now supports precise and exhaustive matching on diagnostics without line information associated with the main test file
+
+Previously, test writers had to sort to `//@ error-pattern` to match against diagnostics which do not have associated line information in the main test file. Now, test writers may use `//~? DIAGNOSTIC_KIND diagnostic_message` to match against a diagnostic with diagnostic kind `DIAGNOSTIC_KIND` (e.g. `ERROR` or `HELP`) with message `diagnostic_message`. See [compiletest: Support matching on diagnostics without a span #138865][pr-138865]. `//~?` may appear on any line of the test, but are conventionally placed at the end.
```
Mark-Simulacrum, line 45:
```suggestion
Previously, test writers had to resort to `//@ error-pattern` to match against diagnostics which do not have associated line information in the main test file. Now, test writers may use `//~? DIAGNOSTIC_KIND diagnostic_message` to match against a diagnostic with diagnostic kind `DIAGNOSTIC_KIND` (e.g. `ERROR` or `HELP`) with message `diagnostic_message`. See [compiletest: Support matching on diagnostics without a span #138865][pr-138865]. `//~?` may appear on any line of the test, but are conventionally placed at the end.
```

Mark-Simulacrum (approved):

content/inside-rust/test-infra-mar-2025.md
jieyouxu: Fixed

content/inside-rust/test-infra-mar-2025.md
```diff
@@ -0,0 +1,264 @@
...
+
+```toml
+[build]
+compiletest-use-stage0-libtest = true
+```
+
+Note that as a trade-off, we do check this force-stage0-libtest configuration in PR CI. This means that there are possible windows of time where libtest may be modified before bumping stage 0 compiler, which may require the PR author modifying libtest to add `cfg(bootstrap)`/`cfg(not(bootstrap))` to compiletest bits depending libtest programmatic API as suitable. This is done under the assumption that libtest programmatic API changes are comparatively much less frequent. If this proves to be a significant burden, the bootstrap team would be open to revisiting this scheme.
+
+Note in over the long term, we'd like to migrate away from depending on in-tree libtest (both to render this hack unnecessary and to have more control over test execution).
```
onur-ozkan, line 88: If we can complete https://github.com/rust-lang/rust/issues/59775, this could be solved easily. Maybe we should mention about it?

content/inside-rust/test-infra-mar-2025.md
```diff
@@ -0,0 +1,264 @@
...
+
+[pr-139386]: https://github.com/rust-lang/rust/pull/139386
+
+### `bootstrap.toml` is now also accepted as bootstrap configuration file under checkout root
+
+Previously, [bootstrap] only accepted `config.toml` as a configuration file under checkout root. This is potentially confusing since Cargo also uses `config.toml` as its configuration file. Now, [bootstrap will also accept `bootstrap.toml` as the configuration file][pr-137081]. The example config file `config.example.toml` has also been renamed to `bootstrap.example.toml`. `config.toml` will continue to work.
```
onur-ozkan, line 94: >  Now, [bootstrap will also accept `bootstrap.toml` as the configuration file][pr-137081]

This isn't fully correct. Bootstrap renamed `config.toml` to `bootstrap.toml`, but it will also support `config.toml` for a while to avoid BC on different checkouts. So `config.toml` won't work forever; it's just a temporary fallback.

content/inside-rust/test-infra-mar-2025.md
jieyouxu: The thing is, I don't think anyone is actively working on that, nor do I know of anyone from T-libs{-api} or T-testing-devex who wants to do that (T-testing-devex AFAIK is working on stabilizing libtest JSON output, but then eventually also working a replacement)

content/inside-rust/test-infra-mar-2025.md
jieyouxu: Right. Let me word this as "`config.toml` has been renamed `bootstrap.toml`" but `config.toml` is currently accepted as a temporary fallback to aid migration, but will eventually be rejected.

jieyouxu: Additionally, I can also move this bit out of *this* blog post, and go with the stage 0 redesign blog post. What do you think?

jieyouxu: In any case, fixed the wording to reflect that `config.toml` is a temporary churn-mitigation fallback, but will be phased out eventually.

jieyouxu:
Fixed `config.toml` wording to reflect it is a temporary fallback. [(diff)](https://github.com/rust-lang/blog.rust-lang.org/compare/4a82e1cd6519ae0f2d1a2c58d718357708a5549c..406b91f12f7b917a54fbd5b31340ef041755c430)

onur-ozkan (approved):

PR #1566, opened 2025-04-14
Suppress warnings for ```text code blocks

senekor:
as discussed here:
https://github.com/rust-lang/blog.rust-lang.org/pull/1563#discussion_r2042665106

RUN_SNAPSHOT_TESTS


senekor:
should be merged after #1567

Issue #1571, opened 2025-04-14
Unhighlighted portions of code in code blocks blend into the background


PatchMixolydic:
Rust blog:
![There is a code block for an installation command, but nothing appears inside of it.](https://github.com/user-attachments/assets/f5f0a290-a67f-4e1b-a268-ce772e9c6abe)

![Highlighting the code block reveals that the command is there, but invisible.](https://github.com/user-attachments/assets/568babfe-a84d-4715-a488-a1f37c860c96)

Inside Rust blog:
![A code example from Inside Rust. Only a few highlighted words are visible.](https://github.com/user-attachments/assets/d125b32b-6de8-4d6b-bcf3-d911c2d66537)

Tested on LibreWolf (Firefox) 136.0.4-1 on Arch Linux (6.13.8-arch1-1, 23 Mar 2025).

Possibly related to 14a515f.

senekor:
Thanks for the report. Will fix ASAP.

senekor:
Okay so the problem should be fixed for users who are using the system theme. Users who have a site-specific theme setting that's different from their system theme will currently have a mismatched code block theme, I'll fix that next.

PatchMixolydic:
Can confirm that this is (partially) fixed:
![The contents of code blocks are now visible.](https://github.com/user-attachments/assets/b23ad7c9-aa79-49fd-9704-9f5a87bf1fa0)

Thank you so much!

PR #1573, opened 2025-04-14
Make syntax hightlight theme respect theme setting

senekor:
It feels slightly wrong to hard-code these themes (twice!) but I think it should be fine in practice. Comments are added to all the related locations.

I didn't find a clean way to conditionally import another css file. Ther [@import](https://developer.mozilla.org/en-US/docs/Web/CSS/@import) rule must always be at the top of the file. It supports association with layers and media-queries. Media-queries are not what we want. I tried conditionally declaring the [@layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) in different orders, but that didn't work either.


senekor:
closes #1571 

There is no rush to review / merge this. Most users probably don't have a site-specific theme setting and aren't affected. Maybe a cleaner solution can be found.

Manishearth (approved):

PR #1582, opened 2025-04-17
Add an Inside Rust blog post for bootstrap sequence redesign

jieyouxu:
Accompanying blog post for bootstrap sequence redesign: https://github.com/rust-lang/rust/pull/119899

r? @onur-ozkan (and @Mark-Simulacrum)
cc @rust-lang/bootstrap (significant stage 0 bootstrap changes)

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/master/content/inside-rust/stage0-redesign/index.md)


jieyouxu:
Hi @senekor, do you know what this zola failure is? Or am I missing sth in the front-matter?

```
Run zola build
Building site...
Checking all internal links with anchors.
> Successfully checked 254 internal link(s) with anchors.
-> Creating 600 pages (0 orphan) and 2 sections
Error: Failed to build the site
Error: Failed to render 'feed.xml'
Error: Reason: Filter call 'nth' failed
Error: Reason: Filter `nth` received an incorrect type for arg `n`: got `-2` but expected a usize
Error: Process completed with exit code 1.
```

content/inside-rust/stage0-std-redesign.md
```diff
@@ -0,0 +1,53 @@
...
+
+We are reworking how the stage 0 bootstrap sequence works (the sequence used to build a stage 1 compiler).
+
+- Before: a stage 0 (possibly beta) compiler is used to build the in-tree std, which in turn is used to build the stage 1 compiler.
+- After: a stage 0 (possibly beta) compiler and a precompiled stage 0 std is instead used to build the stage 1 compiler.
+
+Notably, this means:
+
+- `./x {build,test,check} library --stage 0` becomes no-op, as stage 0 std is no longer the built in-tree std, and the minimum supported stage to build std is now `1`.
+    - Consequently, default (test, check, bench) stage values in the library profile are no longer `0`.
```
Kobzol, line 22: Maybe we could say what is the new stage? :D I assume it's 1?

content/inside-rust/stage0-std-redesign.md
```diff
@@ -0,0 +1,53 @@
...
+## Motivation
+
+The previous stage 0 bootstrapping sequence was a source of endless confusion for compiler, library and bootstrap contributors alike, because std had to support being built by *both* a previous possibly-beta rustc and in-tree rustc, with `cfg(bootstrap)` in std sources necessary to distinguish between them. By adjusting the stage 0 bootstrap sequence to instead use a precompiled stage 0 std instead of building the in-tree std, we hope to (1) simplify library development workflow to no longer need `cfg(bootstrap)` and (2) enable simplifying some bootstrap logic related to building in-tree std in stage 0.
+
+This was [originally proposed by @jyn514 in the MCP rust-lang/compiler-team#619](https://github.com/rust-lang/compiler-team/issues/619).
+
+## What does this mean for a typical library workflow?
+
+- Crucially, `./x {build,test,check} library --stage 0` becomes no-op and are no longer supported. Building the in-tree std now requires a stage 1 compiler.
+    - Consequently, library contributors are **strongly** encouraged to enable `rust.download-rustc = "if-unchanged"` to avoid having to build a stage 1 compiler.
```
Kobzol, line 34: This is the default for the `library` profile, we should mention that if they are using it already, they will be fine.

content/inside-rust/stage0-std-redesign.md
```diff
@@ -0,0 +1,53 @@
...
++++
+
+# Redesigning stage 0 std
+
+## Summary
+
+We are reworking how the stage 0 bootstrap sequence works (the sequence used to build a stage 1 compiler).
+
+- Before: a stage 0 (possibly beta) compiler is used to build the in-tree std, which in turn is used to build the stage 1 compiler.
+- After: a stage 0 (possibly beta) compiler and a precompiled stage 0 std is instead used to build the stage 1 compiler.
```
onur-ozkan, line 18:
```suggestion
- Before: a stage 0 (beta by default) compiler is used to build the in-tree std, which in turn is used to build the stage 1 compiler.
- After: a stage 0 (beta by default) compiler and a precompiled stage 0 std is instead used to build the stage 1 compiler.
```

content/inside-rust/stage0-std-redesign.md
jieyouxu: Yeah

content/inside-rust/stage0-std-redesign.md
jieyouxu: Good point, I mentioned it mostly because some contributors don't use a profile at all

senekor:
Oh if it's an inside-rust post, that should be in the path as well. So: "inside-rust/YYYY/MM/DD/whatever".

If the date of publication is not yet known, it's recommended to use the placeholder "9999/12/31" as the date, CI will prevent that from being published accidentally.

I can definitely improve docs/guidance/tooling/error msgs here.

content/inside-rust/stage0-std-redesign.md
```diff
@@ -0,0 +1,53 @@
...
+
+Notably, this means:
+
+- `./x {build,test,check} library --stage 0` becomes no-op, as stage 0 std is no longer the built in-tree std, and the minimum supported stage to build std is now `1`.
+    - Consequently, default (test, check, bench) stage values in the library profile are no longer `0`.
+- `cfg(bootstrap)` is no longer needed for library development. In (very) rare cases, `cfg(bootstrap)` may be needed in the compiler for dogfooding unstable library features.
```
onur-ozkan, line 24: > In (very) rare cases, `cfg(bootstrap)` may be needed in the compiler for dogfooding unstable library features.

Note that we already use cfg(bootstrap) on compiler tree for some cases, so cfg(bootstrap) usage on compiler tree won't be a new thing.

content/inside-rust/stage0-std-redesign.md
jieyouxu: Fixed in c8d819a5a40b5bc8e0e81314172eeff4471838ec

content/inside-rust/stage0-std-redesign.md
jieyouxu: Further clarified in c8d819a5a40b5bc8e0e81314172eeff4471838ec.

content/inside-rust/stage0-std-redesign.md
```diff
@@ -0,0 +1,54 @@
++++
+path = "inside-rust/9999/12/31/redesigning-stage-0-std"
```
jieyouxu, line 2: FIXME(jieyouxu): update this date when we're ready

jieyouxu:
> It'd be helpful to mention the https://github.com/rust-lang/rust/pull/119899.

I have no idea how I didn't include that when I even remember telling myself not to forget that lmao

content/inside-rust/stage0-std-redesign.md
jieyouxu: Fixed in 222f4feba50e4be53a56843b95563f8ef0ba031a

content/inside-rust/stage0-redesign/index.md
jieyouxu, line 1: TODO: include pre/post CLI invocations and compare their behavioral differences.

Notably,

- `./x build library --stage 0` (no-op in post) and `./x build library` (0 pre, 1 post)
- `./x build library --stage 0`

jieyouxu, line 1: TODO: having someone on t-libs and someone on t-compiler skim it and sign off that it makes sense

jieyouxu, line 1: TODO: move motivation section towards the end

jieyouxu, line 1: TODO: create two dedicated bootstrap support threads (or just one?) and link to them from this blog post

1. For general inqueries (why we're doing this?)
2. For helping with issues and bugs

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+## What does this mean for a typical library workflow?
+
+- Crucially, `./x {build,test,check} library --stage 0` becomes no-op and are no longer supported. Building the in-tree std now requires a stage 1 compiler.
+    - Consequently, library contributors are *strongly* encouraged to enable `rust.download-rustc = "if-unchanged"` to avoid having to build a stage 1 compiler. Note that this is the default for `profile = "library"`, but you may need to specify it manually if you don't use a `profile`.
+- `cfg(bootstrap)` should no longer be needed for library sources.
+
+### Caveat: `libtest` changes
+
+See the *Recommended config: have `compiletest` instead depend on precompiled stage 0 libtest* section below.
```
jieyouxu, line 42: FIXME(jieyouxu) [COMPILETEST_STAGE0 1/2]: drop this if we remove the compiletest force stage0 hack

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+## What does this mean for a typical compiler workflow?
+
+- If unstable library features are being dogfooded, some additional `cfg(bootstrap)` usages may now be needed in compiler sources.
+
+### Caveat: `libtest` changes
+
+See the *Recommended config: have `compiletest` instead depend on precompiled stage 0 libtest* section below.
```
jieyouxu, line 50: FIXME(jieyouxu) [COMPILETEST_STAGE0 2/3]: change this to `internal_output_capture` unstable library feature caveat.

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Recommended config: have `compiletest` instead depend on precompiled stage 0 libtest
+
+`compiletest` currently depends on in-tree libtest. For workflows that involve building the compiler, this can cause `compiletest` rebuilds if stage 1 library is rebuilt as a consequence of compiler changes. If you don't intend to change libtest, you can [specify `build.compiletest-use-stage0-libtest = true` to instead have `compiletest` depend on precompiled stage 0 libtest](https://github.com/rust-lang/rust/pull/139386). That way, compiler test iterations can avoid rebuilding `compiletest` unnecessarily. This is already the default unless the `dist` profile is used.
+
+Note that some CI jobs have `build.compiletest-use-stage0-libtest = true` set while others have `build.compiletest-use-stage0-libtest = false`, meaning that libtest programmatic API changes can require adding `cfg(bootstrap)`/`cfg(not(bootstrap))` to `compiletest`'s libtest API use sites. However, in practice we expect this to be very rare.
```
jieyouxu, line 56: FIXME(jieyouxu) [COMPILETEST_STAGE0 3/3]: drop this once we drop the hack

content/inside-rust/stage0-redesign/index.md
jieyouxu: Added comparisons

jieyouxu: Dedicated support thread: [#t-infra/bootstrap > Stage 0 std redesign support thread](https://rust-lang.zulipchat.com/#narrow/channel/326414-t-infra.2Fbootstrap/topic/Stage.200.20std.20redesign.20support.20thread/with/515096924), and backlinked

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+## What does this mean for a typical library workflow?
+
+- Crucially, `./x {build,test,check} library --stage 0` becomes no-op and are no longer supported. Building the in-tree std now requires a stage 1 compiler.
+    - Consequently, library contributors are *strongly* encouraged to enable `rust.download-rustc = "if-unchanged"` to avoid having to build a stage 1 compiler. Note that this is the default for `profile = "library"`, but you may need to specify it manually if you don't use a `profile`.
+- `cfg(bootstrap)` should no longer be needed for library sources.
+
+## What does this mean for a typical compiler workflow?
+
+- If unstable library features are being dogfooded, some additional `cfg(bootstrap)` usages may now be needed in compiler sources.
```
jackh726, line 61: I think this would *really* benefit from some example(s). For example, what would this look like for e.g. a new trait solving feature that requires some core lang items?

(Aside, not the right place for this, but this effectively means that such features actually will need to a *stage 2* compiler to be built to properly test them, right?)

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Why are we making this change?
+
+The previous stage 0 bootstrapping sequence was a source of endless confusion for compiler, library and bootstrap contributors alike, because std had to support being built by *both* a previous beta rustc and in-tree rustc, with `cfg(bootstrap)` in std sources necessary to distinguish between them. By adjusting the stage 0 bootstrap sequence to instead use a precompiled stage 0 std instead of building the in-tree std, we hope to:
```
jackh726, line 65: Said above, a very brief summary of this could be put at the beginning.

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. Simplify library development workflow to no longer need `cfg(bootstrap)`, and
```
jackh726, line 67: Also an aside and shouldn't really be discussed here, but I personally think `cfg(bootstrap)` in library code is likely easier to maintain and reason about than `cfg(bootstrap)` in compiler code.

Though, perhaps I'm wrong, in which case I think this blog post can do better at convincing me this isn't the case.

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+[extra]
+team = "Bootstrap"
+team_url = "https://www.rust-lang.org/governance/teams/infra#team-bootstrap"
++++
+
+# Redesigning stage 0 std
+
+## Summary
+
+We are reworking how the stage 0 bootstrap sequence works (the sequence used to build a stage 1 compiler).
```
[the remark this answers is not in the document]
jieyouxu: Yeah, previously the motivation was up front, I'll need to balance it a bit.

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+path = "inside-rust/9999/12/31/redesigning-stage-0-std"
+title = "Redesigning stage 0 std"
+authors = ["Jieyou Xu"]
+
+[extra]
+team = "Bootstrap"
+team_url = "https://www.rust-lang.org/governance/teams/infra#team-bootstrap"
++++
+
+# Redesigning stage 0 std
```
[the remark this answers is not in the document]
jieyouxu: I might rename this to sth like "Redesigning the stage 0 bootstrap sequence", not sure if that makes more sense?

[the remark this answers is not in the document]
Kobzol: Btw the heading should be removed, the `title` attribute in the frontmatter is authoritative. Otherwise it creates this:
image

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Summary
+
+We are reworking how the stage 0 bootstrap sequence works (the sequence used to build a stage 1 compiler).
+
+- Before: a stage 0 (beta by default) compiler is used to build the in-tree std, which in turn is used to build the stage 1 compiler.
+- After: a stage 0 (beta by default) compiler and a precompiled stage 0 std is instead used to build the stage 1 compiler.
+
```
BoxyUwU, line 19: I wonder if this could do with a short summary with less of the... I hesitate to say jargon, but something like:
```suggestion
- Before: a stage 0 (beta by default) compiler is used to build the in-tree std, which in turn is used to build the stage 1 compiler.
- After: a stage 0 (beta by default) compiler and a precompiled stage 0 std is instead used to build the stage 1 compiler.

In other words, in the typical workflow, the stage1 compiler now depends on beta std instead of the in-tree std (built by the beta compiler).
```
Generally I find the current "before/after" description to be a little bit hard to parse :thinking:

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+![Difference in stage 0 bootstrap sequence](stage0-redesign-diff.svg)
+
+Notably, this means that after [redesign stage 0 std #119899](https://github.com/rust-lang/rust/pull/119899) PR lands:
+
+- `./x {build,test,check} library --stage 0` becomes no-op, as stage 0 std is no longer the built in-tree std, and the minimum supported stage to build std is now `1`.
+    - Consequently, default (test, check, bench) stage values in the library profile are no longer `0`, but instead defaults to `1`.
+- Some additional `cfg(bootstrap)` usages may be needed in the compiler sources for dogfooding unstable library features.
```
BoxyUwU, line 26: I'm not sure how you'd fit in but I wonder if it'd be helpful to say something like "stage0 now always refers to things from the beta toolchain (unless otherwise customized)".

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+team_url = "https://www.rust-lang.org/governance/teams/infra#team-bootstrap"
++++
+
+# Redesigning stage 0 std
+
+## Summary
+
+We are reworking how the stage 0 bootstrap sequence works (the sequence used to build a stage 1 compiler).
+
+- Before: a stage 0 (beta by default) compiler is used to build the in-tree std, which in turn is used to build the stage 1 compiler.
```
BoxyUwU, line 17: I also think that having to clarify that stage0 is "technically not always the beta compiler" makes things a bit harder to understand :thinking: Almost every contributor is going to have stage0 being the beta compiler so even if its not necessarily true I do wonder if it would make sense to lie a bit with a footnote the first time and then just ignore it from then on.
```suggestion
- Before: the beta compiler[^1] is used to build the in-tree std, which in turn is used to build the stage 1 compiler.

[^1]: In practice, rarely, the stage0 compiler can be something other than the beta compiler. This doesn't matter too much for the purpose of this change though, so we've taken the liberty of conflating the two to make it easier to explain the change to bootstrapping.
```

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+| `./x {check,build,test} library`           | Checks/builds/tests in-tree library                         | Check/build/tests stage 1 library                        |
+| `./x {check,build,test} library --stage 0` | Checks/builds/tests in-tree library                         | No-op in case of build, checks/tests precompiled library |
+| `./x {check,build,test} library --stage 1` | Builds in-tree library, checks/builds/tests stage 1 library | Check/build/tests stage 1 library                        |
+
+For `profile = "compiler"`:
+
+| Invocation                                 | Before stage 0 std redesign                                 | After stage 0 std redesign                               |
+|--------------------------------------------|-------------------------------------------------------------|----------------------------------------------------------|
+| `./x check library`                        | Checks in-tree library                                      | Check/build/tests stage 1 library                        |
```
BoxyUwU, line 46:
```suggestion
| `./x check library`                        | Checks in-tree library                                      | Check stage 1 library                        |
```
or does `x check library` actually build and test it?

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+| `./x {build,test} library`                 | Builds in-tree library, builds/tests stage 1 library        | Check/build/tests stage 1 library                        |
```
BoxyUwU, line 47:
```suggestion
| `./x {build,test} library`                 | Builds in-tree library, builds/tests stage 1 library        | build/tests stage 1 library                        |
```

content/inside-rust/stage0-redesign/index.md
jieyouxu: No, that's a mistake, good catch

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+- Before: a stage 0 (beta by default) compiler is used to build the in-tree std, which in turn is used to build the stage 1 compiler.
+- After: a stage 0 (beta by default) compiler and a precompiled stage 0 std is instead used to build the stage 1 compiler.
+
+![Difference in stage 0 bootstrap sequence](stage0-redesign-diff.svg)
+
+Notably, this means that after [redesign stage 0 std #119899](https://github.com/rust-lang/rust/pull/119899) PR lands:
+
+- `./x {build,test,check} library --stage 0` becomes no-op, as stage 0 std is no longer the built in-tree std, and the minimum supported stage to build std is now `1`.
+    - Consequently, default (test, check, bench) stage values in the library profile are no longer `0`, but instead defaults to `1`.
```
meithecatte, line 25:
```suggestion
    - Consequently, default (test, check, bench) stage values in the library profile are no longer `0`, but instead default to `1`.
```

content/inside-rust/stage0-redesign/index.md
Noratrieb: It may be easier, but it's *much more common*. Instances of cfg(bootstrap) should be really rare, while in the library it's incredibly common, especially for larger refactors *that were simply not done previously because it's so bad*

content/inside-rust/stage0-redesign/index.md
Noratrieb, line 65: I have an alternative description of the motivation that I think more accurately captures the effect on combined compiler/library changes. Feel free to merge and integrate or so whatever with it that you want:

The compiler and standard library are tightly coupled. Intrinsics and lang items form a broad interface between the two, and whenever it is changed, both sides need to adjust. But the standard library supports being compiled with two different compilers, the in-tree compiler and the stage0 (beta) compiler. This means that all such changes require `cfg(bootstrap)` to support both compilers, which results in a lot of churn of anyone wanting to improve things that touch this area.

content/inside-rust/stage0-redesign/index.md
apiraino: > "If unstable library features are being dogfooded"

dogfooding = technical lingo, which might be ok if we assume a specific target audience. Could it be reworded to be less technical?

jieyouxu: I can use "experimenting with unstable library features", sure.

This internal blog post is targeted towards people (compiler/library contributors) who'd already be exposed to terms like "dogfooding", but "experimenting" is less opaque, so sure.

content/inside-rust/stage0-redesign/index.md
jieyouxu: Yeah, I think that's a way better motivation summary. I'll need to think about how to reword this section.

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,70 @@
...
+| `./x {check,build,test} library --stage 1` | Builds in-tree library, checks/builds/tests stage 1 library | Check/build/tests stage 1 library                        |
+
+For `profile = "compiler"`:
+
+| Invocation                                 | Before stage 0 std redesign                                 | After stage 0 std redesign                               |
+|--------------------------------------------|-------------------------------------------------------------|----------------------------------------------------------|
+| `./x check library`                        | Checks in-tree library                                      | Check/build/tests stage 1 library                        |
+| `./x {build,test} library`                 | Builds in-tree library, builds/tests stage 1 library        | Check/build/tests stage 1 library                        |
+| `./x {check,build,test} library --stage 0` | Checks/builds/tests in-tree library                         | No-op in case of build, checks/tests precompiled library |
```
clubby789, line 48: What does it mean to check the precompiled library?

jieyouxu: It's a no-op, I'll reword that

content/inside-rust/stage0-redesign/index.md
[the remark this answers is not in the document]
jieyouxu: Renamed to

> Redesigning the Initial Bootstrap Sequence

content/inside-rust/stage0-redesign/index.md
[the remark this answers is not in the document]
jieyouxu: Moved motivation upfrond in (round 4).

content/inside-rust/stage0-redesign/index.md
jieyouxu: Included in Motivation (round 4).

content/inside-rust/stage0-redesign/index.md
jieyouxu: I tried to include something to that effect.

content/inside-rust/stage0-redesign/index.md
jieyouxu: I used a footnote to "uhh acktually" the first instance of assuming stage 0 compiler in this blog post as an simplifying assumption, but also noted that it's not universally true.

content/inside-rust/stage0-redesign/index.md
jieyouxu: I dropped these "specific" invocations in (round 4). I found them more noise than signal.

content/inside-rust/stage0-redesign/index.md
jieyouxu: I dropped these "specific" invocations in (round 4). I found them more noise than signal.

content/inside-rust/stage0-redesign/index.md
jieyouxu: I dropped these "specific" invocations in (round 4). I found them more noise than signal.

content/inside-rust/stage0-redesign/index.md
jieyouxu: I tried to word the motivation as something to that effect, still not very convincing IMO. (round 4).

content/inside-rust/stage0-redesign/index.md
jieyouxu: I dropped these "specific" invocations in (round 4). I found them more noise than signal.

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,123 @@
...
+[clang-bootstrap]: https://llvm.org/docs/AdvancedBuilds.html#bootstrap-builds
+[zulip-support-thread]: https://rust-lang.zulipchat.com/#narrow/channel/326414-t-infra.2Fbootstrap/topic/Stage.200.20std.20redesign.20support.20thread/with/515096924
+[stage0-redesign-pr]: https://github.com/rust-lang/rust/pull/119899
+
+<!--
+*compiler*<sub>initial</sub>
+*compiler*<sub>dev</sub>
+*compiler*<sub>dist</sub>
+*compiler*<sub>reproducible</sub>
+-->
```
jieyouxu, line 123: FIXME(jieyouxu): remove, is outdated

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,116 @@
...
+2. It significantly reduces cognitive complexity, as the redesigned bootstrap sequence is much more coherent and aligns better with how contributors expect the staging to work. We no longer have a "strange" setup where the stage 1 compiler was built from a *beta* rustc with an *in-tree* std. Now, the stage 1 compiler is built from a beta rustc and a beta std.
+
+# In terms of bootstrap invocations and bootstrap config, what does this redesign mean?
+
+The minimum stage to check, build and test the standard library is now stage 1. `./x {check,build,test} library --stage=0` are now no-ops.
+
+For `profile = "library"` users, like aforementioned, the default check, build and test stage are now bumped to 1. `download-rustc = "if-unchanged"` is enabled by default so a pre-built CI rustc is used to help you avoid needing to rebuild the compiler while working on the standard library if there are no compiler changes.
```
jyn514, line 86: i find this very hard to parse given the comma clause earlier.
```suggestion
For `profile = "library"` users, the default check, build, and test stage are now bumped to 1. `download-rustc = "if-unchanged"` is enabled by default, which downloads a pre-built CI rustc instead of building the compiler if there are no compiler changes, allowing you to build the standard library without building the compiler.
```

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,116 @@
...
+# What does this mean for contributors working on the standard library and the compiler?
+
+- Contributors will now no longer need to use `cfg(bootstrap)` for intrinsics and lang items.
+- Contributors may (rarely) need to use `cfg(bootstrap)` in compiler code if they wish to experiment with unstable library features [^dogfood-unstable-lib-features]
+
+# Frequently asked questions (FAQs) {#faqs}
+
+## Doesn't this just shift `cfg(bootstrap)` from library code to compiler code? {#faqs-shift-cfg-bootstrap}
+
+Not quite. `cfg(bootstrap)` usage in standard library code for using new intrinsics / lang items (as in the current bootstrap sequence) is much more common than potential `cfg(bootstrap)` usage in compiler code for experimenting with unstable library features (as in the redesigned bootstrap sequence). This is because the standard library need to depend on compiler-provided lang items and intrinsics, but the compiler does not (need to) depend on standard library implementation details.
```
jyn514, line 97:
```suggestion
Not quite. `cfg(bootstrap)` usage in standard library code for using new intrinsics / lang items (as in the current bootstrap sequence) is much more common than potential `cfg(bootstrap)` usage in compiler code for experimenting with unstable library features (as in the redesigned bootstrap sequence). This is because the standard library need to depend on compiler-provided lang items and intrinsics, but the compiler does not (need to) depend on standard library implementation details.

Additionally, the compiler only needs to add `cfg(bootstrap)` for unstable library features *not yet on beta*. By our count, the compiler has not used a feature added that recently since before 1.61.
```

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Questions, feedback, bugs?
+
+You can leave a comment in the [zulip support thread for the initial bootstrap sequence redesign effort][zulip-support-thread]
+
+
+[^initial-compiler]: For the vast majority of contributors, the stage 0 "initial" compiler is going to be the beta compiler. However, it is possible to override the initial compiler, such as when further optimizing a compiler through PGO/BOLT. In this blog post, we make a simplifying assumption that the stage 0 compiler is the beta compiler, even though this is not universally true.
+[^dogfood-unstable-lib-features]: Newly added unstable library feature may need to wait until a beta bump before it is usable by the compiler.
+[^adapted]: Much of this is adapted from Jyn's excellent blog post [*Why is Rust's build system uniquely hard to use?*][hard-to-use-blog-post].
```
jyn514, line 106:
```suggestion
[^adapted]: Much of this is adapted from jyn's excellent blog post [*Why is Rust's build system uniquely hard to use?*][hard-to-use-blog-post].
```

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+# TL;DR: What is being changed?
+
+We are [redesigning the stage 0 bootstrap sequence][stage0-redesign-pr] so that the standard library will now only support one compiler version. Building the stage 1 compiler will no longer involve building the in-tree standard library. Instead, the pre-built beta standard library will be used.
+
+The following section is a quick primer on the concept of bootstrapping and the terminology we use in this blog post.
+
+# A quick primer on bootstrapping and terminology used in this blog post {#preliminary}
+
+This section is intended to explain some basic bootstrapping concepts to make it easier for contributors to understand the upcoming bootstrap changes. You can skip this section if you are already familiar with [the `bootstrap` build system][bootstrap] itself.
```
jyn514, line 21:
```suggestion
This section is intended to explain some basic bootstrapping concepts to make it easier for contributors to understand the upcoming bootstrap changes. You can [skip this section](#motivation) if you are already familiar with [the `bootstrap` build system][bootstrap] itself.
```

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+You can leave a comment in the [zulip support thread for the initial bootstrap sequence redesign effort][zulip-support-thread]
+
+
+[^initial-compiler]: For the vast majority of contributors, the stage 0 "initial" compiler is going to be the beta compiler. However, it is possible to override the initial compiler, such as when further optimizing a compiler through PGO/BOLT. In this blog post, we make a simplifying assumption that the stage 0 compiler is the beta compiler, even though this is not universally true.
+[^dogfood-unstable-lib-features]: Newly added unstable library feature may need to wait until a beta bump before it is usable by the compiler.
+[^adapted]: Much of this is adapted from Jyn's excellent blog post [*Why is Rust's build system uniquely hard to use?*][hard-to-use-blog-post].
+
+[rust-lang/rust]: https://github.com/rust-lang/rust
+[bootstrap]: https://github.com/rust-lang/rust/tree/master/src/bootstrap
```
jyn514, line 109:
```suggestion
[bootstrap]: https://rustc-dev-guide.rust-lang.org/building/bootstrapping/intro.html
```

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,116 @@
...
+# Why is the redesigned stage 0 bootstrap sequence better? {#better}
+
+There are several benefits of the redesigned stage 0 bootstrap sequence:
+
+1. We no longer have to use `cfg(bootstrap)` in the standard library sources for intrinsics and lang items to distinguish when being built by the beta rustc vs the in-tree rustc, because the standard library now only has to be buildable by exactly one compiler version (the current stage rustc).
+2. It significantly reduces cognitive complexity, as the redesigned bootstrap sequence is much more coherent and aligns better with how contributors expect the staging to work. We no longer have a "strange" setup where the stage 1 compiler was built from a *beta* rustc with an *in-tree* std. Now, the stage 1 compiler is built from a beta rustc and a beta std.
+
+# In terms of bootstrap invocations and bootstrap config, what does this redesign mean?
+
+The minimum stage to check, build and test the standard library is now stage 1. `./x {check,build,test} library --stage=0` are now no-ops.
```
jyn514, line 84:
```suggestion
The minimum stage to check, build and test the standard library is now stage 1. `./x {check,build,test} library --stage=0` are now no-ops; switch to `--stage 1` instead. `--keep-stage-std=0` is a no-op.
```

jyn514 (approved):
this is really good work. thanks for writing it up, i can tell you spent a lot of time on it.

content/inside-rust/stage0-redesign/index.md
jieyouxu: Reworded (round 5).

content/inside-rust/stage0-redesign/index.md
jieyouxu: Added explanation (round 5).

content/inside-rust/stage0-redesign/index.md
jieyouxu: Added link (round 5).

content/inside-rust/stage0-redesign/index.md
jieyouxu: Added note (round 5).

jieyouxu:
Addressed feedback in (round 5):

- Added a specific example re. the trait solving feature requiring addition of a new lang item.
- Fixed various wording and links.

content/inside-rust/stage0-redesign/index.md
jieyouxu, line 1: FIXME(jieyouxu): squash commits when ready

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,126 @@
...
+path = "inside-rust/9999/12/31/redesigning-the-initial-bootstrap-sequence"
+title = "Redesigning the Initial Bootstrap Sequence"
+authors = ["Jieyou Xu"]
+
+[extra]
+team = "the Bootstrap team"
+team_url = "https://www.rust-lang.org/governance/teams/infra#team-bootstrap"
++++
+
+This blog post accompanies an [upcoming major change to the `bootstrap` build system][stage0-redesign-pr] (see also [Major Change Proposal 619][redesign-stage0-mcp]), which is expected to affect contributors who works on the `rustc` compiler, the standard library, tools, and the distributors of rust toolchains.
```
Mark-Simulacrum, line 11:
```suggestion
This blog post accompanies an [upcoming major change to the rust-lang/rust build system][stage0-redesign-pr] (see also [Major Change Proposal 619][redesign-stage0-mcp]). This will have no impact on the distributed artifacts from that repository, but the way we build those artifacts is changing.
```

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,126 @@
...
+
+# TL;DR: What is being changed?
+
+We are [redesigning the stage 0 bootstrap sequence][stage0-redesign-pr] so that the standard library will now only support one compiler version. Building the stage 1 compiler will no longer involve building the in-tree standard library. Instead, the pre-built beta standard library will be used.
```
Mark-Simulacrum, line 15: I'm not sure if I would use "stage 0" or "stage 1" in the TL;DR section.

One thought is that explaining this by providing the x.py build log before and after the change, or some similarly visual-ish comparison might be clearer than trying to *name* things (e.g., "Building the stage 1 compiler ..." is probably ambiguous to the average reader, I think you mean the compiler that ends up in build/host/stage1/bin/rustc, but that's in some sense the 0th compiler we build...).

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,126 @@
...
+
+This section is intended to explain some basic bootstrapping concepts to make it easier for contributors to understand the upcoming bootstrap changes. You can [skip this section](#motivation) if you are already familiar with [the `bootstrap` build system][bootstrap] itself.
+
+## What is bootstrapping? {#intro-to-bootstrapping}
+
+[*Bootstrapping*][compiler-bootstrapping] is the process of producing a newer version of a compiler with an older version of itself. Usually, bootstrap processes involve the concept of *stages*, where a compiler and associated artifacts from a stage is used to build the compiler of the next stage. For example, a stage 2 compiler is built from a stage 1 compiler and associated stage 1 artifacts. The stage number is named after the compiler in question.
+
+## Bootstrapping `rustc` {#bootstrapping-rust}
+
+C compilers like [`gcc`][gcc-bootstrap] or [`clang`][clang-bootstrap] do not build their standard libraries from source, but instead links against the same specified standard library across all bootstrap stages. Typically, this specified standard library is the system standard library.[^adapted]
```
Mark-Simulacrum, line 29:
```suggestion
Building a C++ compiler toolchain like [`gcc`][gcc-bootstrap] or [`clang`][clang-bootstrap] doesn't usually involve building their standard libraries from source, but instead will link against a prebuilt standard library across all bootstrap stages. Typically, this specified standard library is the system standard library.[^adapted]
```

I don't think C has much of a standard library, at least not with the compiler? C++ seems like the closer example.

It also seems key to note that the library *is* built from source, it's just not part of the build process for the compiler...

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,126 @@
...
+
+In the context of `rustc`, this is not the case, as the compiler and standard library are tightly coupled through *intrinsics* and *lang items*. Bootstrapping `rustc` involves building the standard library from source. Building a stage `N` `rustc` requires building or acquiring the stage `N - 1` `rustc`, as well as the stage `N - 1` standard library built by the stage `N - 1` `rustc`.
```
Mark-Simulacrum, line 31: It feels a little misleading to say that this is not the case for rustc because X when that X is not changing with this change -- both lang items and intrinsics are here to stay, and they're just as tightly coupled to a compiler version.

In fact, the reality that we *can* make this change implies this is basically an arbitrary choice for Rust: do we prefer cfg(bootstrap) in libstd or in librustc*? (Right?)

Not sure if there's a way to rephrase this to avoid that implication...

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,126 @@
...
+- Contributors will now no longer need to use `cfg(bootstrap)` for intrinsics and lang items.
+- Contributors may (rarely) need to use `cfg(bootstrap)` in compiler code if they wish to experiment with unstable library features [^dogfood-unstable-lib-features]
+
+# Frequently asked questions (FAQs) {#faqs}
+
+## Doesn't this just shift `cfg(bootstrap)` from library code to compiler code? {#faqs-shift-cfg-bootstrap}
+
+Not quite. `cfg(bootstrap)` usage in standard library code for using new intrinsics / lang items (as in the current bootstrap sequence) is much more common than potential `cfg(bootstrap)` usage in compiler code for experimenting with unstable library features (as in the redesigned bootstrap sequence). This is because the standard library need to depend on compiler-provided lang items and intrinsics, but the compiler does not (need to) depend on standard library implementation details.
+
+Additionally, the compiler only needs to add `cfg(bootstrap)` for unstable library features *not yet on beta*. By our count, the compiler has not used a feature added that recently since before 1.61.
```
Mark-Simulacrum, line 101: "not yet on beta" feels a bit misleading here -- what it really wants to say is anything in the standard library that has changed unstable API and is used in the compiler.

(Out of interest, how did we perform this analysis?)

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,126 @@
...
+# What does this mean for contributors working on the standard library and the compiler?
+
+- Contributors will now no longer need to use `cfg(bootstrap)` for intrinsics and lang items.
+- Contributors may (rarely) need to use `cfg(bootstrap)` in compiler code if they wish to experiment with unstable library features [^dogfood-unstable-lib-features]
+
+# Frequently asked questions (FAQs) {#faqs}
+
+## Doesn't this just shift `cfg(bootstrap)` from library code to compiler code? {#faqs-shift-cfg-bootstrap}
+
+Not quite. `cfg(bootstrap)` usage in standard library code for using new intrinsics / lang items (as in the current bootstrap sequence) is much more common than potential `cfg(bootstrap)` usage in compiler code for experimenting with unstable library features (as in the redesigned bootstrap sequence). This is because the standard library need to depend on compiler-provided lang items and intrinsics, but the compiler does not (need to) depend on standard library implementation details.
```
Mark-Simulacrum, line 99:
```suggestion
Not quite. `cfg(bootstrap)` usage in standard library code for using new intrinsics / lang items (as in the current bootstrap sequence) is much more common than potential `cfg(bootstrap)` usage in compiler code for experimenting with unstable library features (as in the redesigned bootstrap sequence). This is because the standard library must be changed for new compiler-provided lang items and intrinsics, but the compiler does not (need to) depend on recently added standard library APIs.
```

content/inside-rust/stage0-redesign/stage0-current.svg
Mark-Simulacrum, line 1: Possibly worth calling out: stage 0 std is also downloaded and used in-tree for ToolBootstrap I believe.

Not sure how relevant that is to the diagram here though.

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,126 @@
...
+
+Currently, the standard library currently must support being built with two different compilers, the in-tree compiler and the initial stage 0 compiler[^initial-compiler]. All such changes to intrinsics and lang items thus need to use `cfg(bootstrap)` to gate code that can be built by the in-tree compiler vs the stage 0 compiler. This causes a lot of churn for contributors wanting to introduce, modify or remove intrinsics and lang items (particularly when creating new releases).
+
+The [stage 0 bootstrap sequence redesign][stage0-redesign-pr] aims to mitigate such churn and implementation complexity in the standard library by having the standard library only support *one* version of the compiler.
+
+To better understand this redesign, we will:
+
+1. [Explain how the current stage 0 bootstrapping sequence works](#current-model), and
+2. [Explain how the new stage 0 bootstrapping sequence works after the redesign](#new-model), and
+3. [Discuss why the redesigned stage 0 bootstrapping sequence is more preferable](#better).
```
Mark-Simulacrum, line 55:
```suggestion
3. [Explain the benefits of the redesigned stage 0 bootstrapping sequence](#better).
```

content/inside-rust/stage0-redesign/index.md
Mark-Simulacrum, line 1: I'm not sure what the intended *goal* for this post is -- are we seeking to tell people what they should expect to change in their workflows? Just trying to communicate that the change is happening? Something else?

If I've understood the content correctly, we're sort of just saying "hey stuff is going to look different" but I *think* we're not expecting local workflows to actually meaningfully change in terms of e.g. commands to run or similar? I wonder if making that explicit at the top of the post ("Changes are coming to how things are built, but the majority of contributors shouldn't see any meaningful impact to the commands they run. Please reach out in XXX if that is not the case for you").

jieyouxu: > I'm not sure what the intended goal for this post is -- are we seeking to tell people what they should expect to change in their workflows? Just trying to communicate that the change is happening?

Yes, the intention is to better communicate the change is happening, and help prepare contributors for the differences instead of suddenly finding out staging is different.

content/inside-rust/stage0-redesign/index.md
jieyouxu: > In fact, the reality that we can make this change implies this is basically an arbitrary choice for Rust: do we prefer cfg(bootstrap) in libstd or in librustc*? (Right?)

Yeah, this is the tradeoff that we're changing here, to prefer the less-common `cfg(bootstrap)` for compiler instead of the more frquent `cfg(bootstrap)` for std.

content/inside-rust/stage0-redesign/index.md
Mark-Simulacrum: I think I'd encourage structuring the post to highlight the actual impact we expect to people's workflows (e.g., cfg(bootstrap), different commands to run) up front. IOW, if I link someone this post, the likelihood that they have been impacted ("help I was doing X and that doesn't work anymore") seems high -- we should have a clear guidance on that *somewhere* for the cases we expect to hit. Maybe that's a FAQ on the dev guide or similar though, not this post, but I think that's largely missing in the current contents.

I think having the "mental model" explanation is *good* -- it helps people debug why things aren't working without needing a perfect match to a workflow we call out in the FAQ -- but I think both are needed.

content/inside-rust/stage0-redesign/index.md
jieyouxu: In some sense, yeah, this wording is... hard.

Mark-Simulacrum: FWIW having seen the diagrams... it might be OK to just have those shown here? Maybe with the exact text that x.py prints prior to running that section of the build?

Those seem like a very nice before/after comparison to me...

content/inside-rust/stage0-redesign/index.md
jieyouxu: Right.

> (Out of interest, how did we perform this analysis?)

I can't claim to have done this analysis myself (though it *seems* reasonably accurate to  me). I believe @jyn514 did the analysis originally?

content/inside-rust/stage0-redesign/index.md
jieyouxu: I might drop the "stage 0" vs "stage 1 " here, and instead use "beta" and "in-tree" instead, *even though* this is not technically always the case (e.g. stage 0 doesn't have to be beta). I can move the "simplifying assumption" footnote earlier to here.

content/inside-rust/stage0-redesign/index.md
jieyouxu: Dropped C, stick with C++ and fixed wording as suggested (round 6).

content/inside-rust/stage0-redesign/index.md
jieyouxu: Fixed wording (round 6)

content/inside-rust/stage0-redesign/stage0-current.svg
jieyouxu: I decided not to include that technicality as I find it's probably more confusing for contributors not working on bootstrap than helpful (maybe)?

content/inside-rust/stage0-redesign/index.md
jieyouxu: Fixed (round 6)

content/inside-rust/stage0-redesign/index.md
jieyouxu: I'm not quite sure how to restructure the current contents tbh. I think this blog post might be better off as the "motivation" and "high-level overview of what is being changed". For specific breakages (and workflow changes), we probably need to document them in rustc-dev-guide. We technically can amend this blog post in the future for some more backlinks if needed. Would that work for you?

jieyouxu:
Tried to address some more feedback in (round 6).

content/inside-rust/stage0-redesign/index.md
Mark-Simulacrum: Documenting in rustc-dev-guide and linking there seems good -- maybe something like "See this page for details on how you might need to change your workflow with these changes. This blog post focuses on motivation for the change and attempts to build a mental model for how the system works, rather than deep diving on workflow changes."

jieyouxu: > Documenting in rustc-dev-guide and linking there seems good -- maybe something like "See this page for details on how you might need to change your workflow with these changes. This blog post focuses on motivation for the change and attempts to build a mental model for how the system works, rather than deep diving on workflow changes."

I'll leave a paragraph to that effect, and backlink to https://rustc-dev-guide.rust-lang.org/building/bootstrapping/what-bootstrapping-does.html for now (this page is updated in the stage 0 redesign PR, and it's possible to flush out more concrete examples on that page or backlinked from that page if necessary).

Mark-Simulacrum: That page doesn't seem to have any workflow details? I was expecting something like a page containing the lang item example ("> *Example: Implementing a trait solving feature which requires adding core lang items*") amongst other sample cases.

Perhaps that's a new page? I'm not sure where that example comes from, maybe we should just link to that :)

jieyouxu: > Perhaps that's a new page? I'm not sure where that example comes from, maybe we should just link to that :)

To clarify, this is not a page that yet exists (we'll have to write it in the redesign PR).

This example is taken from the discussion in Jack's [review comment earlier](https://github.com/rust-lang/blog.rust-lang.org/pull/1582#discussion_r2066894701).

content/inside-rust/stage0-redesign/index.md
jyn514: > I can't claim to have done this analysis myself (though it _seems_ reasonably accurate to me). I believe @jyn514 did the analysis originally?

see https://jyn.dev/bootstrapping-rust-in-2023/#what-can-we-do-about-it and in particular https://github.com/jyn514/rust/tree/versions-used

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+This section is intended to explain some basic bootstrapping concepts to make it easier for contributors to understand the upcoming bootstrap changes. You can [skip this section](#motivation) if you are already familiar with [the `bootstrap` build system][bootstrap] itself.
+
+## What is bootstrapping? {#intro-to-bootstrapping}
+
+[*Bootstrapping*][compiler-bootstrapping] is the process of producing a newer version of a compiler with an older version of itself. Usually, bootstrap processes involve the concept of *stages*, where a compiler and associated artifacts from a stage is used to build the compiler of the next stage. For example, a stage 2 compiler is built from a stage 1 compiler and associated stage 1 artifacts. The stage number is named after the compiler in question.
+
+## Bootstrapping `rustc` {#bootstrapping-rust}
+
+Building a C++ compiler toolchain like [`gcc`][gcc-bootstrap] or [`clang`][clang-bootstrap] doesn't usually involving building their stand libraries from source, but instead will link against a pre-built standard library across all bootstrap stages. Typically, this specified standard library is the system standard library.[^adapted]
```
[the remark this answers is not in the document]
jieyouxu: Fixed (round 8)

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+In the context of `rustc`, this is not the case, as the compiler and standard library are tightly coupled through *intrinsics* and *lang items*. Bootstrapping `rustc` involves building the standard library from source. Building a stage `N` `rustc` requires building or acquiring the stage `N - 1` `rustc`, as well as the stage `N - 1` standard library built by the stage `N - 1` `rustc`.
```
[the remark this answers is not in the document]
jieyouxu: Tried to fix this in, not super sure (round 8).

jieyouxu:
Addressed review feedback (round 8).

content/inside-rust/stage0-redesign/index.md
```diff
@@ -0,0 +1,142 @@
...
+authors = ["Jieyou Xu"]
+
+[extra]
+team = "the Bootstrap team"
+team_url = "https://www.rust-lang.org/governance/teams/infra#team-bootstrap"
++++
+
+This blog post accompanies an [upcoming major change to the `rust-lang/rust` build system][stage0-redesign-pr] (see also [Major Change Proposal 619][redesign-stage0-mcp]). This will have no impact on the distributed artifacts from [rust-lang/rust], but the way we build those artifacts is changing.
+
+This blog post focuses on motivation for the change and attempts to build a mental model for how the system works, rather than deep diving on workflow changes. See the [rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/building/bootstrapping/what-bootstrapping-does.html) for details on how you might need to change your workflow with these changes.
```
Mark-Simulacrum, line 13:
```suggestion
This blog post focuses on motivation for the change and attempts to build a mental model for how the system works, rather than deep diving on workflow changes. See the [rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/building/bootstrapping/what-bootstrapping-does.html) for details on how you might need to change your workflow with these changes -- note that it will have new content once the new implementation merges.
```

Or something like that?

jieyouxu: Good point, also note that "PR merges" is not quite sufficient, we also need to do the rust-lang/rust -> rustc-dev-guide sync direction, and only *then* will the actual r-d-g website show the new content.

content/inside-rust/stage0-redesign/index.md
jieyouxu: This was reworded as recommended in (round 8).

content/inside-rust/stage0-redesign/index.md
jieyouxu: We'll probably have to update the page as new concrete examples come up.

onur-ozkan (approved):
LGTM, thank you for working on this!

jieyouxu:
Ok, I resolved two outstanding comments that AFAICT are not blocking.

PR #1584, opened 2025-04-17
Improve error message in case of invalid path

senekor:
Previously, if blog authors didn't set the path key in the front matter, the error message from Zola was:

Filter `nth` received an incorrect type for arg `n`: got `-2` but expected a usize

This is misleading, as it doesn't point out the cause of the problem. The new error message makes authors aware of the required path format.


senekor:
this was reported [here](https://github.com/rust-lang/blog.rust-lang.org/pull/1582#issuecomment-2812022000)

Manishearth (approved):

PR #1588, opened 2025-04-18
Add blog post generator

### LINKED ISSUE #1583: Interactive blog post generator
Blog post authors have to keep a couple things in mind, especially when it comes to the front matter. For now, we try to document this in the readme. That puts the burden on our authors.

Instead, we could have a little program that prompts authors interactively when they're creating a new post.

There is a crate called inquire or something that should make this nice.

With a cargo alias (inspired by cargo xtask) authos could run e.g. `cargo blog` to generate a new post.

Questions might include:
- title (then auto-generate slug and filename from it)
- date (placeholder if unknown)
- inside-rust yes/no
- release post (-> auto-generate aliases)
- team (if yes, also prompt for url)

At the end, valid usable front matter should be generated with minimal effort from the author.

Lastly, as a bonus, the generated post could include some boilerplate markdown that shows all the features available. (I for one always forget how to do footnotes in markdown)

senekor:
This makes it easier for blog authors to get started without having to figure out the front matter format first. All they have to do is run `cargo blog`. The generator asks all relevant questions to generate valid and complete front matter.

closes #1583


crates/generate_blog/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+            Blog::InsideRust => write!(f, r#"The "Inside Rust" blog"#),
+        }
+    }
+}
+
+fn try_parse_version_from_title(title: &str) -> Option<String> {
+    // assuming there won't ever be a Rust 2.0
+    let (_, rest) = title.split_once("1.")?;
+    let (minor, rest) = rest.split_once('.')?;
+    let patch = rest[0..1].parse::<u8>().unwrap_or_default();
```
programmerjake, line 184: this should use `str::get` in case that character isn't ASCII

senekor: Unrealistic but ok :)

Manishearth (approved):

PR #1593, opened 2025-04-25
posts: announce rustup 1.28.2

rami3l:
cc @rust-lang/rustup 

[Rendered](https://github.com/rami3l/blog.rust-lang.org/blob/post/rustup-1-28-2/content/Rustup-1.28.2.md)

### DIFF
--- a/content/Rustup-1.28.2.md
+++ b/content/Rustup-1.28.2.md
@@ -1,0 +1,85 @@
++++
+path = "2025/04/25/Rustup-1.28.2"
+title = "Announcing rustup 1.28.2"
+authors = ["The Rustup Team"]
+aliases = ["2025/04/25/Rustup-1.28.2.html"]
++++
+
+The rustup team is happy to announce the release of rustup version 1.28.2.
+[Rustup][install] is the recommended tool to install [Rust][rust], a programming language that
+empowers everyone to build reliable and efficient software.
+
+## What's new in rustup 1.28.2
+
+The headlines of this release are:
+
+- The cURL download backend and the native-tls TLS backend are now officially deprecated and
+  a warning will start to show up when they are used. [pr#4277]
+
+  - While rustup predates reqwest and rustls, the rustup team has long wanted to standardize on
+    an HTTP + TLS stack with more components in Rust, which should increase security, potentially
+    improve performance, and simplify maintenance of the project.
+    With the default download backend already switched to reqwest since [2019][pr#1660], the team
+    thinks it is time to focus maintenance on the default stack powered by these two libraries.
+
+  - For people who have set `RUSTUP_USE_CURL=1` or `RUSTUP_USE_RUSTLS=0` in their environment to
+    work around issues with rustup, please try to unset these after upgrading to 1.28.2 and file
+    [an issue][issue tracker] if you still encounter problems.
+
+- The version of `rustup` can be pinned when installing via `rustup-init.sh`, and
+  `rustup self update` can be used to upgrade/downgrade rustup 1.28.2+ to a given version.
+  To do so, set the `RUSTUP_VERSION` environment variable to the desired version (for example `1.28.2`).
+  [pr#4259]
+
+- `rustup set auto-install disable` can now be used to disable automatic installation of the toolchain.
+  This is similar to the `RUSTUP_AUTO_INSTALL` environment variable introduced in 1.28.1 but with a
+  lower priority. [pr#4254]
+
+- Fixed a bug in Nushell integration that might generate invalid commands in the shell configuration.
+  Reinstalling rustup might be required for the fix to work. [pr#4265]
+
+[pr#1660]: https://github.com/rust-lang/rustup/pull/1660
+[pr#4254]: https://github.com/rust-lang/rustup/pull/4254
+[pr#4259]: https://github.com/rust-lang/rustup/pull/4259
+[pr#4265]: https://github.com/rust-lang/rustup/pull/4265
+[pr#4277]: https://github.com/rust-lang/rustup/pull/4277
+[issue tracker]: https://github.com/rust-lang/rustup/issues/
+
+## How to update
+
+If you have a previous version of rustup installed, getting the new one is as easy as stopping
+any programs which may be using rustup (e.g. closing your IDE) and running:
+
+```
+$ rustup self update
+```
+
+Rustup will also automatically update itself at the end of a normal toolchain update:
+
+```
+$ rustup update
+```
+
+If you don't have it already, you can [get rustup][install] from the appropriate page on our website.
+
+Rustup's documentation is also available in [the rustup book][book].
+
+## Caveats
+
+Rustup releases can come with problems not caused by rustup itself but just due to having a new release.
+
+In particular, anti-malware scanners might block rustup or stop it from creating or copying
+files, especially when installing `rust-docs` which contains many small files.
+
+Issues like this should be automatically resolved in a few weeks when the anti-malware scanners are updated
+to be aware of the new rustup release.
+
+## Thanks
+
+Thanks again to all the [contributors] who made this rustup release possible!
+
+[book]: https://rust-lang.github.io/rustup/
+[changelog]: https://github.com/rust-lang/rustup/blob/stable/CHANGELOG.md
+[contributors]: https://github.com/rust-lang/rustup/blob/stable/CHANGELOG.md#detailed-changes
+[install]: https://rustup.rs
+[rust]: https://www.rust-lang.org
--- a/content/Rustup-1.28.2.md
+++ b/content/Rustup-1.28.2.md
@@ -1,8 +1,8 @@
 +++
-path = "2025/04/25/Rustup-1.28.2"
+path = "2025/05/05/Rustup-1.28.2"
 title = "Announcing rustup 1.28.2"
 authors = ["The Rustup Team"]
-aliases = ["2025/04/25/Rustup-1.28.2.html"]
+aliases = ["2025/05/05/Rustup-1.28.2.html"]
 +++
 
 The rustup team is happy to announce the release of rustup version 1.28.2.


content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,85 @@
...
+
+  - While rustup predates reqwest and rustls, the rustup team has long wanted to standardize on
+    an HTTP + TLS stack in Rust, which should increase security, potentially improve performance, and
+    simplify maintenance of the project.
+    With the default download backend already switched to reqwest since [2019][pr#1660], the team
+    thinks it is time to start removing the cURL backend and focus on maintaining the Rust-based
+    stack.
+
+  - The rustup team encourages everyone to switch to the reqwest backend, and would love to hear from
+    you about your use case via [GitHub Issues][issue tracker] if it does work against your particular
```
shepmaster, line 27: Does *not* work, perhaps?

rami3l:
Sorry, I forgot to mark it as draft 🙈 it shouldn't really be merged before the release really happens.

djc (approved):

ChrisDenton (approved):

content/Rustup-1.28.2.md
rami3l: @shepmaster The negation is in "against", but I guess it's a bit too easy to misread? Fixed anyway :)

shepmaster: Yes, I would definitely have never read "against" as a negation (and I still don't, even with your response). For example, "I tested my changes against the main branch" has no such negative meaning to me.

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,85 @@
++++
+path = "2025/04/25/Rustup-1.28.2"
+title = "Announcing rustup 1.28.2"
+authors = ["The Rustup Team"]
+aliases = ["2025/04/25/Rustup-1.28.2.html"]
++++
+
+The rustup team is happy to announce the release of rustup version 1.28.2.
+[Rustup][install] is the recommended tool to install [Rust][rust], a programming language that is
+empowering everyone to build reliable and efficient software.
```
shepmaster, line 10:
```suggestion
[Rustup][install] is the recommended tool to install [Rust][rust], a programming language that 
empowers everyone to build reliable and efficient software.
```

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,85 @@
...
+
+## What's new in rustup 1.28.2
+
+The headlines of this release are:
+
+- The cURL download backend is now officially deprecated and a warning will start to show up
+  when it is used. [pr#4277]
+
+  - While rustup predates reqwest and rustls, the rustup team has long wanted to standardize on
```
shepmaster, line 19:
```suggestion
  - The rustup team has long wanted to standardize on
```

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,85 @@
...
+    an HTTP + TLS stack in Rust, which should increase security, potentially improve performance, and
```
shepmaster, line 20: Is the actual stack now all in Rust? I see `native-tls` in the Cargo.lock file, which presumably is non-Rust code.

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,85 @@
...
+    thinks it is time to start removing the cURL backend and focus on maintaining the Rust-based
+    stack.
+
+  - The rustup team encourages everyone to switch to the reqwest backend, and would love to hear from
+    you about your use case via [GitHub Issues][issue tracker] if it does not work well with your
+    particular setup.
+
+- Version pinning of `rustup` is now supported when installing it via `rustup-init.sh` with the `RUSTUP_VERSION`
+  environment variable set to e.g. `1.28.2`; the same configuration will be respected when performing
+  `rustup self update` from this version on, enabling arbitrary downgrades in the meantime. [pr#4259]
```
shepmaster, line 32:
```suggestion
- The version of `rustup` can be pinned when installing via `rustup-init.sh`. To do so, set the `RUSTUP_VERSION` environment variable to the desired version (for example `1.28.2`). 
```

Are there other ways of pinning the version, such as if I've already installed, or *only* via this installation method?

> the same configuration will be respected when performing `rustup self update` from this version on

I don't know what this means. Guessing, it means that `rustup self update` will no longer actually perform an update. I hope that rustup prints a warning to the user that the version is pinned to X at that point and describes how to unpin/change the pin. Guessing further, I assume that you can change the pin by setting the environment variable again when running `rustup self update`.

> enabling arbitrary downgrades in the meantime. 

I don't know what this means. My first reading of this is "the version of rustup with decrease at some point, whenever rustup decides it". I cannot believe that is the actual intention, however. "Arbitrary" is not a feature I look for in my development tools, especially paired with a nebulous timeframe like "meantime".

shepmaster (commented):
Since you asked for it, I put more effort into reading and thinking about what the text meant. Feel free to take or leave any of the suggestions and pay attention to or ignore any of the clarification questions though.

content/Rustup-1.28.2.md
djc: It is not -- reqwest + rustls is now the default but in addition to the cURL fallback there is still a native-tls fallback option.

content/Rustup-1.28.2.md
djc: This seems like useful context to explain why we didn't start out with reqwest/rustls?

content/Rustup-1.28.2.md
weihanglo, line 1: I am aware that we have a call-for-testing post in <https://internals.rust-lang.org/t/seeking-beta-testers-for-rustup-v1-28-2>. Do we want to call for testing also in the "Inside Rust Blog"? I could see this is a patch version release so we skip that part :)

content/Rustup-1.28.2.md
rami3l: @shepmaster Thanks! I think I can change the ending to "... maintaining the reqwest-based stack". A Rust-based stack is our long-term goal but probably it's too early to say that right now.

shepmaster: > there is still a native-tls fallback

Is that intended to be deprecated eventually as well? If not, then it seems like the overall goal isn't "there is only a pure-Rust stack". Since the benefit of "a pure-Rust stack by default" has already occurred, it seems that this change is purely about reducing maintenance burden (still a good thing!).

content/Rustup-1.28.2.md
shepmaster: Maybe, but for a notification post like this, I'm not so worried about why the *previous* decision was made but more interested in what's changing, what I need to do about it, and then maybe why the change is happening. If historical context is needed, I'd hope it could come after the other points have been addressed.

rami3l: I agree with @djc with this, so I guess it'd still be better to keep it here 🙏

content/Rustup-1.28.2.md
ChrisDenton: Oh it might also be worth mentioning this on zulip. One of the things brought up before is that many people involved with rust don't read internals whereas zulip is more active.

content/Rustup-1.28.2.md
rami3l: @shepmaster IIRC the deprecation of `native-tls` is indeed on our current project roadmap.

shepmaster: Depending on how far out you think that may be, it could be worth a tiny mention here and now. That way, people don't jump from cURL to native-tls only to need to switch again, at least without trying the intended stack first (and filing bugs if needed 🤞)

shepmaster (approved):

content/Rustup-1.28.2.md
djc: Fair point; https://github.com/rust-lang/rustup/pull/4317.

rami3l: @djc I have mirrored your change in this PR; I'll soon do the same in https://github.com/rust-lang/rustup/pull/4313.

content/Rustup-1.28.2.md
rami3l: @weihanglo That might indeed be a good idea in general, but the scope of this patch release is relatively small IMHO.

ChrisDenton: @rami3l We should post in general for visibility I think because it should have a wide audience.

ChrisDenton: Perhaps it should link to a t-rustup thread (or vice versa) so there's a cross-reference of where rustup discussions have happened on zulip.

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,84 @@
...
+- The cURL download backend and the native-tls TLS backend are now officially deprecated and
+  a warning will start to show up when they are used. [pr#4277]
+
+  - While rustup predates reqwest and rustls, the rustup team has long wanted to standardize on
+    an HTTP + TLS stack in Rust, which should increase security, potentially improve performance, and
+    simplify maintenance of the project.
+    With the default download backend already switched to reqwest since [2019][pr#1660], the team
+    thinks it is time to focus maintenance on the Rust-based stack.
+
+  - The rustup team encourages everyone to switch to the reqwest backend with rustls, and would love
```
Mark-Simulacrum, line 25: Presumably there's no actual work involved here in this encouragement, right? This is just the default with the new release? Can we rephrase this to make that clear?

content/Rustup-1.28.2.md
rami3l: @Mark-Simulacrum Good point! I've adjusted the phrasing.

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,84 @@
...
+
+  - While rustup predates reqwest and rustls, the rustup team has long wanted to standardize on
+    an HTTP + TLS stack with more components in Rust, which should increase security, potentially
+    improve performance, and simplify maintenance of the project.
+    With the default download backend already switched to reqwest since [2019][pr#1660], the team
+    thinks it is time to focus maintenance on the default stack powered by these two libraries.
+
+  - The rustup team encourages everyone still relying on non-default backends to switch back to
+    reqwest and rustls, and would love to hear from you about your use case via
+    [GitHub Issues][issue tracker] if it does not work well with your particular setup.
```
djc, line 27: I think this could be clearer still, maybe:

"For people who have RUSTUP_USE_CURL=1 or RUSTUP_USE_RUSTLS=0 set in their environment to work around issues with rustup, please try to unset these after upgrading to 1.28.2 and file an issue [..] if you still encounter problems."

rami3l: Fixed!

content/Rustup-1.28.2.md
rami3l: @ChrisDenton Already posted to [#general > Seeking beta testers for rustup v1.28.2](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/Seeking.20beta.20testers.20for.20rustup.20v1.2E28.2E2).

Mark-Simulacrum (approved):

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,85 @@
++++
+path = "2025/04/25/Rustup-1.28.2"
```
Mark-Simulacrum, line 2:
```suggestion
path = "2025/05/05/Rustup-1.28.2"
```

content/Rustup-1.28.2.md
```diff
@@ -0,0 +1,85 @@
...
+title = "Announcing rustup 1.28.2"
+authors = ["The Rustup Team"]
+aliases = ["2025/04/25/Rustup-1.28.2.html"]
```
Mark-Simulacrum, line 5:
```suggestion
aliases = ["2025/05/05/Rustup-1.28.2.html"]
```

PR #1601, opened 2025-04-30
Add autocompletion for team names and URLs

Kobzol:
This should make it simpler to lookup this metadata.

Users will still probably override the team name to get a nicer string in the "on behalf of X" header. We could in theory also add the team type (working group, project group, team, etc.) to help us automatically generate that header string.

### DIFF
--- a/crates/generate_blog/src/main.rs
+++ b/crates/generate_blog/src/main.rs
@@ -96,11 +96,12 @@ fn main() -> Result<(), Box<dyn Error>> {
 
         let url = if let Some(url) = team_data
             .as_ref()
-            .and_then(|teams| find_team_url(teams, &team)) {
+            .and_then(|teams| find_team_url(teams, &team))
+        {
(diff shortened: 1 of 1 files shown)


senekor (approved):
LGTM, can be merged as is.

What's the purpose of prompting for the URL when we were able to determine it? Is there a use case where authors might want to use a non-standard URL for an existing team?

Kobzol:
Hmm, I'm not sure. There are definitely some old posts that have a custom URL, but I consider them to be mostly legacy. I guess that I just wanted to still give users a choice :) Let me know if you want me to automatically select the URL if the team name is known.

senekor:
I'm leaning toward skipping the prompt, users can still change the front matter manually if they absolutely want to.

Kobzol:
Ok, changed it. Also fixed a bug where for some teams the URL was not determined correctly.

PR #1611, opened 2025-05-17
update the std::io::pipe() example to comment deadlocks

oconnor663:
The previous version of the example wasn't a deadlock, but only because the Command object was a temporary. That's quite subtle, and I think it's better to call out the potential deadlock explicitly.

[Rendered](https://github.com/oconnor663/blog.rust-lang.org/blob/pipe_example/content/Rust-1.87.0/index.md)

### DIFF
--- a/content/Rust-1.87.0/index.md
+++ b/content/Rust-1.87.0/index.md
@@ -51,20 +51,25 @@ or platform-specific functions.
 use std::process::Command;
 use std::io::Read;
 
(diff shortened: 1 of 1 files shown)


senekor:
I think this is too big of a change. The post has been released and there isn't anything necessarily wrong with the example. The blog is not documentation that evolves and is polished over time. Post-release updates are usually about fixing typos, broken URLs and so on.

PR #1623, opened 2025-05-27
announce new compiler team members

davidtwco:
Waiting on rust-lang/team#1849

[Rendered](https://github.com/davidtwco/blog.rust-lang.org/blob/compiler-new-team-member-may-25/content/inside-rust/compiler-team-new-members-may-2025.md)

PR #1628, opened 2025-05-30
Fix date of publication

senekor:
[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/senekor-owvnkpytnvlm/content/inside-rust/compiler-team-new-members-may-2025.md)


senekor:
When this post was published, the date was accidentally updated in the `aliases` key, when it shoul've been updated in the `path` key:
https://github.com/rust-lang/blog.rust-lang.org/pull/1623#discussion_r2116105323

I added stricter CI checks to prevent this in the future:
https://github.com/rust-lang/blog.rust-lang.org/pull/1629

PR #1614, opened 2025-05-21
2025h1 project goals April update

lqd:
This is the April 2025h1 project goals update. It's not late at all!

I summarized the All Hands update, turned the async update into its TL;DR, and massaged the RFL one to fix a couple issues.

r? @nikomatsakis

PR #1629, opened 2025-05-30
Forbid migration alias in new posts

senekor:
The change is explained in a comment, copying it here for convenience:

Make sure that posts created after the migration to Zola don't have
the alias for preserving permalinks.

Blog authors would often copy-paste an old post, reusing the
boilerplate for a new one. In that process, they'd usually preserve
that alias as well, probably not knowing what it's for. Technically
such an alias doesn't hurt much, it's just a redirect that isn't
referenced anywhere. So these aliases were allowed, preferring not to
bother blog authors with such unimportant stuff.

However, there was a situation where a blog post was merged with
the wrong publication date because of this. Shortly before merging,
the date was updated, but it was accidentally changed in the useless
alias instead of the authoritative `path` key. Because of this
footgun, we don't allow creating these new aliases anymore.

Blog authors who copy-paste old boilerplate will run into failed CI
and have to fix it by removing the alias. It's annoying, but better
than publishing a post with the wrong publication date.

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/master/content/Rust-1.87.0/index.md)

### DIFF
--- a/content/Rust-1.87.0/index.md
+++ b/content/Rust-1.87.0/index.md
@@ -2,10 +2,7 @@
 path = "2025/05/15/Rust-1.87.0"
 title = "Announcing Rust 1.87.0 and ten years of Rust!"
 authors = ["The Rust Release Team"]
-aliases = [
-    "2025/05/15/Rust-1.87.0.html",
-    "releases/1.87.0",
-]
+aliases = ["releases/1.87.0"]
 
 [extra]
 release = true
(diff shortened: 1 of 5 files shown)


senekor:
I was also once [asked](https://github.com/rust-lang/blog.rust-lang.org/pull/1614#issuecomment-2908911217) about the `aliases` key, so it seems to be causing confusion even when things do go well.

PR #1645, opened 2025-06-19
Fix renovate config

senekor:
This is a follow-up to #1640, which didn't quite do what I wanted.


senekor:
@Manishearth gentle ping in case you missed this

PR #1646, opened 2025-06-19
ci: Fix caching of zola build

senekor:
I noticed that CI was taking longer than expected. The problem was that only `./target` is cached by default, but the zola binary is stored in `./zola/target`.


senekor:
Of course there's no cache hit on this PR, but the output of the action looks good to me, `./zola/target` now seems to be included in the paths to cache:

image

PR #1654, opened 2025-06-26
fix the addition instruction in the naked asm example

folkertdev:
turns out LLVM does not have your back and will just let you emit an invalid `add` instruction with 3 operands.


cuviper:
For good measure, here it is verified, and you can "show asm" too:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9c1fd92c8ecdadeb102fc0487e441dff

PR #1662, opened 2025-07-13
Call for Testing: Speeding up compilation with `hint-mostly-unused`

joshtriplett:
This blog post serves as a call for testing on the `hint-mostly-unused`
feature: giving people guidance on how it can help and how to use it, providing
sample performance numbers, asking for specific testing, and pointing people to
a tracking issue for feedback.

[Rendered](https://github.com/joshtriplett/blog.rust-lang.org/blob/call-for-testing-hint-mostly-unused/content/inside-rust/call-for-testing-hint-mostly-unused.md)

### DIFF
--- a/content/inside-rust/call-for-testing-hint-mostly-unused.md
+++ b/content/inside-rust/call-for-testing-hint-mostly-unused.md
@@ -46,6 +46,9 @@ especially for release builds.
 
 ## How does this perform?
 
+Some build timings for clean release builds of a crate depending on various
+specific large API crates:
+
 | **Dependency Crate** | **Before** | **`hint-mostly-unused`** | **Delta** |
 | :- | -: | -: | -: |
 | `windows`, all Graphics/UI features | 18.3s | 10.7s | -42% |
@@ -63,6 +66,9 @@ every crate. Using it for crates whose API surface is mostly used, and/or used
 in multiple different crates or binaries (e.g. multiple test binaries that each
 test a substantial swath of the API), may result in redoing code generation for
 the same items repeatedly.
+
+Also note that this only provides a performance win if you are rebuilding the
+dependency. If you're only rebuilding the top-level crate, this won't help.
 
 ## Plumbing this through Cargo with profiles
 
--- a/content/inside-rust/call-for-testing-hint-mostly-unused.md
+++ b/content/inside-rust/call-for-testing-hint-mostly-unused.md
@@ -7,8 +7,8 @@ authors = ["Josh Triplett"]
 I'm pleased to announce, and call for testing of, the nightly-only `rustc`
 `-Zhint-mostly-unused` option, and the corresponding nightly Cargo features
 `profile.hint-mostly-unused` and `hints.mostly-unused`. These options can help
-accelerate your Rust compile time in some cases, by avoiding compilation of
-items from your dependencies that you aren't using. Your feedback will help
+accelerate your Rust compile time in some cases, by avoiding code generation
+for items from your dependencies that you aren't using. Your feedback will help
 evaluate these features and make progress towards stabilizing them in the
 future.
 
--- a/content/inside-rust/call-for-testing-hint-mostly-unused.md
+++ b/content/inside-rust/call-for-testing-hint-mostly-unused.md
@@ -17,7 +17,7 @@ future.
 Some crates provide comprehensive APIs with a very large surface area, yet many
 of their users need only a few entry points. In such cases, the compiler
 currently spends time generating code for the entire crate, and then ends up
-throwing most of that code away.
+throwing most of that code away as unused.
 
 This can waste a substantial amount of compile time. Some large crates can take
 minutes to compile, and when you use these large crates as dependencies, they
--- a/content/inside-rust/call-for-testing-hint-mostly-unused.md
+++ b/content/inside-rust/call-for-testing-hint-mostly-unused.md
@@ -62,10 +62,14 @@ instance, the `windows` crate in the first example goes from building in 15.1s
 of which 49% is codegen, to building in 7.5s of which 1% is codegen.
 
 Note that this option does not provide a universal performance improvement for
-every crate. Using it for crates whose API surface is mostly used, and/or used
-in multiple different crates or binaries (e.g. multiple test binaries that each
-test a substantial swath of the API), may result in redoing code generation for
-the same items repeatedly.
+every crate; if used when not applicable, this option can make builds much
+*slower*. Deferring compilation of the items in a crate can lead to redoing
+code generation for those items repeatedly. In particular, avoid using this
+hint for crates whose API surface is mostly used, and/or used in multiple
+different crates or binaries (e.g. multiple test binaries that each test a
+substantial swath of the API). Always do performance analysis when considering
+this hint, and only apply it if it applies obvious and substantial wins for
+your users.
 
 Also note that this only provides a performance win if you are rebuilding the
 dependency. If you're only rebuilding the top-level crate, this won't help.
--- a/content/inside-rust/call-for-testing-hint-mostly-unused.md
+++ b/content/inside-rust/call-for-testing-hint-mostly-unused.md
@@ -157,9 +157,30 @@ your crate, with and without this hint set, using nightly Cargo:
 noticeable performance improvement, consider setting it in your published
 crate.
 
+If you're building atop a crate that you only use a small fraction of, you can
+try setting the profile option in your own crate:
+
+```toml
+[profile.dev.package.huge-mostly-unused-dependency]
+hint-mostly-unused = true
+
+[profile.release.package.huge-mostly-unused-dependency]
+hint-mostly-unused = true
+```
+
 Please report any performance improvements, or unexpected performance issues,
 or *especially* any failures you observe, to the [tracking issue for
 profile-hint-mostly-unused](https://github.com/rust-lang/cargo/issues/15644).
+When reporting, please tell us:
+- What hints or profile settings you added
+- What crates you added them to
+- What top-level crate you're building
+- What features you set when building
+- What build profile you're using (e.g. the default dev profile, or the release
+  profile)
+- Whether you did a clean build or an incremental rebuild
+- What performance numbers you got with and without the option you added
+
 We'll take this feedback into account to fix any issues with either the rustc
 compiler feature or the Cargo features, and to evaluate when those features
 have seen enough testing to be ready to stabilize.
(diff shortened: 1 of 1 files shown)


joshtriplett:
cc @rust-lang/cargo

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Deferring code generation with `-Zhint-mostly-unused`
+
+The latest nightly `rustc` compiler now supports an option
+`-Zhint-mostly-unused`, which tells `rustc` that the crate's API surface will
+mostly go unused. This is a hint, and `rustc` doesn't make guarantees about its
+exact behavior (so that we can extend or improve it in the future), but
+currently it causes the compiler to defer as much of code generation as
+possible. (The compiler does this using the same machinery as
+`#[inline(always)]`.)
```
Kobzol: Maybe just clarify that this won't actually inline everything? It sounds a bit like the flag does that. Or don't mention this at all, it's an implementation detail after all.

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,168 @@
...
+
+Applying this option to key crates you depend on (and use only a small subset
+of) can provide a substantial reduction in compile time, for debug builds and
+especially for release builds.
+
+## How does this perform?
+
+| **Dependency Crate** | **Before** | **`hint-mostly-unused`** | **Delta** |
```
Kobzol, line 55: This should tell if it's a debug or a release build, and specify if it's a clean or an incremental build. Maybe we could mention that this probably won't be all that useful for incremental builds?

Doesn't this actually make incremental builds slower, if you have to then repeat the codegen in the leaf crates on every incremental rebuild, instead of it being cached? Maybe some benchmarks regarding that would also be useful in the post.

content/inside-rust/call-for-testing-hint-mostly-unused.md
joshtriplett: Removing it, since as you said it's an implementation detail.

content/inside-rust/call-for-testing-hint-mostly-unused.md
joshtriplett: I'll add notes that it's a clean release build.

I don't expect this to *hurt* incremental builds, in practice. It might depend on how much of the crate's surface area you actually use. But I expect that real-world performance numbers would help there. I'll add some notes on that.

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,194 @@
...
+`-Zhint-mostly-unused` option, and the corresponding nightly Cargo features
+`profile.hint-mostly-unused` and `hints.mostly-unused`. These options can help
+accelerate your Rust compile time in some cases, by avoiding compilation of
+items from your dependencies that you aren't using. Your feedback will help
+evaluate these features and make progress towards stabilizing them in the
+future.
+
+## Background
+
+Some crates provide comprehensive APIs with a very large surface area, yet many
```
saethlin: Perhaps it will help intuition to say this is for APIs that are monomorphic? Not sure how to put that.

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,194 @@
...
+| `x11rb` and `x11rb-protocol` | 5.3s | 2.6s | -51% |
+| `aws-sdk-ec2` | 4m 07s | 2m 04s | -50% |
+
+This performance improvement comes from deferring code generation. For
+instance, the `windows` crate in the first example goes from building in 15.1s
+of which 49% is codegen, to building in 7.5s of which 1% is codegen.
+
+Note that this option does not provide a universal performance improvement for
+every crate. Using it for crates whose API surface is mostly used, and/or used
+in multiple different crates or binaries (e.g. multiple test binaries that each
```
saethlin: I think this explanation really undersells the downside. The problem is that when misused, the flag either completely removes CGU partitioning or causes more items to be compiled in multiple CGUs than before. The way this paragraph is written currently makes it seem like the downsides don't apply if I know have one bin target and no tests.

joshtriplett: This paragraph does also explicitly say it's a bad idea for crates whose API surface is mostly used.

I'll try to make this paragraph clearer and more emphatic.

joshtriplett: @saethlin I've now expanded this paragraph, and made it much more emphatic about the potential downsides. Let me know if that looks better.

content/inside-rust/call-for-testing-hint-mostly-unused.md
joshtriplett: I attempted to cover this; let me know how the new version looks.

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,207 @@
...
+`profile.hint-mostly-unused` and `hints.mostly-unused`. These options can help
+accelerate your Rust compile time in some cases, by avoiding compilation of
+items from your dependencies that you aren't using. Your feedback will help
+evaluate these features and make progress towards stabilizing them in the
+future.
+
+## Background
+
+When building a Rust library crate, the compiler generates compiled code for as
+much of the crate as it can (everything that isn't generic and isn't inlined),
```
Noratrieb:
```suggestion
much of the crate as it can (everything that isn't generic and doesn't have `#[inline]`),
```

We should avoid using the adjective "inlined" for functions that have the inline attribute (same about using the verb inline for adding the attribute), as that adjective is more commonly used for functions that *actually have been inlined by the optimizer*. Let's avoid this confusion and be very explicit about the attribute.

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,207 @@
++++
+path = "inside-rust/2025/07/14/call-for-testing-hint-mostly-unused"
+title = "Call for Testing: Speeding up compilation with `hint-mostly-unused`"
+authors = ["Josh Triplett"]
++++
+
+I'm pleased to announce, and call for testing of, the nightly-only `rustc`
+`-Zhint-mostly-unused` option, and the corresponding nightly Cargo features
+`profile.hint-mostly-unused` and `hints.mostly-unused`. These options can help
+accelerate your Rust compile time in some cases, by avoiding compilation of
```
saethlin: "by avoiding compilation" may be interpreted as saying we will skip type-checking and/or borrow-checking. Which is not true in the current implementation. But maybe we want to leave open the possibility of doing so in the future?

I don't have a particular suggestion or concern here, I just want to make sure you mean the implication I'm reading.

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,207 @@
...
+which gets linked into later crates into the dependency graph. However, some
+crates provide comprehensive APIs with a very large surface area, yet many of
+their users need only a few entry points. In such cases, the compiler currently
+spends time generating code for the entire crate, and the linker then ends up
+throwing most of that code away as unused.
+
+This can waste a substantial amount of compile time. Some large crates can take
+minutes to compile, and when you use these large crates as dependencies, they
+can take a disproportionate amount of the entire compilation time of your
+top-level crate.
```
saethlin:
```suggestion
can take a disproportionate fraction of the entire compilation time for a full build.
```
Because adding `-Zhint-mostly-unused` will increase the compile time of the top-level crate, while lowering the compile time of the entire multi-crate build.

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,207 @@
...
+
+This performance improvement comes from deferring code generation. For
+instance, the `windows` crate in the first example goes from building in 15.1s
+of which 49% is codegen, to building in 7.5s of which 1% is codegen.
+
+Note that this option does not provide a universal performance improvement for
+every crate; if used when not applicable, this option can make builds much
+*slower*. Deferring compilation of the items in a crate can lead to redoing
+code generation for those items repeatedly. In particular, avoid using this
+hint for crates whose API surface is mostly used, and/or used in multiple
```
saethlin:
```suggestion
code generation for those items repeatedly. In particular, this hint will
probably regress compile time if applied to crates whose API surface is
mostly used, and/or used in multiple
```

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,207 @@
...
+different crates or binaries (e.g. multiple test binaries that each test a
+substantial swath of the API).
+
+Always do performance analysis when considering this hint, and only apply it if
+it applies obvious and substantial wins for your users. Never apply it across
+the board to all your dependencies.
```
saethlin:
```suggestion
it applies obvious and substantial wins for your users. Applying it across
the board to all your dependencies will probably go poorly.
```

content/inside-rust/call-for-testing-hint-mostly-unused.md
```diff
@@ -0,0 +1,207 @@
...
+
+If most of the items in your crate are polymorphic (generic), this hint may be
+redundant, as Rust already defers compilation of polymorphic items until they
+get monomorphized with specific types.
+
+Also note that this only provides a performance win if you are rebuilding the
```
CleanCut: Is this true that it only helps when you are _rebuilding_ the dependency, or does it help whenever you are _building_ the dependency (ie even the first time)?

content/inside-rust/call-for-testing-hint-mostly-unused.md
joshtriplett: @Noratrieb I've added a slight variation on this suggestion: "isn't marked as `#[inline]`" rather than "doesn't have `#[inline]`".

content/inside-rust/call-for-testing-hint-mostly-unused.md
joshtriplett: I definitely didn't intend that interpretation, no. I've rephrased accordingly.

content/inside-rust/call-for-testing-hint-mostly-unused.md
joshtriplett: Fixed.

content/inside-rust/call-for-testing-hint-mostly-unused.md
joshtriplett: Good catch.

ehuss (approved):
Thanks!

PR #1666, opened 2025-07-24
Announce Rust 1.89.0

BoxyUwU:
- Explicitly Inferred Arguments to Const Generics
- Mismatched Lifetime Syntax Lint
- More x86 Target Features and Intrinsics
- Cross compiled doctests
- `i128` and `u128` in `extern "C"`
- in progress demotion of `x86_64-macos-darwin`
- spec compliant wasm32-unknown-unknown C ABI

cc @rust-lang/release 

[Rendered](https://github.com/BoxyUwU/blog.rust-lang.org/blob/1.89.0/content/Rust-1.89.0.md)

### DIFF
--- a/content/Rust-1.89.0.md
+++ b/content/Rust-1.89.0.md
@@ -162,7 +162,7 @@ In accordance with these changes, the Rust project is in the [process of demotin
 
 We expect that the RFC for the demotion to Tier 2 with host tools will be accepted between the releases of Rust 1.89 and 1.90, which means that Rust 1.89 will be the last release of Rust where `x86_64-apple-darwin` is a Tier 1 target.
 
-For users, this change will not immediately cause impact. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods while the target remains at Tier 2.
+For users, this change will not immediately cause impact. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods while the target remains at Tier 2. Over time, it's likely that reduced test coverage for this target will cause things to break or fall out of compatibility with no further announcements.
 
 [apple]: https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon#Timeline
 [gha-sunset]: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down
--- a/content/Rust-1.89.0.md
+++ b/content/Rust-1.89.0.md
@@ -136,7 +136,7 @@ pub fn cool_simd_code(/* .. */) -> /* ... */ {
 
 Doctests will now be tested when running `cargo test --doc --target other_target`, this may result in some amount of breakage due to would-be-failing doctests now being tested.
 
-Failing tests can be disabled by annotating the doctest with `ignore-<target>`:
+Failing tests can be disabled by annotating the doctest with `ignore-<target>` ([docs](https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#ignoring-targets)):
 ```rust
 /// ```ignore-x86_64
 /// panic!("something")
@@ -156,13 +156,13 @@ This is the last bit of follow up to the layout changes from last year: https://
 
 ### Demoting `x86_64-apple-darwin` to Tier 2 with host tools
 
-GitHub will soon [discontinue][gha-sunset] providing free macOS x86\_64 runners for public repositories. Apple has also announced their [plans][apple] for discontinuing support for the x86\_86 architecture.
+GitHub will soon [discontinue][gha-sunset] providing free macOS x86\_64 runners for public repositories. Apple has also announced their [plans][apple] for discontinuing support for the x86\_64 architecture.
 
-In accordance with these changes, the Rust project is in the [process of demoting the `x86_64-apple-darwin` target][rfc] from Tier 1 to Tier 2 with host tools. This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.
+In accordance with these changes, the Rust project is in the [process of demoting the `x86_64-apple-darwin` target][rfc] from [Tier 1 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-1-with-host-tools) to [Tier 2 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-2-with-host-tools). This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.
 
 We expect that the RFC for the demotion to Tier 2 with host tools will be accepted between the releases of Rust 1.89 and 1.90, which means that Rust 1.89 will be the last release of Rust where `x86_64-apple-darwin` is a Tier 1 target.
 
-For users, nothing will change. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods.
+For users, this change will not immediately cause impact. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods while the target remains at Tier 2.
 
 [apple]: https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon#Timeline
 [gha-sunset]: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down
--- a/content/Rust-1.89.0.md
+++ b/content/Rust-1.89.0.md
@@ -1,0 +1,218 @@
++++
+path = "2025/08/07/Rust-1.89.0"
+title = "Announcing Rust 1.89.0"
+authors = ["The Rust Release Team"]
+aliases = ["releases/1.89.0"]
+
+[extra]
+release = true
++++
+
+The Rust team is happy to announce a new version of Rust, 1.89.0. Rust is a programming language empowering everyone to build reliable and efficient software.
+
+If you have a previous version of Rust installed via `rustup`, you can get 1.89.0 with:
+
+```console
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.89.0](https://doc.rust-lang.org/stable/releases.html#version-1890-2025-08-07).
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.89.0 stable
+
+### Explicitly inferred arguments to const generics
+
+Rust now supports `_` as an argument to const generic parameters, inferring the value from surrounding context:
+
+```rust
+pub fn all_false<const LEN: usize>() -> [bool; LEN] {
+  [false; _]
+}
+```
+
+Similar to the rules for when `_` is permitted as a type, `_` is not permitted as an argument to const generics when in a signature:
+
+```rust
+// This is not allowed
+pub const fn all_false<const LEN: usize>() -> [bool; _] {
+  [false; LEN]
+}
+
+// Neither is this
+pub const ALL_FALSE: [bool; _] = all_false::<10>();
+```
+
+### Mismatched lifetime syntaxes lint
+
+[Lifetime elision][elision] in function signatures is an ergonomic aspect of the Rust language, but it can also be a stumbling point for newcomers and experts alike. This is especially true when lifetimes are inferred in types where it isn't syntactically obvious that a lifetime is even present:
+
+```rust
+// The returned type `std::slice::Iter` has a lifetime, 
+// but there's no visual indication of that.
+//
+// Lifetime elision infers the lifetime of the return 
+// type to be the same as that of `scores`.
+fn items(scores: &[u8]) -> std::slice::Iter<u8> {
+   scores.iter()
+}
+```
+
+Code like this will now produce a warning by default:
+
+```text
+warning: hiding a lifetime that's elided elsewhere is confusing
+ --> src/lib.rs:1:18
+  |
+1 | fn items(scores: &[u8]) -> std::slice::Iter<u8> {
+  |                  ^^^^^     -------------------- the same lifetime is hidden here
+  |                  |
+  |                  the lifetime is elided here
+  |
+  = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
+  = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
+help: use `'_` for type paths
+  |
+1 | fn items(scores: &[u8]) -> std::slice::Iter<'_, u8> {
+  |                                             +++
+```
+
+We [first attempted][elided_lifetime_in_path] to improve this situation back in 2018 as part of the [`rust_2018_idioms`][2018-by-default] lint group, but [strong feedback][bevy] about the `elided_lifetimes_in_paths` lint showed that it was too blunt of a hammer as it warns about lifetimes which don't matter to understand the function:
+
+```rust
+use std::fmt;
+
+struct Greeting;
+
+impl fmt::Display for Greeting {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        //                -----^^^^^^^^^ expected lifetime parameter
+        // Knowing that `Formatter` has a lifetime does not help the programmer
+        "howdy".fmt(f)
+    }
+}
+```
+
+We then realized that the confusion we want to eliminate occurs when both
+
+1. lifetime elision inference rules *connect* an input lifetime to an output lifetime
+2. it's not syntactically obvious that a lifetime exists 
+
+There are two pieces of Rust syntax that indicate that a lifetime exists: `&` and `'`, with `'` being subdivided into the inferred lifetime `'_` and named lifetimes `'a`. When a type uses a named lifetime, lifetime elision will not infer a lifetime for that type. Using these criteria, we can construct three groups:
+
+| Self-evident it has a lifetime | Allow lifetime elision to infer a lifetime | Examples                              |
+|--------------------------------|--------------------------------------------|---------------------------------------|
+| No                             | Yes                                        | `ContainsLifetime`                    |
+| Yes                            | Yes                                        | `&T`, `&'_ T`, `ContainsLifetime<'_>` |
+| Yes                            | No                                         | `&'a T`, `ContainsLifetime<'a>`       |
+
+The `mismatched_lifetime_syntaxes` lint checks that the inputs and outputs of a function belong to the same group. For the initial motivating example above, `&[u8]` falls into the second group while `std::slice::Iter<u8>` falls into the first group. We say that the lifetimes in the first group are *hidden*. 
+
+Because the input and output lifetimes belong to different groups, the lint will warn about this function, reducing confusion about when a value has a meaningful lifetime that isn't visually obvious.
+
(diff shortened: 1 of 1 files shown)


content/Rust-1.89.0.md
```diff
@@ -0,0 +1,78 @@
...
+
+The `target_feature` attribute now supports the `sha512`, `sm3`, `sm4`, `kl` and `widekl` target features on x86. Additionally a number of `avx512` intrinsics and target features are also supported on x86:
+
+```rust
+#[target_feature(enable = "avx512bw")]
+pub fn cool_simd_code(/* .. */) -> /* ... */ {
+    /* ... */
+}
+
+```
```
BoxyUwU, line 133: cc @Amanieu @sayantn :3 I don't know much about this

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,78 @@
...
+
+Similar to the rules for when `_` is permitted as a type, `_` is not permitted as an argument to const generics when in a signature:
+
+```rust
+// This is not allowed
+pub fn make_bitset<const LEN: usize>() -> [bool; _] {
+  [false; LEN]
+}
+```
```
BoxyUwU: cc @rust-lang/project-const-generics

content/Rust-1.89.0.md
sayantn: Also specify that all avx512 intrinsics in `std::arch::{x86,x86_64}` are stable now (same for the other target features).

Edit: nvm I didn't see it right

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,78 @@
...
+The `target_feature` attribute now supports the `sha512`, `sm3`, `sm4`, `kl` and `widekl` target features on x86. Additionally a number of `avx512` intrinsics and target features are also supported on x86:
+
+```rust
+#[target_feature(enable = "avx512bw")]
+pub fn cool_simd_code(/* .. */) -> /* ... */ {
+    /* ... */
+}
+
+```
+
```
shepmaster, line 134: We're playing a little fast and loose with the timing of this announcement, but the "worst" case is that we extend support one more release, which people are unlikely to be unhappy about. 

```suggestion
### Demoting `x86_64-apple-darwin` to Tier 2 with host tools 

GitHub will soon [discontinue][gha-sunset] providing free macOS x86\_64 runners for public repositories. Apple has also announced their [plans][apple] for discontinuing support for the x86\_86 architecture.

In accordance with these changes, the Rust project is in the [process of demoting the `x86_64-apple-darwin` target][rfc] from Tier 1 to Tier 2 with host tools. This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.

We expect that the RFC for the demotion to Tier 2 with host tools will be accepted between the releases of Rust 1.89 and 1.90, which means that Rust 1.89 will be the last release of Rust where x86\_64 macOS is a Tier 1 target.

For users, nothing will change. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods.

[apple]: https://en.wikipedia.org/wiki/Mac_transition_to_Apple_silicon#Timeline
[gha-sunset]: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down
[rfc]: https://github.com/rust-lang/rfcs/pull/3841
```

jieyouxu: Maybe s/x86\_64 macOS/`x86_64-apple-darwin`/.

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,78 @@
...
+```rust
+// This is not allowed
+pub fn make_bitset<const LEN: usize>() -> [bool; _] {
+  [false; LEN]
+}
+```
+
+### Mismatched lifetimes syntax lint
+
+TC/LANG to write this
```
shepmaster: ### Mismatched lifetime syntaxes lint

[Lifetime elision][elision] in function signatures is an ergonomic aspect of the Rust language, but it can also be a stumbling point for newcomers and experts alike. This is especially true when lifetimes are inferred in types where it isn't syntactically obvious that a lifetime is even present:

```rust
// The returned type `std::slice::Iter` has a lifetime, 
// but there's no visual indication of that.
//
// Lifetime elision infers the lifetime of the return 
// value to be the same as the argument `scores`.
fn items(scores: &[u8]) -> std::slice::Iter<u8> {
   scores.iter()
}
```

Code like this will now produce a warning by default:

```text
warning: hiding a lifetime that's elided elsewhere is confusing
 --> src/lib.rs:1:18
  |
1 | fn items(scores: &[u8]) -> std::slice::Iter<u8> {
  |                  ^^^^^     -------------------- the same lifetime is hidden here
  |                  |
  |                  the lifetime is elided here
  |
  = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
  = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: use `'_` for type paths
  |
1 | fn items(scores: &[u8]) -> std::slice::Iter<'_, u8> {
  |                                             +++
```

We [first attempted][elided_lifetime_in_path] to improve this situation back in 2018 as part of the [`rust_2018_idioms`][2018-by-default] lint group, but [strong feedback][bevy] about the `elided_lifetimes_in_paths` lint showed that it was too blunt of a hammer as it warns about lifetimes which don't matter to understand the function:

```rust
use std::fmt;

struct Greeting;

impl fmt::Display for Greeting {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        //                -----^^^^^^^^^ expected lifetime parameter
        // Knowing that `Formatter` has a lifetime does not help the programmer
        "howdy".fmt(f)
    }
}
```

We then realized that the confusion we want to eliminate occurs when both

1. lifetime elision inference rules *connect* an input lifetime to an output lifetime
2. it's not syntactically obvious that a lifetime exists 

There are two pieces of Rust syntax that indicate that a lifetime exists: `&` and `'`, with `'` being subdivided into the inferred lifetime `'_` and named lifetimes `'a`. When a type uses a named lifetime, lifetime elision will not infer a lifetime for that type. Using these criteria, we can construct three groups:

| Self-evident it has a lifetime | Allow lifetime elision to infer a lifetime | Examples                              |
|--------------------------------|--------------------------------------------|---------------------------------------|
| No                             | Yes                                        | `ContainsLifetime`                    |
| Yes                            | Yes                                        | `&T`, `&'_ T`, `ContainsLifetime<'_>` |
| Yes                            | No                                         | `&'a T`, `ContainsLifetime<'a>`       |

The `mismatched_lifetime_syntaxes` lint checks that the inputs and outputs of a function belong to the same group. For the initial motivating example above, `&[u8]` falls into the second group while `std::slice::Iter<u8>` falls into the first group. We say that the lifetimes in the first group are *hidden*. 

Because the input and output lifetimes belong to different groups, the lint will warn about this function, reducing confusion about when a value has a meaningful lifetime that isn't visually obvious.

The `mismatched_lifetime_syntaxes` lint supersedes the `elided_named_lifetimes` lint, which did something similar for named lifetimes specifically.

Future work on the `elided_lifetimes_in_paths` lint intends to split it into more focused sub-lints with an eye to warning about a subset of them eventually.

[elision]: https://doc.rust-lang.org/1.89/book/ch10-03-lifetime-syntax.html#lifetime-elision
[elided_lifetime_in_path]: https://github.com/rust-lang/rust/pull/46254
[2018-by-default]: https://github.com/rust-lang/rust/issues/54910
[bevy]: https://github.com/rust-lang/rust/issues/131725

traviscross: That's great.  Thanks for writing that up.

One thing that I would like us to get in there somewhere is that we consider the lifetime in `ContainsLifetime` to be hidden.  We don't need to get into whether that's a subtype of elided or a separate category or whatnot -- that's not important.  But that when it's not self-evident (as you so well put it) that a type path has a lifetime, that the lifetime is "hidden" is nomenclature we adopted deliberately and should introduce in the narrative.

shepmaster: > we consider the lifetime in `ContainsLifetime` to be hidden

I took a stab at that.

Mark-Simulacrum: nit: the compiler output above seems to have lost `  = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default` -- I think we should keep that, to make it very clear this is a lint that can be disabled if the user wants to.

shepmaster: Yeah, I took that out because it comes and goes (is it the first warning, is it set on the command line, etc.) I can add it back if you think it’s useful.

Mark-Simulacrum: I think it's worth keeping, yes.

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Explicitly inferred arguments to const generics
+
+Rust now supports `_` as an argument to const generic parameters, inferring the value from surrounding context:
+
+```rust
+pub fn make_bitset<const LEN: usize>() -> [bool; LEN] {
+  [false; _]
+}
+```
```
nik-rev: It could be helpful to clarify that `const` items' lengths cannot be inferred:

```rs
// this won't work
const A: [u8; _] = [1, 2, 3, 4];
```

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,78 @@
...
+
+Similar to the rules for when `_` is permitted as a type, `_` is not permitted as an argument to const generics when in a signature:
+
+```rust
+// This is not allowed
+pub fn make_bitset<const LEN: usize>() -> [bool; _] {
+  [false; LEN]
+}
```
BoxyUwU:
```suggestion
// This is not allowed
pub const fn make_bitset<const LEN: usize>() -> [bool; _] {
  [false; LEN]
}
// Neither is this
pub const MY_BITSET: [bool; _] = make_bitset::<10>();
```

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,176 @@
...
+
+```rust
+#[target_feature(enable = "avx512bw")]
+pub fn cool_simd_code(/* .. */) -> /* ... */ {
+    /* ... */
+}
+
+```
+
+### Platform Support
```
bjorn3, line 175: https://github.com/rust-lang/blog.rust-lang.org/pull/1666#issuecomment-3129049497

content/Rust-1.89.0.md
```diff
@@ -45,7 +45,81 @@ pub const MY_BITSET: [bool; _] = make_bitset::<10>();
 
 ### Mismatched lifetimes syntax lint
 
-TC/LANG to write this
+### Mismatched lifetime syntaxes lint
```
lcnr: duplicate header

shepmaster: And “lifetime syntaxes” matches the name of the lint

BoxyUwU: already fixed :3

content/Rust-1.89.0.md
```diff
@@ -45,7 +45,81 @@ pub const MY_BITSET: [bool; _] = make_bitset::<10>();
...
+
+[Lifetime elision][elision] in function signatures is an ergonomic aspect of the Rust language, but it can also be a stumbling point for newcomers and experts alike. This is especially true when lifetimes are inferred in types where it isn't syntactically obvious that a lifetime is even present:
+
+```rust
+// The returned type `std::slice::Iter` has a lifetime, 
+// but there's no visual indication of that.
+//
+// Lifetime elision infers the lifetime of the return 
+// value to be the same as the argument `scores`.
```
traviscross:
```suggestion
// type to be the same as that of `scores`.
```

(Nits.)

The paragraph above refers to return types rather than values, so it seems better to keep with that.  Also, it seems better to reword so as to avoid the argument/parameter distinction.

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,176 @@
...
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.89.0 stable
+
+### Explicitly inferred arguments to const generics
+
+Rust now supports `_` as an argument to const generic parameters, inferring the value from surrounding context:
+
+```rust
+pub fn make_bitset<const LEN: usize>() -> [bool; LEN] {
```
kpreid: The term “bitset” implies one element stored per bit. This naming could incidentally mislead users into thinking that e.g. `size_of::<[bool; 8]>() == 1`. I suggest changing it.

As a replacement name, maybe something like `all_false`?

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,190 @@
...
+### Cross-compiled doctests
+
+Doctests will now be tested when running `cargo test --doc --target other_target`, this may result in some amount of breakage due to would-be-failing doctests now being tested.
+
+Failing tests can be disabled by annotating the doctest with `ignore-<target>`:
+```rust
+/// ```ignore-x86_64
+/// panic!("something")
+/// ```
+/// pub fn my_function() { }
```
BoxyUwU:
```suggestion
pub fn my_function() { }
```

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,190 @@
...
+
+Doctests will now be tested when running `cargo test --doc --target other_target`, this may result in some amount of breakage due to would-be-failing doctests now being tested.
+
+Failing tests can be disabled by annotating the doctest with `ignore-<target>`:
+```rust
+/// ```ignore-x86_64
+/// panic!("something")
+/// ```
+pub fn my_function() { }
+```
```
BoxyUwU: cc @rust-lang/cargo

content/Rust-1.89.0.md
tgross35, line 1: This doesn't seem to have picked up https://github.com/rust-lang/rust/issues/141206, was that intentional?

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,218 @@
...
+    /* ... */
+}
+
+```
+
+### Cross-compiled doctests
+
+Doctests will now be tested when running `cargo test --doc --target other_target`, this may result in some amount of breakage due to would-be-failing doctests now being tested.
+
+Failing tests can be disabled by annotating the doctest with `ignore-<target>`:
```
Mark-Simulacrum:
```suggestion
Failing tests can be disabled by annotating the doctest with `ignore-<target>` ([docs](https://doc.rust-lang.org/stable/rustdoc/write-documentation/documentation-tests.html#ignoring-targets)):
```

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,218 @@
...
+
+* The Rust types are ABI- and layout-compatible with (unsigned) `__int128` in C when the type is available.
+* On platforms where `__int128` is not available, `i128` and `u128` do not necessarily align with any C type.
+* `i128` is _not_ necessarily compatible with `_BitInt(128)` on any platform, because `_BitInt(128)` and `__int128` may not have the same ABI (as is the case on x86-64).
+
+This is the last bit of follow up to the layout changes from last year: https://blog.rust-lang.org/2024/03/30/i128-layout-update/.
+
+### Demoting `x86_64-apple-darwin` to Tier 2 with host tools
+
+GitHub will soon [discontinue][gha-sunset] providing free macOS x86\_64 runners for public repositories. Apple has also announced their [plans][apple] for discontinuing support for the x86\_86 architecture.
```
Mark-Simulacrum:
```suggestion
GitHub will soon [discontinue][gha-sunset] providing free macOS x86\_64 runners for public repositories. Apple has also announced their [plans][apple] for discontinuing support for the x86\_64 architecture.
```

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,218 @@
...
+
+In accordance with these changes, the Rust project is in the [process of demoting the `x86_64-apple-darwin` target][rfc] from Tier 1 to Tier 2 with host tools. This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.
```
Mark-Simulacrum:
```suggestion
In accordance with these changes, the Rust project is in the [process of demoting the `x86_64-apple-darwin` target][rfc] from [Tier 1 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-1-with-host-tools) to [Tier 2 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-2-with-host-tools). This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.
```

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,218 @@
...
+
+We expect that the RFC for the demotion to Tier 2 with host tools will be accepted between the releases of Rust 1.89 and 1.90, which means that Rust 1.89 will be the last release of Rust where `x86_64-apple-darwin` is a Tier 1 target.
+
+For users, nothing will change. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods.
```
Mark-Simulacrum:
```suggestion
For users, this change will not immediately cause impact. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods while the target remains at Tier 2.
```

content/Rust-1.89.0.md
```diff
@@ -0,0 +1,218 @@
...
+
+In accordance with these changes, the Rust project is in the [process of demoting the `x86_64-apple-darwin` target][rfc] from [Tier 1 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-1-with-host-tools) to [Tier 2 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-2-with-host-tools). This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.
+
+We expect that the RFC for the demotion to Tier 2 with host tools will be accepted between the releases of Rust 1.89 and 1.90, which means that Rust 1.89 will be the last release of Rust where `x86_64-apple-darwin` is a Tier 1 target.
+
+For users, this change will not immediately cause impact. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods while the target remains at Tier 2.
```
shepmaster: @traviscross I don't agree with this specific wording as now it sounds like *this change will eventually cause impact*, which is not the case to my knowledge. What specific event or impact are you trying to suggest at?

traviscross: (I didn't write that text.)

That said, the change will eventually have effect, at least probabilistically, due to our lower standards and expectations for how changes and decisions affect a Tier 2 target.

Mark-Simulacrum: I think it's me you meant to ping? This will absolutely cause impact in the medium term as we're likely to steadily lose test coverage (by accident or intentionally) as things break and fall out of compatibility with no further announcements. In the longer term, it seems likely we'd drop to tier 3, at least whenever Apple makes it too hard to cross compile for x86_64 macOS.

In the short term, I think framing this as not impacting users is misleading. Users are losing a guarantee *now*, after all.

shepmaster: > (I didn't write that text.)

> I think it's me you meant to ping?

Ah yes, *mea culpa*, apologies.

The intent of my original wording is that people don't have to anything different with respect to getting the target (e.g. interactions with `rustup`). 

I do not think *this* post should imply or assume anything about tier 3, because that's not the change occurring. If there's a demotion to tier 3, we already have an obligation to report that in similar ways (release notes, a blog post), so we don't need to address that possibility here.

If we are worried about communicating reduced test coverage, we should be explicit about that. The current wording is too nebulous and will easily lead to panicked people asking "what do you mean?". 

```suggestion
For users, this change will not immediately cause impact. Builds of both the standard library and the compiler will still be distributed by the Rust Project for use via `rustup` or alternative installation methods while the target remains at Tier 2. Over time, it's likely that reduced test coverage for this target will cause things to break or fall out of compatibility with no further announcements.
```

alex-semenyuk (approved):

pietroalbini (approved):

PR #1672, opened 2025-08-07
Fix submodule path

Kobzol:
It was containing a commit from senekor's fork, but it was pointing to the upstream repo.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "zola"]
 	path = zola
-	url = https://github.com/getzola/zola
+	url = https://github.com/senekor/zola.git


senekor:
That's not what we want though, we don't want to use my fork indefinitely. Upstream should work for us now.

Kobzol:
In general, checking out the recursive submodules and building Zola is... interesting. I would perhaps point people to https://www.getzola.org/documentation/getting-started/installation/ or even better to https://github.com/getzola/zola/releases/tag/v0.21.0, it IMO seems easier than building a 500+ dependency binary in release mode with LTO and CGU=1 and fetching tens of submodules.

senekor:
IIRC @Manishearth is against telling people to install random binaries we don't control.

senekor:
I definitely agree with the fact that it's annoying. Not sure what the best solution is.

Issue #1687, opened 2025-09-03
Change default branch to `main`?


carols10cents:
The rust-lang org has some legacy repos where we've dropped the ball on renaming the default branch from `master` to `main`; it's past time that we finish up.

But the good news is in the meantime [GitHub has implemented more support for renaming branches like redirects](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch), so we don't have to worry about breaking links to files on the master branch.

[Searching for `master` in this repo](https://github.com/search?q=repo%3Arust-lang%2Fblog.rust-lang.org%20master&type=code), I mostly see links to other repos, but spots that should be renamed here include:

- [README.md](https://github.com/rust-lang/blog.rust-lang.org/blob/0f8a6b40d5e171d501b21dc22e98e55aa6c2b45a/README.md?plain=1#L60)
- github workflows
  - [main](https://github.com/rust-lang/blog.rust-lang.org/blob/0f8a6b40d5e171d501b21dc22e98e55aa6c2b45a/.github/workflows/main.yml#L5)
  - [snapshot tests](https://github.com/rust-lang/blog.rust-lang.org/blob/0f8a6b40d5e171d501b21dc22e98e55aa6c2b45a/.github/workflows/snapshot_tests.yml#L21)
- [footer.html](https://github.com/rust-lang/blog.rust-lang.org/blob/0f8a6b40d5e171d501b21dc22e98e55aa6c2b45a/templates/footer.html#L41)

Is there anything else that would be affected?

senekor:
I can't think of anything that could cause problems. Let's do it 🙂

PR #1689, opened 2025-09-03
Change branch name from master to main

carols10cents:
Connects to #1687.

This will need someone with the right permissions to actually do the branch rename :)

Fixes: https://github.com/rust-lang/blog.rust-lang.org/issues/1687

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
 
 env:
@@ -44,7 +44,7 @@ jobs:
 
   pub_date:
     name: Check publication date for placeholder
-    if: ${{ github.ref == 'refs/heads/master' }}
+    if: ${{ github.ref == 'refs/heads/main' }}
 
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


Kobzol:
Yes, this pretty much has to be done by an infra admin. A PR also has to be sent to the team repo to update branch protections.

senekor:
Maybe it's possible to protect both master _and_ main in the pattern? Then the team repo PR can be merged first, so there is no "downtime" for the branch protection.

marcoieni:
I also don't think it's needed to protect both main and master since I will perform the steps quickly 👍
I will execute the plan once https://github.com/rust-lang/team/pull/1976 is approved. I'm pinging someone

Kobzol:
Note that promote-release might require an update for this (https://github.com/rust-lang/promote-release/blob/ee6866331f1d61fc9faa5df00bc0305eb7a2ad65/src/main.rs#L37).

Kobzol:
(My PR doesn't fix this, it only removes master references for r-l/r, not the blog.)

marcoieni:
oh, I see, so we need to open another PR to fix it 👍

Kobzol:
https://github.com/rust-lang/promote-release/pull/96 should unblock this.

marcoieni (approved):

PR #1690, opened 2025-09-06
Make build of the blog much faster by only fetching section metadata

Kobzol:
I was writing a blog and I couldn't understand why the build was so slow. I tracked it down to tera copying large amounts of JSON dictionaries, which I then tracked down to https://github.com/getzola/zola/issues/2508. This flag reduces the build time of the blog locally from ~7-8s to ~1.2s.

I compared the outputs before/after and they seem to be the same.


senekor (approved):
Seems fine. You could snapshot-test the generated output by putting `RUN_SNAPSHOT_TESTS` in the PR description and pushing against. (maybe update committer timestamp or something and force-push)

Kobzol:
Looks like the snapshot tests are green.

PR #1691, opened 2025-09-07
Add Rust 1.90.0 announcement

Mark-Simulacrum:
cc @rust-lang/release 

[Rendered](https://github.com/Mark-Simulacrum/blog.rust-lang.org/blob/release-blog/content/Rust-1.90.0.md)

### DIFF
--- a/content/Rust-1.90.0.md
+++ b/content/Rust-1.90.0.md
@@ -22,15 +22,19 @@ If you'd like to help us out by testing future releases, you might consider upda
 
 ## What's in 1.90.0 stable
 
-### `lld` by default on `x86_64-unknown-linux-gnu`
+# LLD is now the default linker on `x86_64-unknown-linux-gnu`
 
-As [previously announced](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/),
-1.90 brings a change in the default linker to `lld` on
-`x86_64-unknown-linux-gnu`. As discussed in that blog, for the vast majority of
-users this change should just be a compile time reduction. In some edge cases,
-`lld` may have different behavior: please file
-[issues](https://github.com/rust-lang/rust/issues/new/choose) if you run into
-problems as part of the migration.
+The `x86_64-unknown-linux-gnu` target will now use the LLD linker for linking Rust crates by default. This should result in improved linking performance vs the default Linux linker (BFD), particularly for large binaries, binaries with a lot of debug information, and for incremental rebuilds.
+
+In the vast majority of cases, LLD should be backwards compatible with BFD, and you should not see any difference other than reduced compilation time. However, if you do run into any new linker issues, you can always opt out using the `-C linker-features=-lld` compiler flag. Either by adding it to the usual `RUSTFLAGS` environment variable, or to a project's [`.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html) configuration file,
+like so:
+
+```toml
+[target.x86_64-unknown-linux-gnu]
+rustflags = ["-Clinker-features=-lld"]
+```
+
+If you encounter any issues with the LLD linker, please [let us know](https://github.com/rust-lang/rust/issues/new/choose). You can read more about the switch to LLD, some benchmark numbers and the opt out mechanism [here](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/).
 
 ### Cargo adds native support for workspace publishing
 
--- a/content/Rust-1.90.0.md
+++ b/content/Rust-1.90.0.md
@@ -1,0 +1,103 @@
++++
+path = "2025/09/18/Rust-1.90.0"
+title = "Announcing Rust 1.90.0"
+authors = ["The Rust Release Team"]
+aliases = ["releases/1.90.0"]
+
+[extra]
+release = true
++++
+
+The Rust team is happy to announce a new version of Rust, 1.90.0. Rust is a programming language empowering everyone to build reliable and efficient software.
+
+If you have a previous version of Rust installed via `rustup`, you can get 1.90.0 with:
+
+```console
+$ rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.90.0](https://doc.rust-lang.org/stable/releases.html#version-1900-2025-09-18).
+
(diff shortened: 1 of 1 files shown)


content/Rust-1.90.0.md
```diff
@@ -0,0 +1,103 @@
...
+
+If you don't have it already, you can [get `rustup`](https://www.rust-lang.org/install.html) from the appropriate page on our website, and check out the [detailed release notes for 1.90.0](https://doc.rust-lang.org/stable/releases.html#version-1900-2025-09-18).
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.90.0 stable
+
+### `lld` by default on `x86_64-unknown-linux-gnu`
+
+As [previously announced](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/),
```
Kobzol: We prepared some text for this in https://github.com/rust-lang/rust/issues/143115. Maybe we could at least mention the opt out, and the scale of the expected wins?

Mark-Simulacrum: Happy to take that text - we're not good at finding when people do actually write blurbs yet. Feel free to push or suggest it, otherwise I'll try to copy it in the next few days.

content/Rust-1.90.0.md
```diff
@@ -0,0 +1,103 @@
...
+1.90 brings a change in the default linker to `lld` on
+`x86_64-unknown-linux-gnu`. As discussed in that blog, for the vast majority of
+users this change should just be a compile time reduction. In some edge cases,
+`lld` may have different behavior: please file
+[issues](https://github.com/rust-lang/rust/issues/new/choose) if you run into
+problems as part of the migration.
```
Kobzol:
````suggestion
# LLD is now the default linker on `x86_64-unknown-linux-gnu`

The `x86_64-unknown-linux-gnu` target will now use the LLD linker for linking Rust crates by default. This should result in improved linking performance vs the default Linux linker (BFD), particularly for large binaries, binaries with a lot of debug information, and for incremental rebuilds.

In the vast majority of cases, LLD should be backwards compatible with BFD, and you should not see any difference other than reduced compilation time. However, if you do run into any new linker issues, you can always opt out using the `-C linker-features=-lld` compiler flag. Either by adding it to the usual `RUSTFLAGS` environment variable, or to a project's [`.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html) configuration file,
like so:

```toml
[target.x86_64-unknown-linux-gnu]
rustflags = ["-Clinker-features=-lld"]
```

If you encounter any issues with the LLD linker, please [let us know](https://github.com/rust-lang/rust/issues/new/choose). You can read more about the switch to LLD, some benchmark numbers and the opt out mechanism [here](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/).
````

Comment (Kobzol) [reply content/Rust-1.90.0.md]: Uhh, using triple backticks in suggestions doesn't seem to work well. I replaced the backticks in the suggestion with [BACKTICKS].

Comment (bjorn3) [reply content/Rust-1.90.0.md]: You can use a four backtick suggestion block and then use three backticks within this block.

Comment (Kobzol) [reply content/Rust-1.90.0.md]: TIL, thanks!

**Review by jieyouxu — APPROVED** [review]

**Review by alex-semenyuk — APPROVED** [review]

File: content/Rust-1.90.0.md [hunk 3]
```diff
@@ -0,0 +1,107 @@
...
+
+If you encounter any issues with the LLD linker, please [let us know](https://github.com/rust-lang/rust/issues/new/choose). You can read more about the switch to LLD, some benchmark numbers and the opt out mechanism [here](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/).
+
+### Cargo adds native support for workspace publishing
+
+`cargo publish --workspace` is now supported, automatically publishing all of
+the crates in a workspace in the right order (following any dependencies
+between them). This has long been possible with external tooling or manual
+ordering of individual publishes, but this brings the functionality into Cargo
+itself.
```
epage: Should we call out the caveat that the publish is not atomic and network or server side errors can leave you in a partially published state?

This is true for existing workarounds but people might have a different expectation with it built in.

content/Rust-1.90.0.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Demoting `x86_64-apple-darwin` to Tier 2 with host tools
+
+GitHub will soon [discontinue][gha-sunset] providing free macOS x86\_64 runners for public repositories. Apple has also announced their [plans][apple] for discontinuing support for the x86\_64 architecture.
+
+In accordance with these changes, as of Rust 1.90, the project has [demoted the `x86_64-apple-darwin` target][rfc] from [Tier 1 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-1-with-host-tools) to [Tier 2 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-2-with-host-tools). This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.
```
Kobzol:
```suggestion
In accordance with these changes, as of Rust 1.90, we have [demoted the `x86_64-apple-darwin` target][rfc] from [Tier 1 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-1-with-host-tools) to [Tier 2 with host tools](https://doc.rust-lang.org/stable/rustc/platform-support.html#tier-2-with-host-tools). This means that the target, including tools like `rustc` and `cargo`, will be guaranteed to build but is not guaranteed to pass our automated test suite.
```
Not sure if people know what "the project" is here.

Kobzol (approved):

content/Rust-1.90.0.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+If you encounter any issues with the LLD linker, please [let us know](https://github.com/rust-lang/rust/issues/new/choose). You can read more about the switch to LLD, some benchmark numbers and the opt out mechanism [here](https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/).
+
+### Cargo adds native support for workspace publishing
+
+`cargo publish --workspace` is now supported, automatically publishing all of
+the crates in a workspace in the right order (following any dependencies
+between them). This has long been possible with external tooling or manual
+ordering of individual publishes, but this brings the functionality into Cargo
```
epage: If we are comparing to existing approaches, is it worth it to call out how this improves on those?

The biggest change is in the verify step which can verify everything before publishing
- for a real publish, this reduces the chance of it failing in a partially published state
- for dry-run, you can now run the verify

epage (approved):

Issue #1696, opened 2025-09-15
Blog post "Rust compiler performance survey 2025 results" without JS looks odd


apiraino:
When viewing that blog post without javaScript images and text alignment are a bit off; images look small. Examples:

<details><summary>Overall satisfaction</summary>
<p>

<img width="976" height="760" alt="Image" src="https://github.com/user-attachments/assets/45fa3249-8e96-458a-bba6-48a49e398d8d" />

</p>
</details>  

[Workarounds for improving build performance](https://blog.rust-lang.org/2025/09/10/rust-compiler-performance-survey-2025-results/#workarounds-for-improving-build-performance):

<details><summary>screenshot</summary>
<p>

<img width="707" height="808" alt="Image" src="https://github.com/user-attachments/assets/0decb6d1-7ce6-4b53-8894-965fd7e71297" />

</p>
</details>

kpreid:
FWIW, as a related but not identical problem, the post displays even worse when viewed in my feed reader (NetNewsWire for Mac), presumably due to scripts executing but an expected stylesheet being absent.

<details><summary>Screenshot</summary>
<img width="1310" height="872" alt="Image" src="https://github.com/user-attachments/assets/671f63eb-039e-4c8c-80ea-0851623ad797" />
</details>

I believe this also occurred with the previous post with similar charts.

senekor:
@Kobzol What is the tool used to generate these charts? It seems to me the main problem is that the (presumably generated) inline-html contains hardcoded height and width. Just removing those makes stuff look fine. Ideally the tool would be fixed to not generate these hardcoded dimensions anymore.

```html
<!-- Chart satisfaction start -->
<div>
  <div class="bar-chart" id="satisfaction" style="height: 400px; width: 100%">
    <noscript>
      <img alt="satisfaction" height="600" src="satisfaction.png" />
    </noscript>
  </div>
  <!-- ... -->
</div>
<!-- Chart satisfaction end -->
```

(IDK what's going on with NetNewsWire, I'm not on Mac.)

PR #1697, opened 2025-09-15
Update compiler performance charts to render better without JS

Kobzol:
Fixes: https://github.com/rust-lang/blog.rust-lang.org/issues/1696

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/master/content/rust-compiler-performance-survey-2025-results/index.md)

### DIFF
--- a/content/rust-compiler-performance-survey-2025-results/index.md
+++ b/content/rust-compiler-performance-survey-2025-results/index.md
@@ -29,8 +29,8 @@ To understand the overall sentiment, we asked our respondents to rate their sati
 
 <!-- Chart satisfaction start -->
 <div>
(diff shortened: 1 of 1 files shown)


Kobzol:
Yeah, unfortunately plotly always generates random IDs, which is super annoying when writing these posts, because everytime I change something, `zola serve` sees 20 new SVGs and goes into a self-induced panic attack where it rebuilds the site N times. I didn't find a way to tell plotly to not generate them randomly, and it seemed like overkill to try to detect these IDs and renumber them manually from the generated VSG.

senekor (approved):
Thanks!

PR #1698, opened 2025-09-19
Rust-1.89.0.md: split sentence into 2, for readability

tshepang:
[Rendered](https://github.com/tshepang/blog.rust-lang.org/blob/patch-1/content/Rust-1.89.0.md)


senekor:
The blog is not living documentation. Unless there is an error that needs to be corrected, we don't generally go back to change old posts.

PR #1713, opened 2025-10-03
chore(ci): Report typos to contributors

epage:
This is a follow up to #1712, adding `typos` to CI like rust-lang/rust (though through tidy and not GH Actions)


senekor (approved):
Thanks!

PR #1724, opened 2025-10-16
Fix claim about Git's default branch name

traviscross:
The blog post about changing the primary branch name of `rust-lang/rust`, as written, claims that the default branch name of *Git*, the open-source version control tool originally written by Linus Torvalds, has changed.  As best I can determine, that's not true, at least as of Git v2.51.  The mailing list thread from 2020, linked in the post, certainly does not support this claim.  That thread discusses making the default branch name for newly-created repositories configurable, and it causes a warning to be shown if this is not configured (indicating that the default may change in the future), but the default when not configured has not been changed.

Conversely, *GitHub*, the service now owned by Microsoft, did change the default branch name for newly-created repositories on 2020-10-01. Probably this is what we mean instead.  Let's change the post to point to that.

cc @shepmaster @Kobzol @carols10cents

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/main/content/inside-rust/renaming-the-default-branch-of-rust-lang-rust.md)

### DIFF
--- a/content/inside-rust/renaming-the-default-branch-of-rust-lang-rust.md
+++ b/content/inside-rust/renaming-the-default-branch-of-rust-lang-rust.md
@@ -8,7 +8,7 @@ team = "the Infra team"
 team_url = "https://www.rust-lang.org/governance/teams/infra#team-infra"
 +++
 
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Huh, I knew about this fact, I could have sworn that I read it as GitHub and not git.. Thanks!

PR #1729, opened 2025-10-25
Announce Rust 1.91.0 stable release

Mark-Simulacrum:
Release notes - https://github.com/rust-lang/rust/pull/148013

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/main/content/Rust-1.91.0.md)


content/Rust-1.91.0.md
```diff
@@ -0,0 +1,170 @@
...
+  = note: pointers do not have a lifetime; after returning, the `u8` will be deallocated
+    at the end of the function because nothing is referencing it as far as the type system is
+    concerned
+  = note: `#[warn(dangling_pointers_from_locals)]` on by default
+```
+
+Note that the code above is not itself unsafe, as it doesn't perform any dangerous
+operations itself. Only dereferencing the raw pointer would be unsafe. We expect future
+releases of Rust to add more functionality helping authors to safely interact with raw
+pointers, and unsafe code more generally.
```
pthariensflame:
```suggestion
Note that the code above is not unsafe, as it itself doesn't perform any dangerous
operations. Only dereferencing the raw pointer after the function returns would be
unsafe. We expect future releases of Rust to add more functionality helping authors
to safely interact with raw pointers, and with unsafe code more generally.
```

Reword for clarity.

content/Rust-1.91.0.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Stabilized APIs
```
pthariensflame, line 81: This is out of date with the current state of rust-lang/rust#148013.

content/Rust-1.91.0.md
```diff
@@ -0,0 +1,170 @@
...
+### Add lint against dangling raw pointers from local variables
+
+While Rust's borrow checking prevents dangling references from being returned, it doesn't
+track raw pointers. With this release, we are adding a warn-by-default lint on raw
+pointers to local variables being returned from functions. For example, code like this:
+
+```rust
+fn f() -> *const u8 {
+    let x = 0;
+    &x
```
traviscross, line 53:
```suggestion
    &raw const x
```

Obviously this coerces anyway, but it seems maybe a bit better to not leave any question in the example about whether this implicit coercion is part of what's triggering this lint.  Also, it's good to take the opportunity to familiarize people with the raw borrow syntax.

Mark-Simulacrum: I'm okay accepting this but it does seem like it might be a distraction for the purposes of this example. It also seems idiomatic to create the borrow where possible, so that miri / rustc / llvm can lint/optimize based on the knowledge that it's safe to borrow.

traviscross: Either way really.  Pros and cons.

content/Rust-1.91.0.md
```diff
@@ -0,0 +1,170 @@
...
+- [`AtomicPtr::fetch_or`](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicPtr.html#method.fetch_or)
+- [`AtomicPtr::fetch_and`](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicPtr.html#method.fetch_and)
+- [`AtomicPtr::fetch_xor`](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicPtr.html#method.fetch_xor)
+- [`{integer}::strict_add`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.strict_add)
+- [`{integer}::strict_sub`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.strict_sub)
+- [`{integer}::strict_mul`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.strict_mul)
+- [`{integer}::strict_div`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.strict_div)
+- [`{integer}::strict_div_euclid`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.strict_div_euclid)
+- [`{integer}::strict_rem`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.strict_rem)
+- [`{integer}::strict_rem_euclid`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.strict_rem_euclid)
```
kpreid: The div/rem entries all link to `u32` documentation that says overflow cannot happen and these methods exist for consistency. Perhaps it would be better to link to `i32` documentation which does have an overflow case, for less chance of looking silly?

```suggestion
- [`{integer}::strict_div`](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.strict_div)
- [`{integer}::strict_div_euclid`](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.strict_div_euclid)
- [`{integer}::strict_rem`](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.strict_rem)
- [`{integer}::strict_rem_euclid`](https://doc.rust-lang.org/stable/std/primitive.i32.html#method.strict_rem_euclid)
```

Mark-Simulacrum: Can you file this against https://github.com/rust-lang/rust/pull/148013? I'll sync from there soon, but that's the authoritative copy of this section.

content/Rust-1.91.0.md
Mark-Simulacrum: I think I lean towards leaving this as-is at this time.

content/Rust-1.91.0.md
kpreid: Done!

PR #1737, opened 2025-11-06
Add blog post for Rust 1.91.1

emilyalbini:
Scheduled to be published on Monday, along with the point release.

cc @rust-lang/release 
cc @alexcrichton, I tried my best to describe the WASM fix without having to explain what mangling is in the blog post, I'd love your input on the section.

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/main/content/Rust-1.91.1.md)

### DIFF
--- a/content/Rust-1.91.1.md
+++ b/content/Rust-1.91.1.md
@@ -1,0 +1,79 @@
++++
+path = "2025/11/10/Rust-1.91.1"
+title = "Announcing Rust 1.91.1"
+authors = ["The Rust Release Team"]
+aliases = ["releases/1.91.1"]
+
+[extra]
+release = true
++++
+
+The Rust team has published a new point release of Rust, 1.91.1. Rust is a
+programming language that is empowering everyone to build reliable and
+efficient software.
+
+If you have a previous version of Rust installed via rustup, getting Rust
+1.91.1 is as easy as:
+
+```
+rustup update stable
+```
+
+If you don't have it already, you can [get `rustup`][rustup] from the
+appropriate page on our website.
+
+[rustup]: https://www.rust-lang.org/install.html
+
+## What's in 1.91.1
+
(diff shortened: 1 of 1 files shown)


content/Rust-1.91.1.md
```diff
@@ -0,0 +1,78 @@
...
+If you don't have it already, you can [get `rustup`][rustup] from the
+appropriate page on our website.
+
+[rustup]: https://www.rust-lang.org/install.html
+
+## What's in 1.91.1
+
+Rust 1.91.1 includes fixes for two regressions introduced in the 1.91.0 release.
+
+### *"Import module mismatch"* linker errors on WASM
```
alexcrichton: I might recommend rewording this to "Linker and runtime errors on Wasm" to emphasize that this isn't just linker errors but it's also runtime UB in a sense.

Also, a more minor point, but Wasm is abbreviated as "Wasm" as opposed to "WASM" (a very common mistake)

content/Rust-1.91.1.md
```diff
@@ -0,0 +1,78 @@
...
+#[link(wasm_import_module = "hello")]
+extern "C" {
+    pub fn world();
+}
+```
+
+Rust 1.91.0 introduced a regression in the attribute, causing linker failures
+with the "import module mismatch" error (at either compile time or runtime,
+depending on the linking configuration) when a symbol within a module had the
+same name as symbols in other modules or non-WASM symbols.
```
alexcrichton: Along the lines of emphasizing the runtime aspect here too, I might recommend:

> Rust 1.91.0 introduced a regression in the attribute which can lead to compile-time linker failures (such as "import module mismatch") or using the wrong function at runtime (leading to traps or silent corruption). This happened when the same symbol was imported from two different `wasm_import_module` names across at least two different Rust crates.

alexcrichton (approved):
Thanks!

PR #1757, opened 2025-12-03
Announce Rust 1.92 stable release

BoxyUwU:
Draft release notes - https://github.com/rust-lang/rust/issues/149161

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/main/content/Rust-1.92.0.md)

### DIFF
--- a/content/Rust-1.92.0.md
+++ b/content/Rust-1.92.0.md
@@ -30,7 +30,7 @@ It's worth noting that while this can result in compilation errors, it is still 
 
 These lints detect code which is likely to be broken by the never type stabilization. It is highly advised to fix them if they are reported in your crate graph.
 
-We believe there to be approximately 500 crates affected by this lint. Despite that, we believe this to be acceptable, as lints are not a breaking change and it will allow for stabilizing the never type in the future. For more in-depth justification, the Language Team's assessment can be read here: [rust-lang/rust#146167#issuecomment-3363795006](https://github.com/rust-lang/rust/pull/146167#issuecomment-3363795006).
+We believe there to be approximately 500 crates affected by this lint. Despite that, we believe this to be acceptable, as lints are not a breaking change and it will allow for stabilizing the never type in the future. For more in-depth justification, see the [Language Team's assessment](https://github.com/rust-lang/rust/pull/146167#issuecomment-3363795006).
 
 ### `unused_must_use` no longer warns about `Result<(), UninhabitedType>`
 


content/Rust-1.92.0.md
```diff
@@ -0,0 +1,116 @@
...
+
+...
+
+These previously stable APIs are now stable in const contexts:
+
+...
+
+### Platform Support
+
+Refer to Rust’s [platform support page][platform-support] for more information on Rust’s tiered platform support.
```
BoxyUwU: Do we just drop this section when there's no changes to platform support :laughing:

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,116 @@
...
+impl UsesAssocErrorType for CanFail {
+    type Error = std::io::Error;
+    fn method(&self) -> Result<(), Self::Error> {
+        Err(std::io::Error::other("something went wrong"))
+    }
+}
+
+fn main() {
+    CannotFail.method(); // No error
+    CanFail.method(); // Error: unused `Result` that must be used
```
Kobzol:
```suggestion
    CannotFail.method(); // No warning
    CanFail.method(); // Warning: unused `Result` that must be used
```
Those are warnings, not errors, right?

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Validate input to `#[macro_export]`
+
+Over the past few releases, many changes were made to the way built-in attributes are processed in the compiler. This should greatly improve the error messages and warnings rust gives for built-in attributes and especially make these diagnostics more consistent among all of the over 100 built-in attributes.
+
+To give a small example, in this release specifically, Rust became stricter in checking what arguments are allowed to `macro_export` by [upgrading that check to a "deny-by-default lint" that will be reported in dependencies.](https://github.com/rust-lang/rust/pull/143857). 
+
+### Restrictions on user code impls of `DerefMut` for `Pin`
+
+A soundness issue with `Pin` has been solved by preventing third-party crates from implementing `DerefMut` for `Pin<T>` when `T` is a local type that doesn't implement `DerefMut<Target: Unpin>`.
```
Kobzol: This seems a bit hard to grok without context. Maybe at least a PR/issue link would make it clearer?

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Emit unwind tables even when `-Cpanic=abort` is enabled on linux
+
+Backtraces with `-Cpanic=abort` previously worked in Rust 1.22 but were broken in Rust 1.23, as we stopped emitting unwind tables with `-Cpanic=abort`. In Rust 1.45 a workaround in the form of `-Cforce-unwind-tables=yes` was stabilized.
+
+In Rust 1.92 unwind tables will be emitted by default even when `-Cpanic=abort` is specified, allowing for backtraces to work properly. If unwind tables are not desired then users should use `-Cforce-unwind-tables=no` to explicitly disable them being emitted.
+
+### Validate input to `#[macro_export]`
+
+Over the past few releases, many changes were made to the way built-in attributes are processed in the compiler. This should greatly improve the error messages and warnings rust gives for built-in attributes and especially make these diagnostics more consistent among all of the over 100 built-in attributes.
```
tshepang:
```suggestion
Over the past few releases, many changes were made to the way built-in attributes are processed in the compiler. This should greatly improve the error messages and warnings Rust gives for built-in attributes and especially make these diagnostics more consistent among all of the over 100 built-in attributes.
```

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,116 @@
...
+
+If you'd like to help us out by testing future releases, you might consider updating locally to use the beta channel (`rustup default beta`) or the nightly channel (`rustup default nightly`). Please [report](https://github.com/rust-lang/rust/issues/new/choose) any bugs you might come across!
+
+## What's in 1.92.0 stable
+
+### Deny-by-default never type lints
+
+The language and compiler teams continue to work on stabilization of the [never type](https://doc.rust-lang.org/stable/std/primitive.never.html). In this release the [`never_type_fallback_flowing_into_unsafe`](https://doc.rust-lang.org/beta/rustc/lints/listing/deny-by-default.html#dependency-on-unit-never-type-fallback) and [`dependency_on_unit_never_type_fallback`](https://doc.rust-lang.org/beta/rustc/lints/listing/deny-by-default.html#dependency-on-unit-never-type-fallback) future compatibility lints were made deny-by-default, meaning they will cause a compilation error when detected.
+
+These lints detect code which is likely to be broken by the never type stabilization. It is highly advised to fix them if they are reported in your crate.
```
Mark-Simulacrum: I wonder if we should add something about how wide the breakage here is based on crater or something? As is it sounds a bit like "rust is making breaking changes" which while semi-true probably needs caveats for a blog post.

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,116 @@
...
+The `unused_must_use` lint now no longer warns on `Result<(), UninhabitedType>`, or on `ControlFlow<UninhabitedType, ()>`. For instance, it will not warn on `Result<(), !>`. This avoids having to check for an error that can never happen.
+
+```rust
+fn can_never_fail() -> Result<(), !> {
+    // ...
+    Ok(())
+}
+
+fn main() {
+    can_never_fail()
```
Mark-Simulacrum: This doesn't type check, right? You still need a `;` here?

content/Rust-1.92.0.md
Mark-Simulacrum: I suspect we can just delete this item.

content/Rust-1.92.0.md
traviscross: It's worth mentioning too that 1) a deny-by-default lint only causes an error by default --  it's still a lint, and the lint can be suppressed, and 2) the error only occurs when building the crate that contains the linted-on code, not when building its dependents.  (A warning will be reported in the builds of the dependents, when using Cargo.)

content/Rust-1.92.0.md
BoxyUwU: good catch!

content/Rust-1.92.0.md
BoxyUwU: Yeah I think this is probably quite in the weeds for the blog post

content/Rust-1.92.0.md
BoxyUwU: good catch

content/Rust-1.92.0.md
```diff
@@ -26,24 +26,29 @@ If you'd like to help us out by testing future releases, you might consider upda
 
 The language and compiler teams continue to work on stabilization of the [never type](https://doc.rust-lang.org/stable/std/primitive.never.html). In this release the [`never_type_fallback_flowing_into_unsafe`](https://doc.rust-lang.org/beta/rustc/lints/listing/deny-by-default.html#dependency-on-unit-never-type-fallback) and [`dependency_on_unit_never_type_fallback`](https://doc.rust-lang.org/beta/rustc/lints/listing/deny-by-default.html#dependency-on-unit-never-type-fallback) future compatibility lints were made deny-by-default, meaning they will cause a compilation error when detected.
 
-These lints detect code which is likely to be broken by the never type stabilization. It is highly advised to fix them if they are reported in your crate.
+It's worth noting that while this can result in compileration errors, it is still a *lint*, these lints can all be `#[allow`'d. These lints also will only fire when building the affected crates directly, not when built as a dependency (though a warning will be reported by Cargo in such cases).
```
bjorn3:
```suggestion
It's worth noting that while this can result in compileration errors, it is still a *lint*, these lints can all be `#[allow]`'d. These lints also will only fire when building the affected crates directly, not when built as a dependency (though a warning will be reported by Cargo in such cases).
```

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,128 @@
...
+
+## What's in 1.92.0 stable
+
+### Deny-by-default never type lints
+
+The language and compiler teams continue to work on stabilization of the [never type](https://doc.rust-lang.org/stable/std/primitive.never.html). In this release the [`never_type_fallback_flowing_into_unsafe`](https://doc.rust-lang.org/beta/rustc/lints/listing/deny-by-default.html#dependency-on-unit-never-type-fallback) and [`dependency_on_unit_never_type_fallback`](https://doc.rust-lang.org/beta/rustc/lints/listing/deny-by-default.html#dependency-on-unit-never-type-fallback) future compatibility lints were made deny-by-default, meaning they will cause a compilation error when detected.
+
+It's worth noting that while this can result in compileration errors, it is still a *lint*, these lints can all be `#[allow]`'d. These lints also will only fire when building the affected crates directly, not when built as a dependency (though a warning will be reported by Cargo in such cases).
```
kpreid: 1 spelling error, agreement in number:

```suggestion
It's worth noting that while this can result in compilation errors, it is still a *lint;* these lints can all be `#[allow]`ed. These lints also will only fire when building the affected crates directly, not when they are built as dependencies (though a warning will be reported by Cargo in such cases).
```

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,128 @@
...
+
+These lints detect code which is likely to be broken by the never type stabilization. It is highly advised to fix them if they are reported in your crate graph.
+
+We believe there to be approximately ~500 crates affected by this lint. Despite that we believe this to be acceptable as lints are not a breaking change and it will allow for stabilizing the never type in the future. For more in depth justification the Language Team's assessment can be read here: [rust-lang/rust#146167#issuecomment-3363795006](https://github.com/rust-lang/rust/pull/146167#issuecomment-3363795006).
```
kpreid: “approximately ~500” is saying “approximately” two ways.

```suggestion
We believe there to be approximately 500 crates affected by this lint. Despite that, we believe this to be acceptable, as lints are not a breaking change and it will allow for stabilizing the never type in the future. For more in-depth justification, the Language Team's assessment can be read here: [rust-lang/rust#146167#issuecomment-3363795006](https://github.com/rust-lang/rust/pull/146167#issuecomment-3363795006).
```

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,128 @@
...
+
+Backtraces with `-Cpanic=abort` previously worked in Rust 1.22 but were broken in Rust 1.23, as we stopped emitting unwind tables with `-Cpanic=abort`. In Rust 1.45 a workaround in the form of `-Cforce-unwind-tables=yes` was stabilized.
+
+In Rust 1.92 unwind tables will be emitted by default even when `-Cpanic=abort` is specified, allowing for backtraces to work properly. If unwind tables are not desired then users should use `-Cforce-unwind-tables=no` to explicitly disable them being emitted.
+
+### Validate input to `#[macro_export]`
+
+Over the past few releases, many changes were made to the way built-in attributes are processed in the compiler. This should greatly improve the error messages and warnings Rust gives for built-in attributes and especially make these diagnostics more consistent among all of the over 100 built-in attributes.
+
+To give a small example, in this release specifically, Rust became stricter in checking what arguments are allowed to `macro_export` by [upgrading that check to a "deny-by-default lint" that will be reported in dependencies.](https://github.com/rust-lang/rust/pull/143857).
```
kpreid: This sentence has 2 periods at the end.

```suggestion
To give a small example, in this release specifically, Rust became stricter in checking what arguments are allowed to `macro_export` by [upgrading that check to a "deny-by-default lint" that will be reported in dependencies](https://github.com/rust-lang/rust/pull/143857). 
```

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,128 @@
...
+
+We believe there to be approximately 500 crates affected by this lint. Despite that, we believe this to be acceptable, as lints are not a breaking change and it will allow for stabilizing the never type in the future. For more in-depth justification, the Language Team's assessment can be read here: [rust-lang/rust#146167#issuecomment-3363795006](https://github.com/rust-lang/rust/pull/146167#issuecomment-3363795006).
+
+### `unused_must_use` no longer warns about `Result<(), UninhabitedType>`
+
+Rust's `unused_must_use` lint warns when ignoring the return value of a function, if the function or its return type is annotated with `#[must_use]`. For instance, this warns if ignoring a return type of `Result`, to remind you to use `?`, or something like `.expect("...")`.
+
+However, some functions return `Result`, but the error type they use is not actually "inhabited", meaning it can never exist in real code (e.g. the [`!`](https://doc.rust-lang.org/std/primitive.never.html) or [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) types).
```
fmease: The phrasing “[the error type] can never exist in real code” is technically incorrect (*values* of that type can't exist) and generally confusing (what is *real* code? well, it's likely referring to well-formed code but is that even something worth mentioning?)

```suggestion
However, some functions return `Result`, but the error type they use is not actually "inhabited", meaning values of that type don't exist (e.g. the [`!`](https://doc.rust-lang.org/std/primitive.never.html) or [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) types).
```

or *[…] no values of that type exist* or


```suggestion
However, some functions return `Result`, but the error type they use is not actually "inhabited", meaning you cannot construct any values of that type (e.g. the [`!`](https://doc.rust-lang.org/std/primitive.never.html) or [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) types).
```

or sth. like that...

content/Rust-1.92.0.md
```diff
@@ -0,0 +1,128 @@
...
+
+It's worth noting that while this can result in compilation errors, it is still a *lint;* these lints can all be `#[allow]`ed. These lints also will only fire when building the affected crates directly, not when they are built as dependencies (though a warning will be reported by Cargo in such cases).
+
+These lints detect code which is likely to be broken by the never type stabilization. It is highly advised to fix them if they are reported in your crate graph.
+
+We believe there to be approximately 500 crates affected by this lint. Despite that, we believe this to be acceptable, as lints are not a breaking change and it will allow for stabilizing the never type in the future. For more in-depth justification, the Language Team's assessment can be read here: [rust-lang/rust#146167#issuecomment-3363795006](https://github.com/rust-lang/rust/pull/146167#issuecomment-3363795006).
```
Mark-Simulacrum:
```suggestion
We believe there to be approximately 500 crates affected by this lint. Despite that, we believe this to be acceptable, as lints are not a breaking change and it will allow for stabilizing the never type in the future. For more in-depth justification, see the [Language Team's assessment](https://github.com/rust-lang/rust/pull/146167#issuecomment-3363795006).
```

Issue #1796, opened 2026-02-11
Update to Zola 0.22


LawnGnome:
The blog currently requires Zola 0.21, but the current stable version of Zola is 0.22.

Unfortunately, Zola 0.22 [significantly changes the way syntaxes and themes are handled](https://github.com/getzola/zola/blob/daf2f31fc2e2d8dd8089edc17429796f11fbc999/CHANGELOG.md#0220-2026-01-09). I can see two immediate problems:

1. The hack that's in place to declare a `text` syntax will no longer work. While it could probably be reimplemented, only [one blog post](https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/) actually uses it, so I'm more inclined to suggest that it should be removed and that single blog post updated.
2. Theming has also completely changed. To maintain the current formatting style, the `Boron` and `OneHalfLight` themes that are currently implemented as CSS stylesheets would need to be reimplemented in the TextMate JSON format that Zola 0.22 expects. Both are available as XML ([`Boron`](https://github.com/mjio/boron.tmtheme), [`OneHalfLight`](https://github.com/sonph/onehalf-tmtheme)), so this shouldn't be rocket surgery, but it's not completely trivial. (Or, alternatively, the blog could be switched to one of the themes built into Zola.)

There may be more icebergs lurking as well, but before I put any effort into trying to fix the above problems, I wanted to check to see if this work was even wanted, and if so, whether the above plan of action made sense. (Namely: remove the `text` syntax, and either port the TextMate XML themes or ~~bikeshed~~ select new built-in themes.)

senekor:
> The hack that's in place to declare a text syntax will no longer work.

That should be OK. The reason we have it is because we expect(ed) blog authors to use that a lot. Apparently that's not the case. If necessary, we could write a little CI test that makes sure there are no occurrences of the `text` syntax. Worst case scenario, we get a few build warnings that require manual cleanup every once in a while.

For reference, here's the original thread that discussed this:
- https://github.com/rust-lang/blog.rust-lang.org/pull/1563#discussion_r2042665106
- and PR: https://github.com/rust-lang/blog.rust-lang.org/pull/1566

As far as theming goes, I'm fine with switching to different built-in themes provided by Zola. We just have to pay attention that the contrast is good on both themes, keeping accessibility in mind.

I don't think this is very urgent, but PRs are definitely welcome.

PR #1797, opened 2026-02-12
Update to Zola 0.22.1

LawnGnome:
Per the discussion in #1796, this PR updates the blog to use Zola 0.22. As part of that, the `text` syntax has been removed, and the built-in `light-plus` and `dark-plus` syntax highlighting themes are now used.

The snapshot test diffs here are going to look horrific, but spot checking a few blog pages in practice reveals only a couple of apparent differences:

1. The light mode no longer highlights Rust type names with a colour, and
2. The dark mode now highlights quotes with the same colour as the actual string literal within the quotes.

The second change actually slightly improves the WCAG contrast scores. The first one makes no effective difference, although I personally miss the colour.

I obviously have no way to test the GitHub action changes, so we'll just have to see what happens. It works locally if I recreate the steps in a `rust:1.91.1` Docker container. :shrug: 

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/ca587be39470a2bbb09b5e4f8f23b1ce500cd1ad/content/Rust-1.89.0.md)


senekor:
Just regular, unhighlighted, "black on white" text.

senekor:
superseded by #1848

PR #1842, opened 2026-04-28
Added missing mod keywords to blog posts

vallentin:
As the title says.

I was reading the latest 1.95.0 blog post a few days ago, and for some reason under "Stabilized APIs" the `mod core::range` stood out to me. I thought `mod` was a copy paste mistake.

However, after looking through the other blog posts, I noticed the common denominator was to have the `mod` keyword. So instead of removing it, I added the `mod` keyword the few places it was forgotten.

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/63059ce98a0785e64f938b63f40ad952c7bb6bee/content/Rust-1.81.0.md)


senekor:
Blog posts aren't supposed to be living documents. I don't think we should go back in time very far to change minor things.

PR #1855, opened 2026-05-13
Fix CI caching

ehuss:
Whenever the Zola version is updated, this breaks the CI cache. The Swatinem/rust-cache action doesn't know anything about installed versions. This means that the cache won't get updated, but every time CI runs, cargo will install the new version.

This fixes it by including the version of Zola installed in the cache key.

Note that I think the caching in the `snapshot_tests.yml` file is broken. IIRC, PR jobs cannot share a cache between PR jobs. The only way to fix this is to have the job also run on `main`. Since this job is almost never run, I would recommend just removing the caching from it. I will leave that up to @senekor.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,7 @@ on:
 env:
   # renovate: datasource=github-tags depName=rust lookupName=rust-lang/rust
   RUST_VERSION: 1.95.0
+  ZOLA_VERSION: 29540e9897dbe8aca388b13f7bdf615985f6ca2c
 
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


senekor (approved):

Kobzol:
Huh, that behavior of rust-cache surprised me, I thought it would always write the cache, based on:

```
# Determines whether the cache should be saved.
    # If `false`, the cache is only restored.
    # Useful for jobs where the matrix is additive e.g. additional Cargo features,
    # or when only runs from `master` should be saved to the cache.
    # default: "true"
    save-if: ""
    # To only cache runs from `master`:
    save-if: ${{ github.ref == 'refs/heads/master' }}
```
But apparently the saving is still done only if the cache was missing or if it was broken (in a way that is registered by `rust-cache`).

Good catch, I was wondering why `zola` is being rebuilt all the time. Thanks for taking a look.

PR #1867, opened 2026-06-04
josh -> JOSH

RalfJung:


Kobzol:
@christian-schilling Do you want to rename it in the blog post? :laughing: I interpreted your message on Zulip as "maybe we should rename the project", but right now on GitHub and in the docs it's always josh or Josh, so it seems a bit weird to CAPSLOCK it here :) But I'm fine with it if you think it would reduce confusion.

christian-schilling:
> @christian-schilling Do you want to rename it in the blog post? :laughing: I interpreted your message on Zulip as "maybe we should rename the project", but right now on GitHub and in the docs it's always josh or Josh, so it seems a bit weird to CAPSLOCK it here :) But I'm fine with it if you think it would reduce confusion.

Yeah we also have it as "Josh" everywhere including website / logo.
So quite unsure about this. Maybe better to say "Josh Project" to disambiguate.
Names are so hard 🙈

christian-schilling:
need to think a bit more about this 🤔

RalfJung:
Oh sorry, I somehow thought this was already the official name.

PR #1892, opened 2026-07-17
Check publication date in pull requests

nxsaken:
This should make it less likely (or impossible with branch protection) to merge a PR with a placeholder publication date.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - main
   pull_request:
+    branches:
+      - main
 
 env:
   # renovate: datasource=github-tags depName=rust lookupName=rust-lang/rust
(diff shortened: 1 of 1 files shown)


senekor:
This will make it so CI fails every time while a PR is in a "draft" phase, right? when the publication date is not decided yet. I don't want that. It's annoying for authors to get emails all the time with failed CI.

The way it work currently is that CI will fail after merging, which will still send emails to... at least me and probably the PR author? somebody will catch it soon enough and fix it.

If we spam blog authors with failed CI emails, that will incentivize them to fix a specific date right away. That would be much worse though, because we can't detect in CI if that was the actually correct publication date. I prefer the current system.

A proper fix would be to work with merge queues I think.

Kobzol:
Yeah, a merge queue would be a better solution.

senekor:
I thought about it when I initially did the CI stuff, but I think I don't have the needed permissions on the repo. I thought it's not worth bothering someone from the infra team. But if it's not much trouble, I wouldn't say no either :)

nxsaken:
Could we replace the placeholder with the current date in CI and avoid failing at all?

Kobzol:
Opened https://github.com/rust-lang/team/pull/2590 and https://github.com/rust-lang/blog.rust-lang.org/pull/1893 for the merge queue.

> Could we replace the placeholder with the current date in CI and avoid failing at all?

That would require performing a commit on CI, which is a can of worms. I think that failing during a merge, but still keeping PR CI green, is fine.

nxsaken:
Thanks! :) I'll close this, then.

PR #1893, opened 2026-07-17
Use merge queue

Kobzol:
Requested in https://github.com/rust-lang/blog.rust-lang.org/pull/1892#issuecomment-5002484621.

Note that I left the CI running also on `main`, otherwise we wouldn't get proper caching in the merge queue, due to GitHub limitations.

We could deploy from `main`, but since we have a merge queue now, I think that deploying directly from it is slightly better.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
   pull_request:
+  merge_group:
 
 env:
   # renovate: datasource=github-tags depName=rust lookupName=rust-lang/rust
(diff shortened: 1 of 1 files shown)


senekor (approved):
thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1893/aee4fb3ceb4c38e06f28765195a71849c00737bd..07909142ed8131737f5c5d7d2aa8f5f81d53d193)*

PR #1905, opened 2026-08-05
Improve styling of tasklists

### LINKED ISSUE #1904: Tasklist checkboxes are almost unreadable, especially on dark mode
I've noticed that GitHub tasklist don't look particularly appealing on the blog website. Take a look for instance at this set of screenshots:

Image

Especially the display in dark mode has the difference between checked and unchecked boxes barely visible.

The reason is that checkboxes are inherently *input* elements, and they've been disabled because they don't function as inputs here. This makes sense on GitHub because when you're logged in and have edit permissions for a comment, these checkboxes actually *become* clickable, interactive elements.

On the blog website however, we never have any interactable checkboxes anyway. In my opinion it would make sense to just give them the active styling then.

---

Upon closer inspection, I also noticed that the general rendering style is totally different from GitHub. This is of course because GitHub special-cases the styling of these constructs (markdown doesn't even have checkbox elements outside of these). While the blog posts shows a checkbox *after* the bullet point of the unordered list, GitHub actually themes these such that the box replaces the bullet point. For example:

- [ ] Example task list item
- [x] Second item. We'll make this one a bit longer, too… Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed eiusmod tempor incidunt ut labore et dolore magna aliqua.
- a bullet point for comparison next to this

*Edit:* I'm now[^2] noticing that GitHub seems to have a *different* rendering/representation for these elements in issue comments, actually, that's way more complicated than in other rendered markdown files[^1]. To see the simpler representation I was actually looking at for comparison [compare e.g. how the above displays in a gist](https://gist.github.com/steffahn/0d587271214bd8c1022c62caca122199).

[^2]: because the preview in the editor *doesn't* actually do the same thing
[^1]: I would assume this is the case in order to support the fancy ability of drag-and-drop reordering task lists.

Image

The blog will currently render this as follows (screenshot made in Chrome):

Image

---

Solving both of these issues should be possible. My own (perfectly functional) approach to fixing this issue this has unfortunately been closed with relatively little discussion, but I'll link it here for future reference in case anyone else wants to try solve this problem in ways that fulfills the additional design constraints raised by maintainers: 
* #1901

---

With such changes in place, i.e. making the checkbox element appear active instead of greyed out, and making it take the bullet point's position, the above examples would look as follows instead:

Image

<details><summary><i>[Click to show one alternative with slightly different spacing (checkbox further to the left).]</i></summary>
Image
</details>

Image

---

If we wanted to follow GitHub's example on how to pull this off, note that *their approach* does involve changes to the underlying HTML. They add classes to make the tasklist entries easy to target via CSS, and include an `aria-label` in order to make the tasklist entry's state accessible to technology such a screen readers (explaining the status as "Incomplete task" or "Completed task").

E.g. inspecting the source of the above example ~in this very issue post  on GitHub~ *in the linked gist* shows the following HTML

```html
<ul class="contains-task-list">
<li class="task-list-item"><input type="checkbox" id="" disabled="" class="task-list-item-checkbox" aria-label="Incomplete task"> Example task list item</li>
<li class="task-list-item"><input type="checkbox" id="" disabled="" class="task-list-item-checkbox" aria-label="Completed task" checked=""> Second item. We'll make this one a bit longer, too… Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed eiusmod tempor incidunt ut labore et dolore magna aliqua.</li>
<li>a bullet point for comparison next to this</li>
</ul>
```

Changing the HTML code that Markdown compiles down to with Zola is not *directly* possible as far as I can tell. However it's actually **very easy** to find and replace the checkboxes *reliably* with tera filters like [replace](https://keats.github.io/tera/#replace) (or slightly more DRY with a regex and the Zola-profided filter of [regex_replace](https://www.getzola.org/documentation/templates/overview/#regex-replace)), to change them into a form that's more susceptible to proper display (both visually, and also for accessibility tools).

senekor:
closes #1904 

Also aligns text next to checkbox, so the checkbox visually serves the purpose of an item bullet.

before:
Image

after:
image

### DIFF
--- a/sass/styles/app.scss
+++ b/sass/styles/app.scss
@@ -257,6 +257,25 @@ ul {
   list-style-type: disc;
 }
 
+/* Remove bullet from task lists. */
+/* The checkbox we render afterwards serves that purpose. */
+ul:has(> li > input[type="checkbox"]) {
+  list-style-type: none;
+  padding-left: 0;
+}
+
+/* Add checkbox emoji in front of task list item and hide regular checkbox. */
+/* The regular checkbox is disabled, which is rendered with an unreadable gray color. */
+li > input[type="checkbox"] {
+  display: none
+}
+li:has(> input[type="checkbox"]:checked)::before {
+    content: "✅ ";
+}
+li:has(> input[type="checkbox"]:not(:checked))::before {
+    content: "⬜ ";
+}
+
 .posts {
   background-color: var(--gray);
   color: var(--body-foreground-color);
--- a/sass/styles/app.scss
+++ b/sass/styles/app.scss
@@ -76,7 +76,7 @@ body {
   background-color: var(--body-background-color);
   color: var(--body-color);
 
-  /* Ensure the footer is always at the bottom of the screen */
+  // Ensure the footer is always at the bottom of the screen
   min-height: 100vh;
   display: flex;
   flex-direction: column;
@@ -257,30 +257,30 @@ ul {
   list-style-type: disc;
 }
 
-/* Special handling of item markers for lists that contain task items (checkboxes). */
+// Special handling of item markers for lists that contain task items (checkboxes).
 ul:has(> li > input[type="checkbox"]) {
-  /* Create enough space to the left of the items for custom markers. */
+  // Create enough space to the left of the items for custom markers.
   padding-left: 2cap;
 
   > li:has(> input[type="checkbox"]) {
-    /* Hide default markers from task items. */
+    // Hide default markers from task items.
     list-style-type: none;
-    /* Set position to relative, so we can offset the custom marker relative to this element. */ 
+    // Set position to relative, so we can offset the custom marker relative to this element. 
     position: relative;
 
-    /* Hide default checkbox, which is rendered with an unreadable gray color. */
+    // Hide default checkbox, which is rendered with an unreadable gray color.
     > input[type="checkbox"] {
       display: none
     }
   }
-  /* Add custom emoji markers to task items. */
+  // Add custom emoji markers to task items.
   > li:has(> input[type="checkbox"]:checked)::before {
     content: "✅ ";
   }
   > li:has(> input[type="checkbox"]:not(:checked))::before {
       content: "⬜ ";
   }
-  /* Position custom markers for proper alignment. */
+  // Position custom markers for proper alignment.
   > li:has(> input[type="checkbox"])::before {
     position: absolute;
     left: -1.8em;
--- a/sass/styles/app.scss
+++ b/sass/styles/app.scss
@@ -257,11 +257,15 @@ ul {
   list-style-type: disc;
 }
 
-/* Remove bullet from task lists. */
-/* The checkbox we render afterwards serves that purpose. */
+/* For lists containing task items, create enough space for the custom emoji markers. */
 ul:has(> li > input[type="checkbox"]) {
+  padding-left: 1.8em;
+}
+
+/* Remove marker from task list items. */
+/* The checkbox we render afterwards serves that purpose. */
+ul > li:has(> input[type="checkbox"]) {
   list-style-type: none;
-  padding-left: 0;
 }
 
 /* Add checkbox emoji in front of task list item and hide regular checkbox. */
@@ -276,14 +280,13 @@ li:has(> input[type="checkbox"]:not(:checked))::before {
     content: "⬜ ";
 }
 
-/* Make paragraph next to checkbox aligned as if checkbox was a list bullet. */
+/* Offset custom task item marker by the space we created earlier. */
 li:has(> input[type="checkbox"]) {
   position: relative;
-  padding-left: 1.8em;
 }
 li:has(> input[type="checkbox"])::before {
   position: absolute;
-  left: 0;
+  left: -1.8em;
   top: 0;
 }
 
--- a/sass/styles/app.scss
+++ b/sass/styles/app.scss
@@ -276,6 +276,17 @@ li:has(> input[type="checkbox"]:not(:checked))::before {
     content: "⬜ ";
 }
 
+/* Make paragraph next to checkbox aligned as if checkbox was a list bullet. */
+li:has(> input[type="checkbox"]) {
+  position: relative;
+  padding-left: 1.8em;
+}
+li:has(> input[type="checkbox"])::before {
+  position: absolute;
+  left: 0;
+  top: 0;
+}
+
 .posts {
   background-color: var(--gray);
   color: var(--body-foreground-color);
--- a/sass/styles/app.scss
+++ b/sass/styles/app.scss
@@ -257,37 +257,36 @@ ul {
   list-style-type: disc;
 }
 
-/* For lists containing task items, create enough space for the custom emoji markers. */
+/* Special handling of item markers for lists that contain task items (checkboxes). */
 ul:has(> li > input[type="checkbox"]) {
-  padding-left: 1.8em;
-}
+  /* Create enough space to the left of the items for custom markers. */
+  padding-left: 2cap;
 
-/* Remove marker from task list items. */
-/* The checkbox we render afterwards serves that purpose. */
-ul > li:has(> input[type="checkbox"]) {
-  list-style-type: none;
-}
+  > li:has(> input[type="checkbox"]) {
+    /* Hide default markers from task items. */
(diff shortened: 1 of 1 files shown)


sass/styles/app.scss
```diff
@@ -257,6 +257,36 @@ ul {
   list-style-type: disc;
 }
 
+/* Remove bullet from task lists. */
+/* The checkbox we render afterwards serves that purpose. */
+ul:has(> li > input[type="checkbox"]) {
+  list-style-type: none;
+  padding-left: 0;
+}
```
steffahn: Probably better to move this onto the `li` itself to more gracefully handle mixed cases

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1905/be5518a44a184f443010cb53233291aaef670c9a..520ef603c5faae5c859a2e7a47f21fde9504c555)*

sass/styles/app.scss
```diff
@@ -257,6 +257,36 @@ ul {
...
+/* The regular checkbox is disabled, which is rendered with an unreadable gray color. */
+li > input[type="checkbox"] {
+  display: none
+}
+li:has(> input[type="checkbox"]:checked)::before {
+    content: "✅ ";
+}
+li:has(> input[type="checkbox"]:not(:checked))::before {
+    content: "⬜ ";
+}
```
steffahn: Could these use `::marker` instead? Actually.. [perhaps not](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::marker) on Safari :-/

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1905/be5518a44a184f443010cb53233291aaef670c9a..520ef603c5faae5c859a2e7a47f21fde9504c555)*

sass/styles/app.scss
```diff
@@ -257,6 +257,36 @@ ul {
...
+/* Make paragraph next to checkbox aligned as if checkbox was a list bullet. */
+li:has(> input[type="checkbox"]) {
+  position: relative;
+  padding-left: 1.8em;
+}
+li:has(> input[type="checkbox"])::before {
+  position: absolute;
+  left: 0;
+  top: 0;
+}
```
steffahn:
```suggestion
/* Make paragraph next to checkbox aligned as if checkbox was a list bullet. */
li:has(> input[type="checkbox"]) {
  position: relative;
}
li:has(> input[type="checkbox"])::before {
  position: absolute;
  left: -1.8em;
  top: 0;
}
```

(see [above](https://github.com/rust-lang/blog.rust-lang.org/pull/1905#discussion_r3720568226))

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1905/be5518a44a184f443010cb53233291aaef670c9a..520ef603c5faae5c859a2e7a47f21fde9504c555)*

sass/styles/app.scss
steffahn:
```suggestion
/* Remove bullet from task lists. */
/* The checkbox we render afterwards serves that purpose. */
ul:has(> li > input[type="checkbox"]) {
  padding-left: 1.8em;
}

ul > li:has(> input[type="checkbox"]) {
  list-style-type: none;
}
```

this, together with [this](https://github.com/rust-lang/blog.rust-lang.org/pull/1905#discussion_r3720565696) can make mixed lists work in principle

(free feel to then re-shuffle to make the comments apply more sensibly again)

steffahn:
Here's an alternative CSS that makes the checkbox emojis a bit less huge:

```scss
/* Remove bullet from task lists. */
/* Add checkbox emoji in front of task list item and hide regular checkbox. */
/* The regular checkbox is disabled, which was otherwise rendered with an unreadable gray color. */
ul:has(> li > input[type="checkbox"]) {
  padding-left: 2cap;

  > li:has(> input[type="checkbox"]) {
    list-style-type: none;
    position: relative;

    > input[type="checkbox"] {
      display: none
    }
  }
  > li:has(> input[type="checkbox"])::before {
    position: absolute;

    left: calc(-1.7em);
    top: calc(1em - 1cap);
    font-size: 1cap;
  }
  > li:has(> input[type="checkbox"]:checked)::before {
    content: "✅ ";
  }
  > li:has(> input[type="checkbox"]:not(:checked))::before {
    content: "⬜ ";
  }
}
```

### Comparison

*adjusted sizing*

Bildschirmfoto_20260805_154556

*current sizing in PR*
Bildschirmfoto_20260805_154612

sass/styles/app.scss
senekor: We don't have such mixed lists currently. But your suggestion works and is simple enough, here's a screenshot from my testing:

image

steffahn:
I must say, the emoji approach is growing on me, too :3

senekor:
I like the smaller font size and CSS refactoring! applied.

sass/styles/app.scss
```diff
@@ -257,6 +257,38 @@ ul {
...
+  > li:has(> input[type="checkbox"]:checked)::before {
+    content: "✅ ";
+  }
+  > li:has(> input[type="checkbox"]:not(:checked))::before {
+      content: "⬜ ";
+  }
+  /* Position custom markers for proper alignment. */
+  > li:has(> input[type="checkbox"])::before {
+    position: absolute;
+    left: calc(-1.8em);
```
steffahn: oops, I had left an unnecessary calc there

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1905/be5518a44a184f443010cb53233291aaef670c9a..bdca0371132ebb4861fef22df03d5719f7443165)*

sass/styles/app.scss
```diff
@@ -257,6 +257,38 @@ ul {
   list-style-type: disc;
 }
 
+/* Special handling of item markers for lists that contain task items (checkboxes). */
+ul:has(> li > input[type="checkbox"]) {
+  /* Create enough space to the left of the items for custom markers. */
+  padding-left: 2cap;
```
steffahn: In case you're wondering about units, `2cap` here would cancel out exactly `2em` in the `position: absolute; left: -…` sizing further down below (due to the different font size these positions relate to). So with your change from `1.7` to `1.8`, it means the boxes are `0.2cap` indented from the left border ;-)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1905/be5518a44a184f443010cb53233291aaef670c9a..86def9261fbe61b9fb4700e1b640b9ed498f6bbb)*

senekor: That makes sense... I have a vague understanding of CSS units, but I usually just eyeball it :sweat_smile:

steffahn (approved):

steffahn:
scss would support line comments by the way, i.e.
```scss
// this style

/* rather than this */
```

Manishearth (approved):

senekor:
> scss would support line comments by the way, i.e.
> 
> ```css-scss
> // this style
> 
> /* rather than this */
> ```

I see! I was just using the toggle-comment feature of Helix. Turns out it's just not (yet) configured to know that scss (and less) support line comments as well. Will [try to upstream](https://github.com/helix-editor/helix/pull/16155) :)

Manishearth (approved):

sass/styles/app.scss
steffahn: @senekor feel free to press "resolve conversation" here, I don't seem to have the button myself

senekor: ah, forgot about this :)

steffahn: was this actually what had blocked the move onto the merge queue?
(If so, GitHub didn't really indicate that particularly well, at least to me)

senekor: I think so, yeah. Once I clicked the resolve button, the PR got added to the merge queue.

steffahn: I only figured this out because I spotted the "require-conversation-resolution" [in the team repo](https://github.com/rust-lang/team/blob/main/repos/rust-lang/blog.rust-lang.org.toml#L25) whilst searching for any reason for the delay :D

(also, why won't GitHub allow me to resolve *my own* conversation thread 🤷🏻‍♂️)

PR #1927, opened 2026-08-26
Build Zola in debug mode

Kobzol:
I just waited >5 minutes for Zola to be built, and almost missed a deadline to publish a blog post that was synchronized with other Rust Foundation blog posts :sweat_smile: I think that we can make Zola's uncached compilation a bit faster.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,7 +42,9 @@ jobs:
         with:
           env-vars: CARGO RUST ZOLA_VERSION
       - name: Install Zola
-        run: cargo install --locked --git https://github.com/getzola/zola --rev ${{ env.ZOLA_VERSION }}
+        # Installing in release can take >5 minutes (in case the GitHub cache is not primed).
+        # We don't really need the performance, so we prefer faster compilation.
+        run: cargo install --locked --git https://github.com/getzola/zola --rev ${{ env.ZOLA_VERSION }} --debug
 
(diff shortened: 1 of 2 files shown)


senekor:
I think the build should already be cached. If it isn't, then we should fix that. If it is a requirement that the blog can be deployed with precise timing, then we should rethink CI with that in mind.

Kobzol:
It is usually cached, but today it wasn't, because the cache lives only for 7 (or 10?) days, IIRC. Or maybe it disappeared because GHA is having issues today. In any case it would be nice if the uncached build was faster.

kpreid:
It’s possible to increase the cache retention time in the repository settings. Or, instead, a scheduled workflow could keep the cache fresh and rebuild if it’s missing.

Kobzol:
Uncached:
- [Release](https://github.com/rust-lang/blog.rust-lang.org/actions/runs/32982975702/job/98223966600): 6m 43s Zola build, 1s blog build
- [Debug](https://github.com/rust-lang/blog.rust-lang.org/actions/runs/33000842815/job/98281959187): 2m 11s Zola build, 8s blog build

I think that it's a better trade-off here to build Zola in debug mode.

senekor (approved):

senekor:
Actually, can you update `snapshot_test.yml` as well before merging?

Kobzol:
Good catch, done.

PR #1937, opened 2026-09-02
Fix unsized locals link in Rust 1.35.0.md

abellaismail7:
Updated the broken link for unsized locals documentation from nightly to beta.

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/main/content/Rust-1.35.0.md)


programmerjake:
rather than linking to nightly or beta since those constantly change, imo it would be better to link to the version matching that release, so https://doc.rust-lang.org/1.35.0/unstable-book/language-features/unsized-locals.html

abellaismail7:
That makes sense. I updated the link

senekor (approved):

PR #1941, opened 2026-09-15
Add 32-bit Windows targets demotion

mati865:
[Rendered](https://github.com/mati865/blog.rust-lang.org/blob/i686-windows-demotions/content/demoting-i686-windows-targets-to-std-only.md)


programmerjake:
maybe more strongly emphasize that users can just install the x86-64 toolchains with the i686 std and everything should still just work (but you have to pass `--target` or change a config file to make that default)

mati865:
Dunno, Rustup already prevents you from installing i686 toolchain on other arch, unless you pass `--force-non-host`: https://github.com/rust-lang/rustup/pull/4935

programmerjake:
yeah, just I think we want to prevent people from reading the first line or two and thinking "Rust doesn't support 32-bit windows anymore", so I think we should explicitly point how to still make 32-bit windows binaries very close to the top so people who don't read very far aren't alarmed, rather than having it be a mention of still having `std` near the end which people who aren't super familiar with Rust may not understand very well.

content/demoting-i686-windows-targets-to-std-only.md
```diff
@@ -0,0 +1,27 @@
...
+authors = ["Mateusz Mikuła"]
+
+[extra]
+team = "the Compiler team"
+team_url = "https://www.rust-lang.org/governance/teams/compiler#team-compiler"
++++
+
+With Rust 1.100.0, the Tier 1 with host tools target `i686-pc-windows-msvc` will be demoted to Tier 1 without host tools. Respectively, the Tier 2 with host tools target `i686-pc-windows-gnu` will be demoted to Tier 2 without host tools.
+
+Builds of the standard library will continue to be distributed, but compiler will be no longer available. `i686-pc-windows-msvc` as the Tier 1 target, still undergoes CI testing.
```
RalfJung:
```suggestion
Builds of the standard library will continue to be distributed, but host tools such as the compiler will be no longer available. `i686-pc-windows-msvc` as a Tier 1 target still undergoes CI testing.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1941/2ec4e779ecc5a97cf5628017fb440d49723b9290..eec0c8612566dd4eaabce8df543a93138844df45)*

content/demoting-i686-windows-targets-to-std-only.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Background
+
+Desktop and Server 32-bit only x86 CPUs are no longer sold for over 15 years, and general 32-bit Windows support has ended in October 2025. That means the development platforms these targets are meant for, hardly exist these days. And they aren't capable enough for the development.
```
RalfJung:
```suggestion
Desktop and Server 32-bit only x86 CPUs are no longer sold for over 15 years, and general 32-bit Windows support has ended in October 2025. This means that the development platforms these targets are meant for hardly exist these days, and even if they do exist they typically aren't capable enough for development.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1941/2ec4e779ecc5a97cf5628017fb440d49723b9290..eec0c8612566dd4eaabce8df543a93138844df45)*

content/demoting-i686-windows-targets-to-std-only.md
```diff
@@ -0,0 +1,27 @@
...
+
+Even on the modern x86_64 hardware, building i686 Windows toolchains has proven to be problematic. From compiler binaries crashing when built with the i686 MSVC target, to the GNU C++ toolchain failing with OOMs during LLVM build.
```
RalfJung:
```suggestion
Even on the modern x86_64 hardware, building i686 Windows toolchains has proven to be problematic. We have encountered compiler binaries crashing when built with the i686 MSVC target, and the GNU C++ toolchain failing with OOMs during LLVM build.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1941/2ec4e779ecc5a97cf5628017fb440d49723b9290..eec0c8612566dd4eaabce8df543a93138844df45)*

content/demoting-i686-windows-targets-to-std-only.md
```diff
@@ -0,0 +1,27 @@
...
+
+Considering all these things, cross-compiling these targets from a better supported one is what we have found to be the best solution forward. As part of that, we stopped producing host tools for these targets. Prebuilt standard library is still available, and in case of `i686-pc-windows-msvc` still tested on the CI, by the time this post is published.
```
RalfJung:
```suggestion
Considering all these things, cross-compiling these targets from a better supported one is what we have found to be the best solution forward. As part of that, we stopped producing host tools for these targets. For the time being, the prebuilt standard library is still available, and in case of `i686-pc-windows-msvc` still tested on CI.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1941/2ec4e779ecc5a97cf5628017fb440d49723b9290..eec0c8612566dd4eaabce8df543a93138844df45)*

content/demoting-i686-windows-targets-to-std-only.md
```diff
@@ -0,0 +1,27 @@
...
+
+## What Changes?
+
+After Rust 1.100, it will no longer be possible to install toolchains on 32-bit Windows. Other 32-bit platforms that we already provide host tools for, are not impacted by this change.
```
RalfJung:
```suggestion
After Rust 1.100, it will no longer be possible to install toolchains on 32-bit Windows hosts. We recommend cross-compiling from a 64-bit Windows host instead. Other 32-bit platforms that we already provide host tools for are not impacted by this change.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1941/2ec4e779ecc5a97cf5628017fb440d49723b9290..eec0c8612566dd4eaabce8df543a93138844df45)*

RalfJung:
People are running into this on nightly: https://github.com/rust-lang/rustup/issues/5085.
So it'd be good to get a blog post published soon.

content/demoting-i686-windows-targets-to-std-only.md
```diff
@@ -0,0 +1,27 @@
...
+path = "9999/12/31/demoting-i686-windows-targets-to-std-only"
+title = "Demoting i686 Windows targets to std-only"
+authors = ["Mateusz Mikuła"]
+
+[extra]
+team = "the Compiler team"
+team_url = "https://www.rust-lang.org/governance/teams/compiler#team-compiler"
++++
+
+With Rust 1.100.0, the Tier 1 with host tools target `i686-pc-windows-msvc` will be demoted to Tier 1 without host tools. Respectively, the Tier 2 with host tools target `i686-pc-windows-gnu` will be demoted to Tier 2 without host tools.
```
djc: I think this might read more clearly as a bullet list highlighting two changes for two different targets?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1941/2ec4e779ecc5a97cf5628017fb440d49723b9290..eec0c8612566dd4eaabce8df543a93138844df45)*

content/demoting-i686-windows-targets-to-std-only.md
T-256: Shouldn't have mention how outdated (version <1.100) users could update to latest available version (1.100)?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1941/2ec4e779ecc5a97cf5628017fb440d49723b9290..eec0c8612566dd4eaabce8df543a93138844df45)*

mati865: How about?
> We recommend cross-compiling from a still-supported host instead.

RalfJung: 64-bit Win seems like the best candidate for most affected users, no?

mati865: In case of MSVC 64-bit Windows is the only way supported by MS, but unofficial solutions exist: https://github.com/mstorsjo/msvc-wine
For Windows GNU it doesn't matter what OS you use, so using your main OS might be more convenient than booting into Windows.

mati865: You cannot update 32-bit Windows host tools to 1.100 and you'll have to cross-compile from a supported host. Is this addressed by https://github.com/rust-lang/blog.rust-lang.org/pull/1941#discussion_r4024867389 ?

mati865:
> yeah, just I think we want to prevent people from reading the first line or two and thinking "Rust doesn't support 32-bit windows anymore", so I think we should explicitly point how to still make 32-bit windows binaries very close to the top so people who don't read very far aren't alarmed, rather than having it be a mention of still having `std` near the end which people who aren't super familiar with Rust may not understand very well.

This should be addressed by the updated "What Changes?" section.

content/demoting-i686-windows-targets-to-std-only.md
RalfJung: How about

> We recommend cross-compiling from a still-supported host (such as a 64-bit Windows MSVC/GNU toolchain) instead.

RalfJung: Or maybe

> Generally, we recommend building 64-bit binaries for Windows instead. If building 32-bit binaries is required, we recommend cross-compiling from a still-supported host (such as a 64-bit Windows MSVC/GNU toolchain).

PR #1946, opened 2026-09-17
Be more precise about prior attacks

Manishearth:
We're not yet sure if these are all the same campaign.

This moves the prior art into the "what we've seen" section anchoring that section in prior instances we've seen, with context that doesn't require clicking through.

[Rendered](https://github.com/rust-lang/blog.rust-lang.org/blob/main/content/targeted-attacks.md)


LawnGnome (approved):
I don't love pushing down the "what you can do" section, but I do like your framing, so let's go with this.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/blog.rust-lang.org/1946/20b0d2ea77c598fc01b63d80bc8069fef3219a91..3d0cc5b726dabfb9558303991f83703ffe12f55e)*

lqd (approved):

